PR #29, opened 2015-01-07
1.2 improvements

ecino:
Few issues corrected and improvements including:

- When changing an active contract (payment_term or contract_lines), a problem was setting a wrong next_invoice_date and was not syncing well GP.
- Various GP sync problems found and fixed.
- Autocompletion : invoices generated for contact persons instead of companies
- Add ability to link invoice_line to contract manually for more flexibility
- Encoding error of export_tools (lsv,dd)
- Improved CRM views to better track origin of sponsorships (add salesperson, won sponsorships, ...)

### DIFF
--- a/account_statement_completion_compassion/model/statement.py
+++ b/account_statement_completion_compassion/model/statement.py
@@ -12,9 +12,12 @@
 from openerp.osv import orm, fields
 from openerp.addons.account_statement_base_completion.statement \
     import ErrorTooManyPartner
+from openerp.tools import DEFAULT_SERVER_DATE_FORMAT as DF
 from openerp import netsvc
 
 from sponsorship_compassion.model.product import GIFT_TYPES
+from datetime import datetime
+from dateutil.relativedelta import relativedelta
 import time
 
 
@@ -265,6 +268,26 @@ class AccountStatementCompletionRule(orm.Model):
                     wf_service.trg_validate(
                         uid, 'account.invoice', invoice_id, 'invoice_open',
                         cr)
+                elif product.name == GIFT_TYPES[0]:
+                    # Set date of invoice two months before child's birthdate
+                    child_birthdate = res.get('child_birthdate')
+                    if child_birthdate:
+                        inv_date = datetime.strptime(st_line['date'], DF)
+                        birthdate = datetime.strptime(child_birthdate, DF)
+                        new_date = inv_date
+                        if birthdate.month >= inv_date.month + 2:
+                            new_date = inv_date.replace(
+                                day=28,
+                                month=birthdate.month-2)
+                        elif birthdate.month + 3 < inv_date.month:
+                            new_date = birthdate.replace(
+                                day=28, year=inv_date.year+1) + relativedelta(
+                                months=-2)
+                            new_date = max(new_date, inv_date)
+                        invoice_obj.write(
+                            cr, uid, invoice_id, {
+                                'date_invoice': new_date.strftime(DF)
+                            }, context)
 
         return res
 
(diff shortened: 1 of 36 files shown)


CSester:
Are you going to fix Travis before my review? Or should I review quickly?

ecino:
Travis is fixed.

ecino:
Ok, this review is needed soon because we will need to put in production version 1.4 in three weeks. And by the time both 1.3+1.4 should be reviewed (and they contain lot of changes :o:)

Thank you for your help ;-)

account_statement_completion_compassion/model/statement.py
```diff
@@ -151,16 +155,20 @@ def get_from_amount(self, cr, uid, id, st_line, context=None):
                 [('type', '=', 'in_invoice'), ('state', '=', 'open'),
                  ('amount_total', '=', abs(amount))], context=context)
             res = {}
+            partner_obj = self.pool.get('res.partner')
             if invoice_ids:
                 if len(invoice_ids) == 1:
```
CSester: Test here seems to be useless. You can simply keep the else part. The loop checking that all invoices will have no side effect on the len = 1 case.

account_statement_completion_compassion/model/statement.py
```diff
@@ -259,6 +268,26 @@ def _generate_invoice(self, cr, uid, st_line, partner, context=None):
                     wf_service.trg_validate(
                         uid, 'account.invoice', invoice_id, 'invoice_open',
                         cr)
+                elif product.name == GIFT_TYPES[0]:
```
CSester: Is GIFT_TYPES[0] birthday gift? If yes, please add a comment.

account_statement_completion_compassion/model/statement.py
```diff
@@ -259,6 +268,26 @@ def _generate_invoice(self, cr, uid, st_line, partner, context=None):
...
+                    # Set date of invoice two months before child's birthdate
+                    child_birthdate = res.get('child_birthdate')
+                    if child_birthdate:
+                        inv_date = datetime.strptime(st_line['date'], DF)
+                        birthdate = datetime.strptime(child_birthdate, DF)
+                        new_date = inv_date
```
CSester: I don't get the purpose of all this. If you simply want to substract 2 months from birthdate, why not simple do
```python
inv_date = child_birthdate + relativedelta(months=-2)
```
? relativedelta handles inter-year deltas !

partner_compassion/model/partner_compassion.py
```diff
@@ -18,12 +18,19 @@
 
 
 class ResPartner(orm.Model):
+
```
CSester: Why this useless line?

partner_compassion/model/partner_compassion.py
```diff
@@ -18,12 +18,19 @@
...
     """ This class upgrade the partners to match Compassion needs.
         It also synchronize all changes with the MySQL server of GP.
     """
 
     _inherit = 'res.partner'
 
+    def _lang_get(self, cr, uid, context=None):
+        lang_pool = self.pool.get('res.lang')
```
CSester:
```python
lang_obj = ...
```
to stay consistent.

CSester:
Good job. :+1:

account_statement_completion_compassion/model/statement.py
ecino: In fact it is not as easy as the comment suggest it... if the payment is made less than three months after the birthday, the invoice_date will be at the same date of the payment.
In other situation, it is set two months before the birthday.

Moreover, child_birthdate has a very old year, we should keep the current year for the invoice, or next year if birthday is in future.

CSester: Ok, so if the code fits the needs, please update the comment. A bad comment is worse than no comment at all...

CSester:
Emanuel, you can merge this one. This way, the review for 1.3 and 1.4 will be easier for me.

ecino:
I'll fix the bad comment in a following branch.

PR #32, opened 2015-01-27
bug fix on end date for contract

davidZIVI:
bug fix on end date for contract

### DIFF
--- a/recurring_contract/model/recurring_contract.py
+++ b/recurring_contract/model/recurring_contract.py
@@ -124,7 +124,8 @@ class recurring_contract(orm.Model):
             states={'draft': [('readonly', False)]},
             track_visibility="onchange"),
         'end_date': fields.date(
-            _('End date'), readonly=True,
+            _('End date'), readonly=False,
+            states={'terminated': [('readonly', True)]},
             track_visibility="onchange"),
         'next_invoice_date': fields.date(
             _('Next invoice date'), readonly=False,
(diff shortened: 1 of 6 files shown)


recurring_contract/model/recurring_contract.py
```diff
@@ -124,8 +124,7 @@ def _get_total_amount(self, cr, uid, ids, name, args, context=None):
             states={'draft': [('readonly', False)]},
             track_visibility="onchange"),
         'end_date': fields.date(
-            _('End date'), readonly=False,
-            states={'terminated': [('readonly', True)]},
+            _('End date'), readonly=True,
```
ecino: You can make this change in module sponsorship_compassion instead of module recurring_contract.

recurring_contract/model/recurring_contract.py
```diff
@@ -343,8 +342,7 @@ def contract_active(self, cr, uid, ids, context=None):
         return True
 
     def contract_terminated(self, cr, uid, ids, context=None):
-        today = datetime.today().strftime(DF)
-        self.write(cr, uid, ids, {'state': 'terminated', 'end_date': today})
+        self.write(cr, uid, ids, {'state': 'terminated'})
```
ecino: You can make this change in module sponsorship_compassion as well (and you are not obliged to call super() from the inherited method, to avoid calling this write).

ecino:
:+1:

PR #34, opened 2015-01-30
1.3 better invoicing

ecino:
### This branch modifies a lot the behavior of invoicing and adds functionalities to ease the process of reconciliation.

- Many bug fixes (as always)

#### Clean invoices

- When a contract or a contract group is modified, the goal is to modify all existing open invoices so that all changes are directly committed to existing invoices (except for old invoices), and so that the reconciliation can be done directly after having modified a contract.
- When a contract is terminated, if the sponsor did pay in advance, we automatically unreconcile the payment so that we can later decide for what we use the payment. The invoices are then canceled.

#### Others

- New split_invoice module for splitting an existing invoice into two invoices (useful if sponsor pays for one over several sponsorships)
- Birthday gifts invoices : set automatically the date of invoice two months before the child's birthday
- CRM improvements : allow event deletion, keep track of the salesperson of a sponsorship, ...
- Generate invoices from bank statements, given some information on the payment (useful when autocompletion is not feasible)
- Allow changing the partner of an existing contract (when someone else will pay)
- LSV/DD references issues fixed (don't attribute reference to those terms except multi-months for groupping the direct debit order).
- Complete refactoring of file sponsorship_compassion/model/contracts.py as this file is becoming huge ! Maybe we should even think of splitting the module into several ones ?


CSester:
Ok. I will review as soon as 1.2 is merged into master (to see a clear diff).

account_invoice_split_invoice/model/account_invoice.py
```diff
@@ -0,0 +1,37 @@
...
+import logging
+
+from openerp.osv.orm import Model, fields
+from openerp.tools.translate import _
+
+logger = logging.getLogger(__name__)
+
+
+class account_invoice_line(Model):
+
```
CSester: Useless empty line

account_invoice_split_invoice/model/account_invoice.py
```diff
@@ -0,0 +1,37 @@
...
+    ''' Adds a field for moving a line to a new invoice'''
+    _inherit = "account.invoice.line"
+
+    def _get_split_line(self, cr, uid, ids, field_name, args, context=None):
```
CSester: What does this function exactly

account_invoice_split_invoice/wizard/account_invoice_split_wizard.py
```diff
@@ -0,0 +1,110 @@
...
+            obj='account.invoice.line', method=True,
+            string=_('Invoice lines')),
+    }
+
+    _defaults = {
+        'invoice_id': _get_invoice_id,
+        'invoice_line_ids': _get_default_ids,
+    }
+
+    def split_invoice(self, cr, uid, ids, context=None):
```
CSester: What is the purpose of a function doing nothing?

ecino: When the wizard is closed, it calls the _write_lines (when lines are changed). I don't know if I can have a save button that can close the wizard without calling any function but triggering the write. Maybe you have an idea or I will look if I find something.

account_invoice_split_invoice/model/account_invoice.py
ecino: I added a comment.

PR #39, opened 2015-02-09
Child delegation

davidZIVI:
- Add function to delegate/undelegate childrens
- Add function to add/remove childrens to/from the website


account_statement_completion_compassion/__openerp__.py
```diff
@@ -36,7 +36,7 @@
     "depends": ['account_statement_base_completion',
                 'sponsorship_compassion',
                 ],
-    "data": ['view/statement_view.xml',
+    "data": [# 'view/statement_view.xml',
```
ecino: Please remove this change from PR.

ecino:
Please add field delegated_to and delegated_comment in tree_view and form_view of children.

child_compassion/model/child_compassion.py
```diff
@@ -504,13 +547,122 @@ def child_remove_from_typo3(self, cr, uid, ids, context=None):
...
 
-        self._typo3_scripts_fetch(scripts_url, api_key, "upd_db")
+        return self._typo3_scripts_fetch(scripts_url, api_key, "upd_db")
+
+    def child_add_to_typo3(self, cr, uid, ids, context=None):
+        reload(sys)
+        sys.setdefaultencoding('UTF8')
+
+        # To avoid readding children
+        self.child_remove_from_typo3(cr, uid, ids, context=None)
```
ecino: This check is useless if we cannot add children that has already the state "On Internet".

child_compassion/wizard/child_on_internet_wizard.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def put_child_on_internet(self, cr, uid, ids, context=None):
+        child_obj = self.pool.get('compassion.child')
+        childrens = child_obj.browse(
+            cr, uid, context.get('active_ids', list()), context)
+        child_ids = []
+
+        for child in childrens:
+            possible_states = ['N', 'R']
+            if child.state in possible_states:
```
ecino: Please also prevent to add children that miss either french or german description.

child_compassion/view/child_on_internet_wizard.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+	<data>
+        <record id="child_on_internet_wizard_view" model="ir.ui.view">
+            <field name="name">compassion.child.on.internet.wizard</field>
+			<field name="model">child.on.internet.wizard</field>
+            <field name="type">form</field>
+			<field name="arch" type="xml">
+                <form string="Put child on internet" version="7.0">
+                    <footer>
```
ecino: Please add a small paragraph explaining the procedure (<p>description</p>).
Explain this will put available children on the website, and add a sentence recalling the user to go update the index on typo3 after new children are sent.

child_compassion/view/child_remove_from_internet.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+	<data>
+        <record id="child_remove_from_internet_view" model="ir.ui.view">
+            <field name="name">compassion.child.remove.from.internet</field>
+			<field name="model">child.remove.from.internet</field>
+            <field name="type">form</field>
+			<field name="arch" type="xml">
+                <form string="Put child on internet" version="7.0">
+                    <footer>
```
ecino: Add also a paragraph telling it will remove the children and recalling to update the index on typo3.

ecino:
Can you add a field date_delegation that is set to the date at which the delegation is made and is set to null when a child is undelegated ?
The field should be displayed in the tree view and in the form view between delegated to and delegated comment.

ecino:
In all your wizards, you can add in the view the list of selected children that will be treated (and exclude those that are not valid) so that the user can see a summary of what will be done. If no children appears he knows nothing will happen.

child_compassion/model/child_compassion.py
```diff
@@ -476,25 +485,59 @@ def child_departed(self, cr, uid, ids, context=None):
...
+        res = self._sel_request_to_typo3(
+            cr, uid,
+            "select * "
+            "from tx_drechildpoolmanagement_domain_model_children "
+            "where child_key='%s'" % child_code,
+            context)
+
+        return json.loads(res)[0]['uid']
+
+    def _sel_request_to_typo3(self, cr, uid, request, context=None):
```
ecino: `_sel_request_to_typo3` and `_update_request_to_typo3` are almost the same method except for file_name used and script_name called. You could merge these two methods and add parameters for these different values to save lines of code.

child_compassion/model/child_compassion.py
```diff
@@ -504,13 +547,115 @@ def child_remove_from_typo3(self, cr, uid, ids, context=None):
             with sftp.cd(path):
                 sftp.put(filename)
 
-        self._typo3_scripts_fetch(scripts_url, api_key, "upd_db")
+        return self._typo3_scripts_fetch(scripts_url, api_key, "upd_db")
+
+    def child_add_to_typo3(self, cr, uid, ids, context=None):
+        reload(sys)
```
ecino: Add a small comment explaining our encoding problem.

child_compassion/model/child_compassion.py
```diff
@@ -504,13 +547,115 @@ def child_remove_from_typo3(self, cr, uid, ids, context=None):
...
+        sys.setdefaultencoding('UTF8')
+
+        for child in self.browse(cr, uid, ids, context):
+
+            child_gender = self._get_gender(cr, uid, child.gender, context)
+            child_image = child.code+"_f.jpg,"+child.code+"_h.jpg"
+            child_birth_date = calendar.timegm(
+                datetime.strptime(child.birthdate, DF).utctimetuple())
+
+            child_desc_de = child.desc_de.replace('\'', '\'\'')
```
ecino: Add a small comment on why you replace this character.

child_compassion/model/child_compassion.py
```diff
@@ -504,13 +547,115 @@ def child_remove_from_typo3(self, cr, uid, ids, context=None):
...
+                context)
+
+        self._add_child_pictures_to_typo3(cr, uid, ids, context)
+
+    def _add_child_pictures_to_typo3(self, cr, uid, ids, context=None):
+        for child in self.browse(cr, uid, ids, context):
+            head_image = child.code+"_h.jpg"
+            full_image = child.code+"_f.jpg"
+
+            if not(child.portrait and child.fullshot):
```
ecino: I would move this check in method `put_child_on_internet` of the wizard, so that if a child is missing its picture, it is not inserted in the typo3 database.

child_compassion/model/child_compassion.py
```diff
@@ -504,13 +522,115 @@ def child_remove_from_typo3(self, cr, uid, ids, context=None):
...
+
+        self._add_child_pictures_to_typo3(cr, uid, ids, context)
+
+    def _add_child_pictures_to_typo3(self, cr, uid, ids, context=None):
+        for child in self.browse(cr, uid, ids, context):
+
+            head_image = child.code+"_h.jpg"
+            full_image = child.code+"_f.jpg"
+
+            if not(child.portrait and child.fullshot):
```
ecino: This if block is also to be moved in the wizard, as it cannot happen here.

ecino:
:+1:

PR #55, opened 2015-03-04
Change date to datetime into gmc_message_pool

davidZIVI:
Add cron to process messages

Child same picture checked

### DIFF
--- a/child_compassion/model/child_pictures.py
+++ b/child_compassion/model/child_pictures.py
@@ -15,7 +15,7 @@ from openerp.tools.translate import _
 
 from datetime import date
 import requests
-
+import pdb
 
 class child_pictures(orm.Model):
     """ Holds two pictures of a given child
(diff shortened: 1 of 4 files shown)


child_compassion/model/child_pictures.py
```diff
@@ -77,8 +77,43 @@ def create(self, cr, uid, vals, context=None):
...
+        child_picture = self.browse(
+            cr, uid, res_id, context)
+        same_picture_ids = self._find_same_picture(
+            cr, uid, child.id,
+            child_picture.fullshot, child_picture.headshot,
+            context)
+        same_picture_ids.remove(child_picture.id)
+
+        if same_picture_ids:
+            self._unlink_related_attachment(cr, uid, child_picture.id, context)
```
ecino: Nice method, you can call it as well after line 77, in order to remove attachments in case of failure.

child_compassion/model/child_pictures.py
```diff
@@ -77,8 +77,43 @@ def create(self, cr, uid, vals, context=None):
...
+            context=context)
+        attachment_obj.unlink(cr, uid, attachment_ids, context)
+
+    def _find_same_picture(
+            self, cr, uid, child_id, fullshot, headshot, context):
+        pict_ids = self.search(
+            cr, uid, [('child_id', '=', child_id)], context=context)
+        same_pict_ids = list()
+        for picture in self.browse(cr, uid, pict_ids, context):
+            if picture.fullshot == fullshot:
```
ecino: So does this test work finally ?

child_compassion/model/child_pictures.py
```diff
@@ -77,8 +77,43 @@ def create(self, cr, uid, vals, context=None):
...
+                if picture.headshot == headshot:
```
ecino: Instead of having two consecutive if, you can add "and" in the first if.

ecino:
:+1:

PR #61, opened 2015-03-19
1.4 gmc prepare stablization

ecino:
Those are the latest commits on branch 1.4 which might not have been reviewed since we have been commiting a lot those last days (in order to improve based on the users feedbacks).

### DIFF
--- a/child_compassion/model/project_compassion.py
+++ b/child_compassion/model/project_compassion.py
@@ -60,15 +60,16 @@ class compassion_project(orm.Model):
         By default: log a message.
         """
         self.pool.get('mail.thread').message_post(
-                cr, uid, project_id,
-                "The project was suspended and funds are retained<b>"
-                "{}</b>.<br/>".format(
-                    date_end.strftime(" until %B %Y") if date_end else ""),
-                "Project Suspended", 'comment',
-                context={'thread_model': self._name})
+            cr, uid, project_id,
+            "The project was suspended and funds are retained<b>"
+            "{}</b>.<br/>".format(
+                date_end.strftime(" until %B %Y") if date_end else ""),
+            "Project Suspended", 'comment',
+            context={'thread_model': self._name})
         return True
 
     def _has_desc(self, cr, uid, ids, field_names, args, context=None):
+
         res = dict()
         field_res = dict()
         for child in self.browse(cr, uid, ids, context):
@@ -234,6 +235,10 @@ class compassion_project(orm.Model):
             readonly=True, track_visibility="onchange"),
     }
 
+    _defaults = {
+        'name': '/'
+    }
+
     def update_informations(self, cr, uid, ids, context=None):
         """ Get the most recent informations for selected projects and update
             them accordingly. """
@@ -303,7 +308,7 @@ class compassion_project(orm.Model):
             'name': _('Description generation'),
             'type': 'ir.actions.act_window',
             'view_type': 'form',
-            'view_mode': 'form',
+            'view_mode': 'auto_description_form',
             'res_model': 'project.description.wizard',
             'context': context,
             'target': 'new',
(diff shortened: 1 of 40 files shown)


child_compassion/model/project_compassion.py
```diff
@@ -60,15 +60,16 @@ def suspend_funds(self, cr, uid, project_id, context=None,
...
+            cr, uid, project_id,
+            "The project was suspended and funds are retained<b>"
+            "{}</b>.<br/>".format(
+                date_end.strftime(" until %B %Y") if date_end else ""),
+            "Project Suspended", 'comment',
+            context={'thread_model': self._name})
         return True
 
     def _has_desc(self, cr, uid, ids, field_names, args, context=None):
+
```
CSester: not necessary

child_compassion/model/translated_value.py
```diff
@@ -39,6 +39,14 @@ class translated_value(orm.Model):
         'is_tag': False,
     }
 
+    def get_translated_value(self, cr, uid, id, lang, context):
+        value = self.browse(cr, uid, id, context)[0]
+        id = value.value_en
+        translated_value = getattr(value, "value_"+lang) or value.value_en
+        color = 'red' if not getattr(value, "value_"+lang) else 'blue'
+        return u'<span id="{}" style="color:{}">{}</span>'.format(
```
CSester: index of argument rather than {}

child_compassion/model/translated_value.py
```diff
@@ -39,6 +39,14 @@ class translated_value(orm.Model):
         'is_tag': False,
     }
 
+    def get_translated_value(self, cr, uid, id, lang, context):
+        value = self.browse(cr, uid, id, context)[0]
+        id = value.value_en
```
CSester: Confusing, because id refers always to object id in odoo

child_compassion/static/src/js/child_description_wizard.js
```diff
@@ -0,0 +1,59 @@
...
+            // Child description case
+            if (this.fields.hasOwnProperty('child_property_value_ids')) {
+                var values = this.get_field_value('child_property_value_ids');
+            }
+            // Project description case
+            if (this.fields.hasOwnProperty('project_property_value_ids')) {
+                var values = this.get_field_value('project_property_value_ids');
+            }
+            
+            for(index in values) {
```
CSester: Values is declared as a list. But here you iterate over all elements and then call values[element] as if it where a dictionnary. Please explain, I'm lost.

message_center_compassion/model/contracts.py
```diff
@@ -140,7 +142,7 @@ def _invoice_paid(self, cr, uid, invoice, context=None):
                 # 1. Check if a GIFT was sent to GMC and prevent unrec
                 mess_ids = message_obj.search(cr, uid, [
                     ('invoice_line_id', '=', invoice_line.id),
-                    ('state', '=', 'success'),
+                    ('state', 'in', ['success', 'fondue']),
```
CSester: ???

message_center_compassion/model/gmc_message.py
```diff
@@ -107,6 +107,7 @@ def _get_object_id(self, cr, uid, ids, field_name, args, context=None):
         'state': fields.selection(
             [('new', _('New')),
              ('pending', _('Pending')),
+             ('fondue', _('To deliver')),
```
CSester: ???

child_compassion/static/src/js/child_description_wizard.js
ecino: I was confused too by this code. It's because of javascript loops syntax. The for statement will only put a counter in variable index, so in index will be 0 to length of values.

message_center_compassion/model/contracts.py
ecino: fondue is a new state meaning "Fonds dû" => we must send the money of the gift to Compassion US.

message_center_compassion/model/gmc_message.py
ecino: Don't you like cheese ?

ecino:
Are you ok with these changes ? You can ignore the commits more recent than your last review, I included them in another pull request in order to better see what has changed since. If yes, we can close this request.

ecino:
I move this pull request so that no new commits are included since your last review.

PR #62, opened 2015-03-23
Set sds state (no sub, sub accept, sub reject) at install

davidZIVI:
Set cancelled sds state at install

Set last sds state date at install depending on sds state

Add a kanban view to SDS Tracking View

Allow user to send workflow signal from kanban view

Fix group by for last sds state date

Set project state at install

Add search by partner reference

Change sub accept/reject cron method behavior

Other few little changes


ecino:
Rebase on branch master.

PR #65, opened 2015-03-26
Remove everthing concerning the sync with typo3 (New module in compassio...

davidZIVI:
- New module in compassion-switzerland


ecino:
Thanks!

PR #67, opened 2015-03-26
Review - Bound Case Studies and Child Pictures

ecino:
This contains the latest changes on 1.4 branch (this branch has still some corrections pending !).

We bounded case studies with child pictures, because on GP, we send updates to sponsors only when we have both a new case study and a new image. We had a problem  : when updating info of child, we sometimes put new pictures to old case studies in GP. With these changes, we assure we don't mess with GP until we have complete information to be useful.

The update contains migration scripts (v1.4.2) to bound last case study to last image.
Migration script 1.4.1 is because we changed a field from char text to translated value.

All these changes are also in other review pull request but can be reviewed here for more clarity.

Thanks.

### DIFF
--- a/child_compassion/__openerp__.py
+++ b/child_compassion/__openerp__.py
@@ -30,7 +30,7 @@
 
 {
     'name': 'Compassion Children',
-    'version': '1.4',
+    'version': '1.4.2',
     'category': 'Other',
     'description': """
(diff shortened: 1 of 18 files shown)


ecino:
I will redo this pull request from a copy of branch 1.4 so that we don't integrate other commits in the future (keep small pull requests).

PR #68, opened 2015-03-26
1.4 review gmc prepare stablization 19/03

ecino:
Those are the commits on branch 1.4 which might not have been reviewed since we have been commiting a lot those last days (in order to improve based on the users feedbacks).


ecino:
Pull request as it was one week ago ;)

CSester:
:+1:

PR #69, opened 2015-03-26
1.4 review - Bound Case Studies and Child Pictures (26/03)

ecino:
Those are the latest commits on branch 1.4 which might not have been reviewed since we have been commiting a lot those last days (in order to improve based on the users feedbacks).

Warning : merge #68 before this.

### DIFF
--- a/child_compassion/model/child_pictures.py
+++ b/child_compassion/model/child_pictures.py
@@ -67,6 +67,7 @@ class child_pictures(orm.Model):
         and attach the pictures to the last case study.
         """
         res_id = super(child_pictures, self).create(cr, uid, vals, context)
+
         child = self.pool.get('compassion.child').browse(
             cr, uid, vals['child_id'], context)
         # Retrieve Fullshot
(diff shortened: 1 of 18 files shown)


child_compassion/model/child_pictures.py
```diff
@@ -124,11 +138,13 @@ def _get_picture(self, cr, uid, child_id, child_code, attach_id,
                      type='Headshot', dpi=72, width=400, height=400,
                      format='jpeg', context=None):
         ''' Gets a picture from Compassion webservice '''
-        url = self.pool.get('compassion.child').get_url(child_code, 'image')
+        url = self.pool.get('compassion.child').get_url(
+            child_code, 'image/2015/03')
```
CSester: March 2015? Is this for test purpose ?

CSester:
:+1: For everything except the strange "march 2015" subfolder. :+1:

ecino:
It is strange, but it's how GMC decided to name it's new webservice based on our change requirements. Instead of updating the existing service, they added a new one with this url ! This works in production.

PR #72, opened 2015-03-31
1.4 review - Prepare for updating all child case studies (31/03)

ecino:
Change structure of case studies and child pictures so that we can update all information and synchronize GP in a good way.

Corrections in german descriptions.

Small additions and corrections.


child_compassion/__openerp__.py
```diff
@@ -30,7 +30,7 @@
 
 {
     'name': 'Compassion Children',
-    'version': '1.4.2',
+    'version': '1.4.1',
```
CSester: Downgrade ?

CSester: Downgrade?

child_compassion/model/child_compassion.py
```diff
@@ -276,8 +276,13 @@ def get_infos(self, cr, uid, ids, context=None):
         proj_obj = self.pool.get('compassion.project')
         res = True
         for child in self.browse(cr, uid, ids, context):
-            res = res and self._get_case_study(cr, uid, child, context)
-            self._get_basic_informations(cr, uid, child.id)
+            if child.type != 'LDP':
```
CSester: Suggestion:
```python
if child.type == 'LDP':
```
and invert test. To me (and it's very subjective), it's easier to read a conditional test rather than the "inverse-test". But, once more, this is only a style suggestion.

child_compassion/model/child_compassion.py
```diff
@@ -340,12 +374,19 @@ def _get_case_study(self, cr, uid, child, context=None):
...
-        json_data = r.json()
-        if not r.status_code / 100 == 2:
+        html_res = r.text
+        json_data = dict()
+        error = r.status_code != 200
+        try:
+            json_data = r.json()
+        except:
+            error = True
+        if error:
```
CSester: You could raise the custom exception from inside the except block. What is the reason to use a flag rather than directly do the job ?

child_compassion/model/child_pictures.py
```diff
@@ -134,28 +148,33 @@ def _find_same_picture(
                     same_pict_ids.append(picture.id)
         return same_pict_ids
 
-    def _get_picture(self, cr, uid, child_id, child_code, attach_id,
+    def _get_picture(self, cr, uid, child_id, child_code, attach_id, context,
```
CSester: By convention, context is an optional keyword-argument for every function.

child_update_picture_date/wizard/__init__.py
```diff
@@ -0,0 +1,12 @@
+# -*- encoding: utf-8 -*-
```
CSester: A new module only to run a cron job? I think this would be acceptable to include this in child_compassion.

ecino: We put it in a separate module because we needed to perform the update only once, and then we can uninstall the module and never use this code anymore.

CSester: Ok. But then why setup a cron job to perform the task once? You could add a button to the child tree view header to launch the task manually

ecino: Because we launched the update on friday night (the process take one day!), and the cron is convenient for executing background tasks.

PR #73, opened 2015-04-02
Cancel old openend invoices when activating sponsorship

davidZIVI:


sponsorship_compassion/model/contracts.py
```diff
@@ -63,6 +75,42 @@ def _get_contract_from_invoice(invoice_obj, cr, uid, invoice_ids,
...
+
+        for invoice in invoice_obj.browse(cr, uid, invoice_ids, context):
+            invoice_line_ids = invoice_line_obj.search(
+                cr, uid,
+                [('invoice_id', '=', invoice.id)],
+                context=context)
+            invoice_lines = invoice_line_obj.browse(
+                cr, uid,
+                invoice_line_ids,
+                context)
```
ecino: You can replace lines 91-98 by : invoice_lines = invoice.invoice_line

sponsorship_compassion/model/contracts.py
```diff
@@ -63,6 +75,42 @@ def _get_contract_from_invoice(invoice_obj, cr, uid, invoice_ids,
...
+            contract_ids = [
+                invoice_line.contract_id.id for invoice_line in invoice_lines]
```
ecino: add check if invoice_line.contract_id, otherwise this line will fail if there are some invoice lines unlinked to a contract_id.

sponsorship_compassion/model/contracts.py
```diff
@@ -48,11 +48,23 @@ def _get_contract_from_invoice(invoice_obj, cr, uid, invoice_ids,
...
-                                     if move_line.credit > 0] or [0])
+
+                pay_dates = [move_line.date
+                             for move_line in invoice.payment_ids
+                             if move_line.credit > 0] or [0]
+
+                last_pay_date = max(pay_dates)
+                first_pay_date = min(pay_dates)
+
+                self._cancel_invoices(
```
ecino: If you call cancel_invoices from here, you will call it each time an invoice is paid, which is not correct. We want to call it only if a contract is activated.

sponsorship_compassion/model/contracts.py
```diff
@@ -61,8 +67,43 @@ def _get_contract_from_invoice(invoice_obj, cr, uid, invoice_ids,
                         contract.write({
                             'activation_date': datetime.today().strftime(DF)})
 
+                        # Cancel the invoices if a contract is activated
```
ecino: Please change comment and name method to "cancel_old_invoices", as it is strange to read "Cancel invoices when sponsorship is activated." We don't wan't to cancel all invoices, but only those that are old.

sponsorship_compassion/model/contracts.py
```diff
@@ -61,8 +67,43 @@ def _get_contract_from_invoice(invoice_obj, cr, uid, invoice_ids,
...
+                        self._cancel_invoices(
+                            cr, uid,
+                            invoice.partner_id.id,
+                            first_pay_date,
+                            context)
         return list(res)
 
+    def _cancel_invoices(
+            self, cr, uid, partner_id, date_invoice, context=None):
```
ecino: You should also pass as parameter the contract_id that is activated, because you wan't to filter the invoice lines only on those for which invoice_line.contract_id.id = contract_id (i.e. you want only to cancel invoices that are for the same contract, not all invoices of the partner).

sponsorship_compassion/model/contracts.py
```diff
@@ -61,8 +67,43 @@ def _get_contract_from_invoice(invoice_obj, cr, uid, invoice_ids,
...
+
+            wf_service = netsvc.LocalService('workflow')
+            if contract_ids:
+                if len(contract_ids) == 1:
+                    wf_service.trg_validate(uid, 'account.invoice',
+                                            invoice.id, 'invoice_cancel', cr)
+                else:
+                    invoice_obj.action_cancel_draft(
+                        cr, uid, invoice.id, context)
+                    invoice_obj.trg_validate(uid, 'account.invoice',
```
ecino: Before validating again the invoice, you should remove the invoice line from the invoice (otherwise you are only cancelling and validating an invoice, which has no effect).

sponsorship_compassion/model/contracts.py
```diff
@@ -61,8 +67,52 @@ def _get_contract_from_invoice(invoice_obj, cr, uid, invoice_ids,
...
+             ('state', '=', 'open'),
+             ('date_invoice', '<', date_invoice)
+             ],
+            context=context)
+
+        for invoice in invoice_obj.browse(cr, uid, invoice_ids, context):
+            invoice_lines = invoice.invoice_line
+            contract_ids = [
+                invoice_line.contract_id.id for invoice_line in invoice_lines
+                if invoice_line.contract_id.id]
```
ecino: if invoice_line.contract_id

sponsorship_compassion/model/contracts.py
```diff
@@ -61,8 +67,52 @@ def _get_contract_from_invoice(invoice_obj, cr, uid, invoice_ids,
...
+            [('partner_id', '=', partner_id),
+             ('state', '=', 'open'),
+             ('date_invoice', '<', date_invoice)
+             ],
+            context=context)
+
+        for invoice in invoice_obj.browse(cr, uid, invoice_ids, context):
+            invoice_lines = invoice.invoice_line
+            contract_ids = [
+                invoice_line.contract_id.id for invoice_line in invoice_lines
```
ecino: you could rename invoice_line to invl inside the brackets because it is just used here, and it could save a bit of line length. But if you prefer like this, you can keep it.

sponsorship_compassion/model/contracts.py
```diff
@@ -61,8 +67,52 @@ def _get_contract_from_invoice(invoice_obj, cr, uid, invoice_ids,
...
+                if len(contract_ids) == 1:
+                    wf_service.trg_validate(uid, 'account.invoice',
+                                            invoice.id, 'invoice_cancel', cr)
+                else:
+                    invoice_obj.action_cancel_draft(
+                        cr, uid, invoice.id, context)
+                    invoice_lines = [
+                        invoice_line for invoice_line in invoice_lines
+                        if invoice_line.contract_id.id != contract_id]
+                    invoice_obj.write(
```
ecino: I would simply unlink invoice_lines which has contract_id = contract_id instead.

sponsorship_compassion/model/contracts.py
```diff
@@ -61,8 +67,55 @@ def _get_contract_from_invoice(invoice_obj, cr, uid, invoice_ids,
...
+                if len(contract_ids) == 1:
+                    wf_service.trg_validate(uid, 'account.invoice',
+                                            invoice.id, 'invoice_cancel', cr)
+                else:
+                    invoice_obj.action_cancel_draft(
+                        cr, uid, invoice.id, context)
+
+                    inv_line_ids = invoice_line_obj.search(
+                        cr, uid,
+                        [('contract_id', '=', contract_id)])
```
ecino: This will unlink all invoice lines attached to this contract ! You should limit to those in the current invoice.

ecino:
:+1:

PR #74, opened 2015-04-10
Odoo CRM changes

ecino:
# CRM Changes
## Events
- Opportunities can now link to many events
- The Events are now not automatically linked to a Project. Before, we used Projects to create an analytic account. Now, we only used Projects for having tasks (to-do list) for an event, but the analytic account is created directly from the event.
- We renamed analytic account of events and event names to remove the year from it (migration scripts).
- We added a type of analytic accounts specific for events.

## Contract Origins
- We now have the field user_id (Ambassador) that can be manually changed (no more a functional field). It is however set by default based on the origin.
- Ambassador is propagated in invoice lines (to count how much we gained from an ambassador).


crm_compassion/model/event_compassion.py
```diff
@@ -114,65 +129,82 @@ def _get_year(self, cr, uid, ids, field_name, arg, context=None):
...
-                if 'type' in vals:
-                    # Change parent of analytic account
-                    project_vals.update({
-                        'parent_id': self._find_parent_analytic(
-                            cr, uid, vals['type'], event.year, context)})
-                event.project_id.write(project_vals, context=ctx)
+
+        if context is None:
+            context = dict()
+        context['from_event'] = True
```
CSester: This is no longer a copy of context. Use carefully or the context will contain a lot of unneeded things (when going back to caller).

CSester:
:+1:

CSester:
Hum, seems to have conflicts here

ecino:
Rebased plus correction of context. I wonder if the diff changed after rebasing.

PR #78, opened 2015-04-20
1.5 Split compassion contract

davidZIVI:
Add a new module for compassion contract
Override compassion_contract in sponsorship
Add a type field to contract to manage sponsorship exceptions


contract_compassion/__openerp__.py
```diff
@@ -0,0 +1,50 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+
+{
+    'name': 'Compassion Contracts',
+    'version': '1.5',
```
ecino: Put 0.x until it is not stable. We will use 1.0 for first stable version.

contract_compassion/migrations/1.5/post-migration.py
```diff
@@ -0,0 +1,56 @@
...
+        UPDATE recurring_contract
+        SET type = 'O'
+        WHERE child_id IS NULL
+        '''
+    )
+    cr.execute(
+        '''
+        UPDATE product_product
+        SET type = 'S'
+        WHERE name_template IN ('Sponsorship', 'LDP', 'Sponsorship LDP')
```
ecino: Sponsorship LDP and LDP don't exist. You can replace by 'LDP Sponsorship' only.

contract_compassion/model/contract_group.py
```diff
@@ -0,0 +1,246 @@
...
+import time
+
+from openerp.osv import orm, fields
+from openerp.tools.translate import _
+from openerp.tools import mod10r
+from openerp import netsvc
+
+
+class contract_group(orm.Model):
+    ''' Add BVR on groups and add BVR ref and analytics_id
```
ecino: Please use double quotes for docstrings as we always used double quotes... Replace everywhere you can find them.

contract_compassion/model/contracts.py
```diff
@@ -0,0 +1,720 @@
...
+
+        if 'contract_line_ids' in vals:
+            self._on_contract_lines_changed(cr, uid, ids, context)
+        if 'group_id' in vals or 'partner_id' in vals:
+            self._on_group_id_changed(cr, uid, ids, context)
+
+        return res
+
+# just to modify access rights...
+class recurring_invoicer(orm.Model):
```
ecino: I saw that these overrides are useless. You can simply find the model in the security.xml file by putting module_name.model_name.

contract_compassion/model/product.py
```diff
@@ -0,0 +1,36 @@
...
+
+from openerp.osv import orm, fields
+from openerp.tools.translate import _
+import pdb
+
+class product(orm.Model):
+    _inherit = 'product.product'
+
+    _columns = {
+        'type': fields.selection([
```
ecino: Is it needed to add a product type ? If yes, you could add the type 'G', 'Gift', which would be useful everywhere I check product.name is in gift names...

ecino:
Please do pull request on branch 1.5-devel.

PR #83, opened 2015-04-29
1.5 - SDS Tracking review

ecino:
# SDS Tracking review

## sponsorship_tracking
New module sponsorship_tracking which adds two workflows to contracts in order to follow the sponsorship evolution:
* sds_state : new workflow for following all sds events that can occur for a contract from the beginning until the end. It helps to track child derpartures and be certain to propose sub sponsorships.
* project_state : new workflow for handling project suspensions with better details. 
* gmc_state is a third state but is not attached to a worfklow. It is used to keep track of events we receive from GMC when we receive Child Update messages.
* We created a new view for sponsorship_tracking and added a color convention in the kanban view to help visualize the whole lifecycle.

## child_sync_typo3
All functionality related to our typo3 website was removed from module child_compassion and put into a new module in the compassion_switzerland repository.

## partner_compassion
All functionality which was not only specific for compassion switzerland was taken out of module partner_compassion which was previously in switzerland repository. Now we have a module named partner_sync_gp in switzerland repository which only takes care of switzerland specific needs.


CSester:
:+1:

PR #84, opened 2015-05-04
1.5 split contracts review

ecino:
# Split contracts and sponsorships

@davidZIVI worked at splitting contracts and sponsorships into two categories of contracts, so that we can have different requirements for both. Changes are globally the following:

- [IMP] module recurring_contract improvements, corrections and addition of methods that are generic for all kind of contracts.
- [NEW] module contract_compassion depends on recurring_contract and redefines generic methods to fit compassion specific needs. Most importantly, it adds the type field which can be extended to add different types of contracts.
- [CHANGE] module sponsorship_compassion adds the sponsorship type to the contracts and contains only specific things related to sponsorships.

TODO : We still have to split functionality that is specific to switzerland only (like num_pol_ga or things related to bvr) into another module which would be defined in compassion-switzerland repository.

Note : Review and Merge #83 before to remove SDS tracking from this review...


ecino:
Review is still needed :innocent:

contract_compassion/model/contract_group.py
```diff
@@ -0,0 +1,256 @@
...
+        count_groups = str(self.search(
+            cr, uid, [('partner_id', '=', partner.id)], context=context,
+            count=True))
+        result += '0' * (5 - len(count_groups)) + count_groups
+        # Type '0' = Sponsorship
+        result += '0'
+        result += '0' * 4
+
+        if is_lsv:
+            result = '004874969' + result[9:]
```
CSester: This shouldn't be barcoded. 004874969 is the lsv prefix for compassion account? The bank account should hold this, and if nothing is available, it's time to add it.

contract_compassion/wizard/__init__.py
```diff
@@ -0,0 +1,15 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (C) 2014 Compassion CH (http://www.compassion.ch)
+#    Releasing children from poverty in Jesus' name
+#    @author: Cyril Sester <csester@compassion.ch>
```
CSester: I didn't remember to have coded this. Worse, if I blame the code it shows me that DavidZIVI committed this

ecino:
Is this review complete ? If yes, I will close it and open another which contains the fixes (but not all commits of DavidZIVI for the split).

ecino:
Old thread not needed anymore.

PR #86, opened 2015-05-29
1.5 devel

steveferry:
Following features added:
- Source model name in generated invoices
- Bank account and mandate tracking
- Date end for delegate
- Cron used to check delegate date expiration or beginning


child_compassion/model/child_compassion.py
```diff
@@ -357,6 +359,32 @@ def _create_empty_case_study(self, cr, uid, child, context=None):
             return child_prop_obj.create(cr, uid, vals, context)
         return True
 
+    def update_delegate(self, cr, uid, context=None):
+        obj_children = self.pool.get('compassion.child')
```
ecino: Since you are in the compassion.child object, you don't need to have obj_children because, self = obj_children.

child_compassion/model/child_compassion.py
```diff
@@ -357,6 +359,32 @@ def _create_empty_case_study(self, cr, uid, child, context=None):
...
+        obj_undelegate_wizard = self.pool.get('undelegate.child.wizard')
+
+        child_ids = obj_children.search(cr, uid, [], context=context)
+        child_ids_to_delegate = []
+        child_ids_to_undelegate = []
+
+        for child in obj_children.browse(cr, uid, child_ids, context=context):
+            if child.date_delegation:
+                if datetime.strptime(child.date_delegation, DF) \
+                   <= datetime.today():
```
ecino: you can add the condition that child.state must be in a valid state (same as the one tested in delegate_wizard).

child_compassion/view/child_compassion_view.xml
```diff
@@ -127,7 +127,7 @@
                     <field name="state"/>
                     <field name="delegated_to"/>
                     <field name="date_delegation"/>
-                    <field name="delegated_comment"/>
```
ecino: You removed the comment field. Add it again.

child_compassion/wizard/delegate_child_wizard.py
```diff
@@ -38,13 +39,18 @@ def _default_child_ids(self, cr, uid, context):
         'partner': fields.many2one(
             'res.partner', string=_('Partner'), required=True),
         'comment': fields.text(_('Comment'), required=True),
+        'date_delegation': fields.date(_('Delegation\'s beginning'),
+                                       required=True),
+        'date_end_delegation': fields.date(_('Delegation\'s end'),
+                                           required=False),
```
ecino: required=False is not necessary, it is implicit.

child_compassion/wizard/delegate_child_wizard.py
```diff
@@ -52,11 +58,22 @@ def delegate(self, cr, uid, ids, context=None):
         child_ids = self._default_child_ids(cr, uid, context)
         child_obj = self.pool.get('compassion.child')
 
+        if wizard.date_end_delegation:
+            if datetime.strptime(wizard.date_delegation, DF) > \
+               datetime.strptime(wizard.date_end_delegation, DF):
+                raise orm.except_orm("Invalid value", "End date must "
```
ecino: Put the error message inside _() function, so that it can be translated in French if needed.

partner_compassion/__openerp__.py
```diff
@@ -45,7 +45,7 @@
     'author': 'Compassion CH',
     'website': 'http://www.compassion.ch',
     'depends': ['base', 'partner_firstname', 'base_location',
-                'email_template'],
+                'email_template', 'contract_compassion'],
```
ecino: Why did you add dependency to module contract_compassion ? I think it is a bad idea because it can add a cyclic dependency...

partner_compassion/model/account_banking_mandate_compassion.py
```diff
@@ -0,0 +1,111 @@
...
+        return result
+
+    def validate(self, cr, uid, ids, context=None):
+        """Override function to notify validation in a message on partner feed
+        """
+
+        obj_mandate = self.pool['account.banking.mandate']
+
+        for mandate in obj_mandate.browse(cr, uid, ids, context=context):
+            mandate.partner_id.pool.get('mail.thread') \
```
ecino: Better to have a variable declared before the loop :  `thread_obj = self.pool.get('mail.thread')`

partner_compassion/model/partner_bank_compassion.py
```diff
@@ -0,0 +1,54 @@
...
+
+    _inherit = 'res.partner.bank'
+
+    def create(self, cr, uid, data, context):
+        """Override function to notify creation in a message
+        """
+        obj_partner = self.pool['res.partner']
+        part = obj_partner.browse(cr, uid, data['partner_id'], context=context)
+        part.pool.get('mail.thread').message_post(cr, uid, part.id,
+                                                  "<b>Account number: </b>" +
```
ecino: To avoid having many lines of code, you can return rewrite like this:
```
message_post(
    cr, uid, part.id, "...",
    context={...})```

partner_compassion/model/account_banking_mandate_compassion.py
ecino: You can still fix this one here.

ecino:
Thanks ! :+1:

PR #88, opened 2015-06-01
1.5 devel ferry

steveferry:
Add company name in analytic account name


crm_compassion/model/portal_wizard.py
```diff
@@ -30,6 +30,11 @@ def action_apply(self, cr, uid, ids, context=None):
                 cr, uid, [('name', '=', user.partner_id.name)],
                 context=ctx)
             partner_name = user.partner_id.name
+
+            if user.partner_id.parent_id:
```
ecino: I would do :
```
if user.partner_id and user.partner_id.parent_id
```
to avoid an error if partner_id is null.

ecino:
One more feature request : add a value in created analytic account :
```
'use_timesheets': True
```

ecino:
:+1:

PR #99, opened 2015-07-17
Port partner compassion

steveferry:


ecino:
So the travis tests are not working in pull requests, but they succeed when we mege.. not handy for viewing errors before merging... From last pull request :

./hr_planning/model/hr_planning_day.py:12:1: F401 '_' imported but unused
./hr_planning/model/resource_calendar.py:38:37: F821 undefined name 'calendar'

You can fix those issues in this PR.

partner_compassion/model/partner_compassion.py
```diff
@@ -8,79 +8,76 @@
...
+    church_id = fields.Many2one(
+        'res.partner', 'Church', domain=[('is_church', '=', True)])
+    church_unlinked = fields.Char(
+        "Church (N/A)",
+        help=_("Use this field if the church of the partner"
+               " can not correctly be determined and linked."))
+    deathdate = fields.Date('Death date')
+    birthdate = fields.Date('Birthdate')
+    lang = fields.Selection(
+        string='Language', compute='_lang_get', required=True,
```
ecino: lang is not a computed field, it is a function for returning the selection values.
so '_lang_get' should be the first parameter (not compute parameter).

partner_compassion/model/partner_compassion.py
```diff
@@ -8,79 +8,76 @@
...
+        help="If the selected language is loaded in the system, all "
+        "documents related to this contact will be printed in this "
+        "language. If not, it will be English.")
+
+    ##########################################################################
+    #                             FIELDS METHODS                             #
+    ##########################################################################
+
+    @api.multi
```
ecino: not a compute method, should be api.model and should return the value instead of self.lang =...

sponsorship_compassion/model/account_banking_mandate.py
```diff
@@ -9,23 +9,23 @@
...
-    def validate(self, cr, uid, ids, context=None):
+    @api.multi
+    def validate(self):
         """Validate LSV/DD Contracts when mandate is validated."""
-        if super(account_mandate, self).validate(cr, uid, ids, context):
+        if super(account_mandate, self).validate():
             wf_service = netsvc.LocalService('workflow')
-            for mandate in self.browse(cr, uid, ids, context):
+            for mandate in self:
                 contract_ids = self.pool.get('recurring.contract').search(
```
ecino: would be better to use self.env['recurring.contract']. Then search method will return contracts recordset.

ecino:
:+1:

PR #100, opened 2015-07-17
Port contract compassion

ecino:
- Tests are missing
- Some functionalities were moved to or from other modules, therefore the changes in __unported__


steveferry:
:+1:

PR #101, opened 2015-07-21
Port message center

ecino:
Don't merge until `sponsorship_compassion` is ported and working, as this module does not work without it.

Port message_center_compassion.


ecino:
PR are now made on master branch.

PR #102, opened 2015-07-21
Port message center

ecino:
Don't merge until `sponsorship_compassion` is ported and working, as this module does not work without it.

Port message_center_compassion.


message_center_compassion/model/child_compassion.py
```diff
@@ -72,144 +77,134 @@ def allocate(self, cr, uid, args, context=None):
...
                 if contract.state in ('waiting', 'active', 'mandate'):
                     # Terminate contract and mark the departure
-                    ctx = context.copy()
-                    ctx['default_type'] = 'S'
-                    self.pool.get('recurring.contract').write(
-                        cr, uid, contract.id, {
-                            'end_reason': '1',  # Child departure
-                            'end_date': datetime.today().strftime(DF),
-                            'gmc_state': 'depart'}, ctx)
+                    self.env['recurring.contract'].write({
```
steveferry: `contract.write(...)`

message_center_compassion/model/child_compassion.py
```diff
@@ -72,144 +77,134 @@ def allocate(self, cr, uid, args, context=None):
...
+        referencing it."""
+        child_id = self.create(vals).id
+        # Link existing messages to the child
+        messages = self.env['gmc.message.pool'].search([
+            ('incoming_key', '=', vals.get('code')),
+            ('object_id', '=', 0)])
+        messages.write({
+            'object_id': child_id,
+            'child_id': child_id})
+        return child_id
```
steveferry: We have to agree on what we return, either the `id` or the `Record`

message_center_compassion/model/child_properties.py
```diff
@@ -8,21 +8,17 @@
...
-    def attach_pictures(self, cr, uid, ids, pictures_id, context=None):
-        res = super(child_properties, self).attach_pictures(
-            cr, uid, ids, pictures_id, context)
-        if not isinstance(ids, list):
-            ids = [ids]
-        property = self.browse(cr, uid, ids, context)[0]
-        for contract in property.child_id.sponsorship_ids:
+    def attach_pictures(self, pictures_id):
+        res = super(child_properties, self).attach_pictures(pictures_id)
+        for contract in self.child_id.sponsorship_ids:
```
steveferry: May be prettier to filter `sponsorship_ids` by contract state `('waiting', 'active', 'mandate')` and then call `new_biennal()` on the returned RecordSet

message_center_compassion/model/contracts.py
```diff
@@ -0,0 +1,250 @@
...
+        message_obj = self.env['gmc.message.pool']
+        action_id = self.get_action_id('CreateGift')
+
+        for invoice_line in invoice.invoice_line:
+            message_vals = {
+                'action_id': action_id,
+                'date': invoice.date_invoice,
+            }
+            contract = invoice_line.contract_id
+            if not contract:
```
steveferry: `if contract` and do the rest

message_center_compassion/model/contracts.py
```diff
@@ -0,0 +1,250 @@
...
+    @api.multi
+    def contract_active(self):
+        """ Create messages to GMC when new sponsorship is activated. """
+        super(recurring_contract, self).contract_active()
+        message_obj = self.env['gmc.message.pool']
+        action_id = 0
+        message_vals = dict()
+
+        for contract in self:
+            if 'S' in contract.type:
```
steveferry: `contract.type == 'S'`

message_center_compassion/model/contracts.py
```diff
@@ -0,0 +1,250 @@
...
+        res = super(recurring_contract, self).contract_terminated()
+        if res:
+            message_obj = self.env['gmc.message.pool'].with_context(
+                force_send=True)
+            action_id = self.get_action_id('CancelCommitment')
+            message_vals = {'action_id': action_id}
+
+            for contract in self:
+                end_reason = int(contract.end_reason)
+                if 'S' in contract.type:
```
steveferry: Same as above

message_center_compassion/model/contracts.py
```diff
@@ -0,0 +1,250 @@
...
+                        self._change_gift_dates(messages)
+                        messages.process_messages()
+                    else:
+                        # Abort gifts and clean invoices
+                        messages.unlink()
+                        self.clean_invoices_paid(gifts=True)
+
+                # Contract must have child and not terminated by
+                # partner move -> CancelCommitment message
+                if 'S' in contract.type and end_reason != 4:
```
steveferry: Same as above

message_center_compassion/model/contracts.py
```diff
@@ -0,0 +1,250 @@
...
+        messages = self.env['gmc.message.pool'].search([
+            ('invoice_line_id', 'in', invoice_line_ids),
+            ('state', 'in', ['new', 'failure']),
+            ('action_id', '!=', cancel_id)])
+        messages.unlink()
+
+    def _change_gift_dates(self, messages):
+        """ Put gift invoices today if they are in the future. """
+        today = datetime.today()
+        for message in messages:
```
steveferry: `messages_update = messages.filtered(lambda m: datetime.strptime(m.invoice_line_id.invoice_id.date_invoice, DF) > today)`
`invoices_update = messages_update.mapped('invoice_line_id.invoice_id')`
`invoices_update.write({'date_invoice': today.strftime(DF)})`
`message_update.write({'date': today.strftime(DF + ' %H:%M:%S')})`

message_center_compassion/model/gmc_action.py
```diff
@@ -0,0 +1,92 @@
...
+            raise ValidationError(
+                _("Invalid action (%s, %s, %s).") % (
+                    self.direction, self.model, self.action_type))
+
+    def get_action_id(self, name):
+        """ Returns the id of the action given its name. """
+        action_id = self.search([('name', '=', name)], limit=1)
+        if action_id:
+            return action_id[0].id
+        else:
```
steveferry: Not needed

message_center_compassion/model/gmc_message.py
```diff
@@ -0,0 +1,425 @@
...
+    def process_update_messages(self):
+        gmc_action_ids = self.env['gmc.action'].search(
+            [('type', '=', 'update'), ('direction', '=', 'in')]).ids
+        gmc_update_messages = self.search(
+            [('action_id', 'in', gmc_action_ids)])
+        gmc_update_messages.process_messages()
+
+    @api.multi
+    def force_success(self):
+        for message in self:
```
steveferry: `self.write({ 'state': 'sucess', 'failure_reason': False})`
`self.filtered(lambda m: m.name == 'CreateGift').write({'state': 'fondue'})`

message_center_compassion/model/invoice_line.py
```diff
@@ -0,0 +1,38 @@
...
+        if self.product_id.gmc_name == 'ProjectGift':
+            need_key = need_key[:5]
+        self.need_key = need_key
+
+    @api.one
+    def _get_instructions(self):
+        lang = self.partner_id.lang
+        product = self.env['product.product'].with_context(
+            lang=lang).browse(self.product_id.id)
+        if self.product_id.name == self.name or \
```
steveferry: `if ... != ... or ... != ...` and then: `self.gift_instructions = self.name`

message_center_compassion/model/res_partner.py
```diff
@@ -0,0 +1,64 @@
...
+        """If partner has active contracts, UPSERT Constituent in GMC."""
+        for partner in self:
+            contract_count = self.env['recurring.contract'].search_count([
+                ('correspondant_id', '=', partner.id),
+                ('state', 'not in', ('terminated', 'cancelled'))])
+            if contract_count:
+                # UpsertConstituent Message
+                action_id = self.env['gmc.action'].search(
+                    [('name', '=', 'UpsertConstituent')],
+                    limit=1)[0].id
```
steveferry: Won't it try to access to the `0` index of the record? Are you sure it's a recordset?

message_center_compassion/tests/test_messages.py
```diff
@@ -252,13 +224,13 @@ def test_gmc_scenario(self):
         """This is the test scenario detailed in file
         ..data/test_scenario.docx
         """
-        context = {'test_mode': True}
         # Simulate GMC Allocation of 4 new children
         child_keys = ["PE3760148", "IO6790210"]  # , "UG8320012", "UG8350016"]
-        child_ids = self._allocate_new_children(child_keys)
+        children = self._allocate_new_children(child_keys)
+        child_ids = children.ids
```
steveferry: Not used

steveferry:
Missing a few `@api` decorators a little bit everywhere

message_center_compassion/model/child_compassion.py
ecino: In this case, the method is only called from this file and uses only the id outside. But it's true the standard is more to return the record. Do you think we should always return Records or is it ok to have some exceptions ?

message_center_compassion/model/res_partner.py
ecino: Search method always returns a recordset, even if only one record matches the search criteria. Maybe we can directly call .id if the recordset contains only one record, but I am not sure that's better to do it. However I will replace this by the method self.env.ref() which directly finds a record given the xml_id.

ecino:
api.multi is not mandatory when functions are called only from v8 code and not from views, so I don't put it when not needed. Other decorators should not miss, if you see some, can you please point at them ?

ecino:
Are you ok with this one ?

message_center_compassion/model/child_compassion.py
steveferry: Ok

message_center_compassion/model/res_partner.py
steveferry: Did you remove the `[0]`, after our discussion with the `search` method and `limit=1`?

ecino: No I forgot it, I will do this. I am also fixing a lot of other issues and will tell you when ready. Maybe another review will be needed before merging this one.

steveferry: Ok, no problem

ecino:
Ok, I finished the corrections. Can you do a quick review of the last changes (you may want to look at the last commits) ? Thanks !

sponsorship_compassion/model/res_partner.py
```diff
@@ -119,9 +119,7 @@ def create_contract(self):
             'view_mode': 'form',
             'res_model': 'recurring.contract',
             'target': 'current',
-            'context': self.env.context.with_context({
-                'default_partner_id': self,
-                'default_type': 'S', })
+            'context': context
```
steveferry: Set `context` key with the dictionnary used in `with_context` above, and remove line 111

ecino: Won't we lose other values in current context ? But I don't know if Odoo will automatically keep context and add the values in dictionary or start fresh...

steveferry: Ah ok, I don't know either... So keep it like that

message_center_compassion/model/gmc_action.py
```diff
@@ -0,0 +1,91 @@
...
+        if not valid:
+            raise ValidationError(
+                _("Invalid action (%s, %s, %s).") % (
+                    self.direction, self.model, self.action_type))
+
+    def get_action_id(self, name):
+        """ Returns the id of the action given its name. """
+        action_id = self.search([('name', '=', name)], limit=1)
+        if action_id:
+            return action_id[0].id
```
steveferry: With the `search(limit=1)` you don't need to use the index access `[0]` on the returned record

message_center_compassion/model/gmc_message.py
```diff
@@ -0,0 +1,433 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class gmc_message_pool_process(models.TransientModel):
+    _name = 'gmc.message.pool.process'
+
+    @api.multi
+    def process_messages(self):
+        active_ids = self.env.context.get('active_ids')
```
steveferry: Are we sure that `active_ids` is in the context? Wouldn't be safer to use get like that: `self.env.context.get('active_ids',[])` ?

ecino: Normally it should always be set (called from a More action). But I can still modify it in case...

steveferry:
:+1:

PR #107, opened 2015-07-28
Test of sponsorship compassion

albertsh:
- add of four scenarios
- test of the sponsorship type contract
- test of the creation of more than one contract for the same partner


ecino:
Pull request must be for branch 7.0 instead of master.

PR #108, opened 2015-08-03
Add of the test of sponsorship_compassion

albertsh:
- add of four scenarios
- test of the sponsorship type contract
- test of the creation of more than one contract for the same partner

### DIFF
--- a/contract_compassion/__openerp__.py
+++ b/contract_compassion/__openerp__.py
@@ -38,7 +38,8 @@ Compassion contract module.
     'author': 'Compassion CH',
     'website': 'http://www.compassion.ch',
     'depends': ['recurring_contract', 'account_banking_mandate',
-                'child_compassion'],
+                'child_compassion', 'account_analytic_compassion', 
+                'account_accountant'],
     'data': [
         'data/install.xml',
         'view/end_contract_wizard_view.xml',
--- a/contract_compassion/__openerp__.py
+++ b/contract_compassion/__openerp__.py
@@ -38,7 +38,7 @@ Compassion contract module.
     'author': 'Compassion CH',
     'website': 'http://www.compassion.ch',
     'depends': ['recurring_contract', 'account_banking_mandate',
-                'child_compassion', 'account_analytic_compassion', 
+                'child_compassion', 'account_analytic_compassion',
                 'account_accountant'],
     'data': [
         'data/install.xml',
--- a/contract_compassion/model/contract_group.py
+++ b/contract_compassion/model/contract_group.py
@@ -10,7 +10,7 @@
 ##############################################################################
 
 import time
-import pdb
+
 from openerp.osv import orm, fields
 from openerp.tools.translate import _
 from openerp.tools import mod10r
--- a/contract_compassion/model/contract_group.py
+++ b/contract_compassion/model/contract_group.py
@@ -10,7 +10,7 @@
 ##############################################################################
 
 import time
-
+import pdb
 from openerp.osv import orm, fields
 from openerp.tools.translate import _
 from openerp.tools import mod10r
--- a/contract_compassion/tests/test_contract_compassion.py
+++ b/contract_compassion/tests/test_contract_compassion.py
@@ -13,8 +13,8 @@ from openerp.tests import common
 from datetime import datetime
 from openerp import netsvc
 from openerp.tools import DEFAULT_SERVER_DATE_FORMAT as DF
+from dateutil.relativedelta import relativedelta
 import logging
-import pdb
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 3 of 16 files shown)


sponsorship_compassion/tests/test_sponsorship_compassion.py
```diff
@@ -0,0 +1,459 @@
...
+            'property_account_payable': property_account_payable,
+            'notification_email_send': 'comment',
+        })
+        # Retrieve a payment term
+        payment_term_obj = self.registry('account.payment.term')
+        self.payment_term_id = payment_term_obj.search(self.cr, self.uid, [
+            ('name', '=', '15 Days')
+        ])[0]
+        product_obj = self.registry('product.product')
+        product_obj.write(self.cr, self.uid, 90, {
```
ecino: You should search for the product you are looking for instead on writing on a given ID, as you are not guaranteed that the product will have the same ID on all test databases...

contract_compassion/__openerp__.py
```diff
@@ -38,7 +38,8 @@
     'author': 'Compassion CH',
     'website': 'http://www.compassion.ch',
     'depends': ['recurring_contract', 'account_banking_mandate',
-                'child_compassion'],
+                'child_compassion', 'account_analytic_compassion', 
+                'account_accountant'],
```
ecino: No, we don't need this dependency. Please remove account_accountant.

contract_compassion/model/contract_group.py
```diff
@@ -10,7 +10,7 @@
 ##############################################################################
 
 import time
-
+import pdb
```
ecino: Remove pdb

contract_compassion/tests/test_contract_compassion.py
```diff
@@ -14,6 +14,7 @@
 from openerp import netsvc
 from openerp.tools import DEFAULT_SERVER_DATE_FORMAT as DF
 import logging
+import pdb
```
ecino: Remove pdb

contract_compassion/tests/test_contract_compassion.py
```diff
@@ -84,6 +85,19 @@ def test_contract_compassion_first_scenario(self):
             In this test we are testing states changement of a contract and if
             the old invoice are well cancelled when we pay one invoice.
         """
+        current_year = str(int(datetime.today().strftime(DF)[:4]) + 1)
```
ecino: would be better to call this variable `next_year`

contract_compassion/tests/test_contract_compassion.py
```diff
@@ -84,6 +85,19 @@ def test_contract_compassion_first_scenario(self):
             In this test we are testing states changement of a contract and if
             the old invoice are well cancelled when we pay one invoice.
         """
+        current_year = str(int(datetime.today().strftime(DF)[:4]) + 1)        
+        fiscal_year_id = self.registry('account.fiscalyear').create(
+            self.cr, self.uid, {
+                'name': current_year,
+                'code': current_year,
+                'date_start': str(datetime(
```
ecino: It is not good to convert a date using str() method. You should use .strptime() method on date objects. Look for python documentation if needed.

contract_compassion/tests/test_contract_compassion.py
```diff
@@ -84,6 +85,19 @@ def test_contract_compassion_first_scenario(self):
             In this test we are testing states changement of a contract and if
             the old invoice are well cancelled when we pay one invoice.
         """
+        current_year = str(int(datetime.today().strftime(DF)[:4]) + 1)        
+        fiscal_year_id = self.registry('account.fiscalyear').create(
+            self.cr, self.uid, {
+                'name': current_year,
+                'code': current_year,
+                'date_start': str(datetime(
+                    int(datetime.today().strftime(DF)[:4]) + 1, 1, 1)),
```
ecino: Why not using your `next_year` variable here ? You don't need to do the same computation again...

crm_compassion/model/crm_lead.py
```diff
@@ -11,6 +11,7 @@
 
 from openerp.osv import orm, fields
 from openerp.tools.translate import _
+import pdb
```
ecino: Remove pdb

crm_compassion/tests/test_crm_compassion.py
```diff
@@ -0,0 +1,282 @@
...
+#    The licence is in the file __openerp__.py
+#
+##############################################################################
+
+from openerp.tests import common
+from datetime import datetime
+from openerp import netsvc
+from openerp.tools import DEFAULT_SERVER_DATE_FORMAT as DF
+import logging
+import pdb
```
ecino: Remove pdb

crm_compassion/tests/test_crm_compassion.py
```diff
@@ -0,0 +1,282 @@
...
+logger = logging.getLogger(__name__)
+
+
+class test_crm_compassion(common.TransactionCase):
```
ecino: It seems that most of your tests (recurring_contract, contract_compassion, sponsorship_compassion, ...) have the same setup method and needs. Instead of copying again and again the same code, use inheritance !

So instead of inheriting the class `common.TransactionCase` you can inherit something like `recurring_contract.test_recurring_contract`.

In that way, when you call `super` inside the setUp method, it will call the setUp method of the test in recurring_contract. You will also be able to call all private methods like `_create_contract()`, `_create_group()`, etc...

Please consider this change.

sponsorship_compassion/tests/test_sponsorship_compassion.py
```diff
@@ -0,0 +1,459 @@
...
+
+from openerp.tests import common
+from datetime import datetime
+from openerp import netsvc
+from openerp.tools import DEFAULT_SERVER_DATE_FORMAT as DF
+import logging
+logger = logging.getLogger(__name__)
+
+
+class test_sponsorship_compassion(common.TransactionCase):
```
ecino: Use inheritance

ecino:
:+1:

PR #111, opened 2015-08-06
Remove fiscal year creation from sponsorship_compassion

albertsh:
Remove fiscal year creation from sponsorship_compassion


ecino:
Bad fix.

PR #113, opened 2015-08-18
8.0 tests for compassion-modules

albertsh:
- add of a bridge test between contract_compassion and sponsorship_compassion
- change to 8.0 version of the tests
- there is now 1 method (instead of 2 before) to create all types of contracts


ecino:
Please do the pull request on branch master instead of 8.0

PR #115, opened 2015-08-20
[fix] auto description and date of picture

### LINKED ISSUE #97: [8.0] child_compassion javascript is broken
field_on_change methods are never called in javascript

ndtran:
fixes #97


ecino:
:+1:

PR #117, opened 2015-08-21
8.0 test for crm_compassion

albertsh:
- change test_crm_compassion on 8.0 version


ecino:
PEP8 errors. Please verify more carefully before sending pull request...

contract_compassion/tests/test_base_module.py
```diff
@@ -20,8 +20,8 @@ class test_base_module(test_base_contract):
 
     def setUp(self):
         # Retrieve of income account
-        account_obj = self.env['account.account']
-        property_account_income = account_obj.search([
+        super(test_base_module, self).setUp()
```
ecino: Right before you told me you would not call super from here and now you add it. Please explain.

ecino:
Remove all pdb

contract_compassion/tests/test_base_module.py
albertsh: if i don't add it, i have an error saying that test_sponsorship_compassion or test_crm_compassion object have no env attribute when i do a simple self.env[something]

albertsh: oh my bad i didn't understood the question well, i thought the question was why i haven't inherit from test_recurring_contract

ecino:
:+1:

PR #121, opened 2015-08-28
compassion-modules - add features

steveferry:
- Add channel selection in sub_sponsorship wizard
- Gift can't be paid if sponsorship is not active
- Set False to all delegated fields for a child on creation of a sponsorship


ecino:
:+1:

PR #122, opened 2015-08-28
put test_message to 8.0 version

albertsh:
- some corrections done in the module's code:
 - write can take only one argument
 - self.env.with_context doesn't work
 - add of paranthesis


contract_compassion/tests/test_base_module.py
```diff
@@ -21,10 +21,10 @@ class test_base_module(test_base_contract):
     def setUp(self):
         # Retrieve of income account
         super(test_base_module, self).setUp()
-        property_account_income = self.env['account.account'].search([
+        self.property_account_income = self.env['account.account'].search([
```
ecino: Why changing this ? I see no use...

albertsh: because we use it in a method latter (_create_gift)

ecino: Ok,

ecino:
:+1:

PR #123, opened 2015-08-31
correction of contract_origin

albertsh:
- set_name in place of name_get

### DIFF
--- a/contract_compassion/model/contracts.py
+++ b/contract_compassion/model/contracts.py
@@ -416,7 +416,7 @@ class recurring_contract(models.Model):
                 # with new contract information and validate them
                 cancel_invoices = invoice_obj.search([
                     ('state', '=', 'cancel'),
-                    ('id', 'in', invoices_canceled.ids)])
+                    ('id', 'in', list(invoices_canceled))])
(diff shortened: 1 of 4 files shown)


contract_compassion/model/contracts.py
```diff
@@ -416,7 +416,7 @@ def _reset_open_invoices(self):
                 # with new contract information and validate them
                 cancel_invoices = invoice_obj.search([
                     ('state', '=', 'cancel'),
-                    ('id', 'in', invoices_canceled.ids)])
+                    ('id', 'in', list(invoices_canceled))])
```
ecino: This is a bad change, revert it.

crm_compassion/model/contract_origin.py
```diff
@@ -23,9 +23,9 @@ class contract_origin(models.Model):
     def _set_name(self):
         if self.type == 'event':
             name = self.event_id.full_name
+            self.name = name
```
ecino: You can remove this line and directly do `self.name = self.event_id.full_name`

ecino:
:+1:

PR #129, opened 2015-10-07
Fix allows the user to edit the child's description

loikki:


ecino:
I think this is not needed. Descriptions are editable in child object, not in case study object (child.properties)

ecino:
After verification, :+1:

PR #133, opened 2015-11-18
Importation letters

loikki:
Finalize the last step for the letters importation (convert import_letter_line to sponsorship_correspondence) and allows to read a page in the wrong direction and a multipage TIFF.

Two small fixes have been made too. Now the QR code's position is stored in ir.config_parameter (need to have it before finding the template) and the module tempfile is used almost everywhere in order to avoid tempfile not removed


ecino:
./sbc_compassion/models/sponsorship_correspondence.py:180:29: E126 continuation line over-indented for hanging indent

./sbc_compassion/models/sponsorship_correspondence.py:194:56: E126 continuation line over-indented for hanging indent

./sbc_compassion/models/sponsorship_correspondence.py:207:62: E126 continuation line over-indented for hanging indent

contract_compassion/tests/test_contract_compassion.py
```diff
@@ -41,8 +41,8 @@ def test_contract_compassion_first_scenario(self):
         """
         # Creation of a the next fiscal year
         next_year = (datetime.strptime(
-            (datetime.today().strftime(DF)), DF)
-            + relativedelta(years=+1)).year
+            (datetime.today().strftime(DF)), DF) +
+                     relativedelta(years=+1)).year
```
ecino: not properly indented

ecino:
You can add a `demo` folder in the module structure, and put a `template_demo.xml` file inside it. In the odoo manifest file, you can include the demo file in the `demo` section. 

Then, you can move all records related to test template, inside that `template_demo.xml` file.

This way, the template will only be loaded in demo mode (can ben changed in the module page).

sbc_compassion/models/correspondence_template.py
```diff
@@ -67,23 +67,27 @@ class CorrespondenceTemplate(models.Model):
         help='Y Position of the upper-right corner of the bluesquare '
              'in pixels')
     qrcode_x_min = fields.Integer(
+        compute="_compute_size",
```
ecino: Is it useful to keep this field if the `qr_code` area will always be in the same place ? Is it useful to display the information in the view ? I feel that we can remove these 4 fields along with the `_compute_size` method.

sbc_compassion/models/correspondence_template.py
```diff
@@ -168,11 +172,72 @@ def _set_image(self):
...
+                            _("The pattern could not be detected in given "
+                              "template image."))
+                    # find center of the pattern
+                    pattern_center = pr.keyPointCenter(pattern_keypoints)
+                    template.pattern_center_x = pattern_center[0]
+                    template.pattern_center_y = pattern_center[1]
+                    # blue corner detection
+                    bluecorner = bcf.BlueCornerFinder(
+                        img).getIndices()
+                    template.bluesquare_x = bluecorner[0]
```
ecino: Since `bluesquare_x` and `bluesquare_y` are computed in this method, they must also be attached to the method by adding `compute='_coompute_template_keypoints', store=True` to the field definition.

sbc_compassion/models/correspondence_template.py
```diff
@@ -168,11 +172,72 @@ def _set_image(self):
...
+    def _compute_size(self):
+        for template in self:
+            if template.template_image:
+                with tempfile.NamedTemporaryFile() as template_file:
+                    template_file.write(base64.b64decode(
+                        template.with_context(bin_size=False).template_image))
+                    template_file.flush()
+                    # Find the pattern inside the template image
+                    img = cv2.imread(template_file.name)
+                    template.page_height, template.page_width = img.shape[:2]
```
ecino: Add compute method for `page_height` and `page_width` fields.

sbc_compassion/models/correspondence_template.py
```diff
@@ -168,11 +172,72 @@ def _set_image(self):
...
+        for template in self:
+            if (template.template_image and template.pattern_image and
+                    _verify_template(template)):
+                with tempfile.NamedTemporaryFile() as template_file:
+                    template_file.write(base64.b64decode(
+                        template.with_context(bin_size=False).template_image))
+                    template_file.flush()
+                    # Find the pattern inside the template image
+                    img = cv2.imread(template_file.name)
+                    template.page_height, template.page_width = img.shape[:2]
```
ecino: Add to `page_height` and `page_width` fields the `compute` method.

sbc_compassion/models/import_letter_line.py
```diff
@@ -27,6 +27,8 @@ class ImportLetterLine(models.Model):
                                      compute='_set_sponsorship_id')
     sponsorship_status = fields.Boolean(compute='_set_sponsorship_id',
                                         readonly=True)
+    person = fields.Boolean(compute='_set_sponsorship_id',
```
ecino: What is this field for ? Isn't it the same as `sponsorship_status` ?
You could rename `_set_sponsorship_id` to `_find_sponsorship` since it is used for all fields related to sponsorship.

sbc_compassion/models/import_letter_line.py
```diff
@@ -40,8 +42,8 @@ class ImportLetterLine(models.Model):
 
     status = fields.Selection([
         ("lang", _("Error in Language")),
-        ("sponsor", _("Error in Sponsorship")),
-        ("encourager", _("Sponsorship not found")),
+        ("sponsorship", _("Sponsorship not Found")),
```
ecino: even if the first value is what is stored in database, you can give more precise names (like `no_sponsorship`)

sbc_compassion/models/import_letter_line.py
```diff
@@ -79,10 +81,25 @@ def _set_sponsorship_id(self):
                     ('child_id.code', '=', line.child_code),
                     ('partner_codega', '=', line.partner_codega),
                     ('is_active', '=', True)], order='end_date desc', limit=1)
+                if len(line.sponsorship_id) == 0:
```
ecino: you can simply do `if not line.sponsorship_id`

sbc_compassion/models/import_letter_line.py
```diff
@@ -79,10 +81,25 @@ def _set_sponsorship_id(self):
...
+                    line.sponsorship_id = line.env[
+                        'recurring.contract'].search(
+                            [('child_id.code', '=', line.child_code),
+                             ('partner_codega', '=', line.partner_codega)],
+                            order='end_date desc', limit=1)
```
ecino: Active sponsorships won't have end_dates so they may be put at the end of the list. It is safer to use `start_date` desc

sbc_compassion/models/import_letter_line.py
```diff
@@ -79,10 +81,25 @@ def _set_sponsorship_id(self):
...
                 if len(line.sponsorship_id) == 1:
```
ecino: `if line.sponsorship_id`

sbc_compassion/models/import_letter_line.py
```diff
@@ -79,10 +81,25 @@ def _set_sponsorship_id(self):
...
-                    line.sponsorship_status = None
+                    line.sponsorship_status = True
+                    line.person = True
                 else:
                     line.sponsorship_status = False
+                    if not (len(line.env['recurring.contract'].search([
```
ecino: You don't need to test len() > 0. An empty list will always evaluate to `False`

sbc_compassion/models/import_letter_line.py
```diff
@@ -79,10 +81,25 @@ def _set_sponsorship_id(self):
...
+                            ('child_id.code', '=', line.child_code)])) > 0 or
+                            len(line.env['recurring.contract'].search([
```
ecino: len() not needed.

sbc_compassion/models/import_letter_line.py
```diff
@@ -92,3 +109,23 @@ def _set_name(self):
                 line.name = str(
                     line.sponsorship_id.partner_codega) + " - " + str(
                         line.child_code)
+
+    @api.multi
+    def save_lines(self, mandatory_review=False):
+        """
+        """
```
ecino: Comment missing

sbc_compassion/models/import_letter_line.py
```diff
@@ -92,3 +109,23 @@ def _set_name(self):
...
+        ids = []
+        for letter in self:
```
ecino: Please rename `letter` to `line` to avoid confusion.

sbc_compassion/models/import_letter_line.py
```diff
@@ -92,3 +109,23 @@ def _set_name(self):
                 line.name = str(
                     line.sponsorship_id.partner_codega) + " - " + str(
                         line.child_code)
+
+    @api.multi
+    def save_lines(self, mandatory_review=False):
+        """
+        """
+        ids = []
```
ecino: `letter_datas` would be a better variable name.

sbc_compassion/models/import_letter_line.py
```diff
@@ -92,3 +109,23 @@ def _set_name(self):
                 line.name = str(
                     line.sponsorship_id.partner_codega) + " - " + str(
                         line.child_code)
+
+    @api.multi
+    def save_lines(self, mandatory_review=False):
```
ecino: `get_letter_datas` would be more appropriate to what the function is doing (it doesn't save anything).

sbc_compassion/models/import_letter_line.py
```diff
@@ -92,3 +109,23 @@ def _set_name(self):
...
+        """
+        """
+        ids = []
+        for letter in self:
+            dict_ = {
```
ecino: Since `dict` is a reserved name in python, I think `vals` would be less similar.

sbc_compassion/models/import_letters_history.py
```diff
@@ -59,32 +60,33 @@ class ImportLettersHistory(models.Model):
...
                                   compute="_count_nber_letters")
-    # path where the zipfile are store
-    path = "/tmp/sbc_compassion/"
-    # list zip in the many2many
-    # use ';' in order to separate the files
-    list_zip = fields.Text("LIST_ZIP", readonly=True)
+    is_mandatory_review = fields.Boolean("Mandatory Review", default=False)
 
     data = fields.Many2many('ir.attachment', string=_("Add a file"))
     letters_line_ids = fields.Many2many('import.letter.line')
```
ecino: Please rename to `import_line_ids` to avoid confusion.

sbc_compassion/models/correspondence_template.py
loikki: It is a shortcut to this field (and, as the value is stored in %, avoid to compute it everytime).
I don't know if it is useful to have it visible, but I prefer to keep it inside the model.

sbc_compassion/models/import_letters_history.py
```diff
@@ -59,32 +60,33 @@ class ImportLettersHistory(models.Model):
...
-                check = False
-        if check and len(self.letters_line_ids) > 0:
-            self.state = "ready"
-        elif len(self.letters_ids) > 0:
-            self.state = "saved"
-        elif len(self.letters_line_ids) > 0:
-            self.state = "pending"
-        else:
-            self.state = "draft"
+        for letters in self:
```
ecino: I think this method would be more easy to read in `@api.one` mode.
You can then remove this `for` loop and replace `letters` by `self` directly.

sbc_compassion/models/import_letters_history.py
```diff
@@ -59,32 +60,33 @@ class ImportLettersHistory(models.Model):
...
+            if len(letters.letters_ids) > 0:
```
ecino: len() is not necessary in the test.

sbc_compassion/models/import_letters_history.py
```diff
@@ -59,32 +60,33 @@ class ImportLettersHistory(models.Model):
...
+                letters.state = "saved"
+            elif len(letters.letters_line_ids) > 0:
```
ecino: len() not necessary.

sbc_compassion/models/correspondence_template.py
loikki: I do not remember why, but I think I had some troubles with theses fields when it was with the method compute. I will try again later.

sbc_compassion/models/import_letter_line.py
loikki: It is for the status of the partner or the child (thus not a sponsorship), but I did not spend enough time to think about the name

sbc_compassion/models/import_letter_line.py
loikki: I add a 'no_' before all the fields, is it enough?

sbc_compassion/models/import_letters_history.py
```diff
@@ -192,121 +162,172 @@ def _run_analyze(self):
...
-                                if func.isPDF(absname):
-                                    raise exceptions.Warning(
-                                        _("PDF not implemented yet"))
-                                inst._analyze_attachment(absname)
-                        # delete all the tmp files
-                        # extracted data
-                        shutil.rmtree(path_zip)
-                        # zip file
-                        os.remove(inst.path + attachment.name)
+                        # create a temp file
```
ecino: Since this method is quite big and has a lot of indentation, you can remove the first loop `for inst in self` and begin with `self.ensure_one()` to be sure `self` contains only one record. Then you can directly use `self` instead of `inst`

sbc_compassion/models/import_letters_history.py
```diff
@@ -192,121 +162,172 @@ def _run_analyze(self):
...
                         raise exceptions.Warning(
                             'Still a file in a non-accepted format')
                 else:
                     raise exceptions.Warning(_('Two files are the same'))
             # remove all the files (now there are inside import_line_ids)
-            for attachment in inst.data:
-                attachment.unlink()
 
-    def _analyze_attachment(self, file_, data=None):
+    def _analyze_attachment(self, file_, filename):
```
ecino: Consider refactoring this method into smaller chunks to improve code readability (more than 100 lines for a function is usually too much).

sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -79,7 +79,7 @@ class SponsorshipCorrespondence(models.Model):
         readonly=True)
     # First spoken lang of partner
     original_language_id = fields.Many2one(
-        'res.lang.compassion', compute='_set_languages',
+        'res.lang.compassion', readonly=True,
```
ecino: What if someone enters manually a letter ? Will he be able to select the language ?

sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -98,8 +98,7 @@ class SponsorshipCorrespondence(models.Model):
     relationship = fields.Selection([
         ('Sponsor', _('Sponsor')),
         ('Encourager', _('Encourager'))], default='Sponsor')
-    mandatory_review = fields.Boolean(compute='_set_partner_review',
-                                      readonly=False, store=True)
+    mandatory_review = fields.Boolean(readonly=False, store=True)
```
ecino: readonly False and store True are useless, it is by default like this.

sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -172,38 +171,50 @@ def get_communication_types(self):
...
+    @api.multi
     @api.depends('sponsorship_id')
     def _set_name(self):
-        if self.sponsorship_id and self.communication_type_ids:
-            self.name = self.communication_type_ids[0].name + ' (' + \
-                self.sponsorship_id.partner_codega + " - " + \
-                self.child_id.code + ')'
-        else:
-            self.name = _('New correspondence')
+        for corr in self:
```
ecino: Choose either `correspondence` or `letter` for naming one record of `sponsorship.correspondence` and stick to that name in the whole code for consistency.

ecino:
One general remark : you have several models with similar names (import_letters_history, import_lines, sponsorship_correspondence, etc..) and you mix the variable names which is very confusing.

Please define some standard names for all models you have and use the same everywhere so that we know when reading `letter` should be a `sponsorship.correspondence` record and `import_line` should be a `import.letter.line` record.

sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -172,38 +171,50 @@ def get_communication_types(self):
...
+                                                       original_language_id
+                    else:
+                        corr.destination_language_id = corr.child_id\
+                                                           .project_id\
+                                                           .country_id\
+                                                           .spoken_langs_ids[0]
 
-    @api.depends('sponsorship_id')
+            if corr.direction == 'Beneficiary To Supporter':
+                # TODO: take into account the language sent
```
ecino: Is it still TODO ?

sbc_compassion/models/import_letters_history.py
loikki: The api.one will be deprecated in the next version, therefore I am trying to avoid it

sbc_compassion/models/sponsorship_correspondence.py
ecino: By the way I think that for Beneficiary To Supporter letters, the US will always send us the original language information, so we may not need to compute it.

sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -172,38 +171,50 @@ def get_communication_types(self):
...
+                # by the child's country
+                if corr.child_id.project_id.country_id.spoken_langs_ids:
+                    corr.original_language_id = corr.child_id.project_id\
+                                                             .country_id.\
+                                                             spoken_langs_ids[
+                                                                 0]
+                    if corr.correspondant_id.spoken_langs_ids:
+                        corr.destination_language_id = corr.correspondant_id\
```
ecino: Check first if the partner speaks also the `original_language_id`. If yes, set it to this lang, else, you can take his first spoken lang.

sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -172,38 +171,50 @@ def get_communication_types(self):
...
+                corr.name = corr.communication_type_ids[0].name + ' (' + \
+                            corr.sponsorship_id.partner_codega + " - " + \
+                            corr.child_id.code + ')'
+            else:
+                corr.name = _('New correspondence')
 
-    @api.depends('sponsorship_id', 'direction')
+    @api.depends('sponsorship_id', 'direction', 'original_language_id',
+                 'destination_language_id')
     def _set_languages(self):
```
ecino: You can rename the method to `_set_destination_language` as we won't compute the original_language.

sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -79,7 +79,7 @@ class SponsorshipCorrespondence(models.Model):
         readonly=True)
     # First spoken lang of partner
     original_language_id = fields.Many2one(
-        'res.lang.compassion', compute='_set_languages',
+        'res.lang.compassion', readonly=True,
         inverse='_change_language', store=True)
```
ecino: You can remove `_inverse` method as it is no more a computed field.

sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -172,38 +171,50 @@ def get_communication_types(self):
...
+            if corr.sponsorship_id and corr.communication_type_ids:
+                corr.name = corr.communication_type_ids[0].name + ' (' + \
+                            corr.sponsorship_id.partner_codega + " - " + \
+                            corr.child_id.code + ')'
+            else:
+                corr.name = _('New correspondence')
 
-    @api.depends('sponsorship_id', 'direction')
+    @api.depends('sponsorship_id', 'direction', 'original_language_id',
+                 'destination_language_id')
```
ecino: remove 'destination_language_id' from the depends

sbc_compassion/models/correspondence_template.py
ecino: Maybe rename `_compute_size` method then, because the method does not that.

sbc_compassion/models/sponsorship_correspondence.py
loikki: It was more a TOCO (to correct) than a TODO. If I remember correctly, you said that you will need to wait on the US to know exactly what you need

sbc_compassion/models/sponsorship_correspondence.py
loikki: Comment in your last comment ^^

sbc_compassion/models/sponsorship_correspondence.py
loikki: and the store?

sbc_compassion/models/import_letter_line.py
ecino: A comment explaining `_set_sponsorship_id` would be welcome.

sbc_compassion/models/import_letter_line.py
ecino: fine, except write fully `no_template` as `temp` can be misunderstood.

sbc_compassion/models/import_letter_line.py
```diff
@@ -40,8 +42,8 @@ class ImportLetterLine(models.Model):
 
     status = fields.Selection([
         ("lang", _("Error in Language")),
-        ("sponsor", _("Error in Sponsorship")),
-        ("encourager", _("Sponsorship not found")),
+        ("sponsorship", _("Sponsorship not Found")),
+        ("person", _("Error in Partner or Child")),
```
ecino: This error is not clear to me : error in partner or child. What does it mean exactly ?

sbc_compassion/models/correspondence_template.py
loikki: _compute_img_constant?

sbc_compassion/models/import_letters_history.py
ecino: Good point, but rename `letters` then :)

sbc_compassion/models/sponsorship_correspondence.py
ecino: yes you can remove it too!

sbc_compassion/models/import_letters_history.py
loikki: letters -> import_letters?

sbc_compassion/models/import_letter_line.py
loikki: I think I should rename it to "Partner or child not found". It is in order to have different errors depending on where the error is for the sponsorship. The two mains problems can be that the child or the partner is not found and the other one is that there is no sponsorship between them

sbc_compassion/models/correspondence_template.py
```diff
@@ -168,11 +172,72 @@ def _set_image(self):
...
+                    template.qrcode_x_min = template.page_width * float(
+                        template.env['ir.config_parameter'].get_param(
+                            'qrcode_x_min'))
+                    template.qrcode_x_max = template.page_width * float(
```
ecino: Maybe you can move this code inside `_compute_template_keypoints`method. and link it to the compute fields `qr_code`.
Then, inside this onchange method, you could simply call `self._compute_template_keypoints()`

It would be more clear no ? But you have to test it to see if it works.

ecino: To summarize what I mean : 
```
@api.onchange('template_image')
     def _compute_size(self):  # to rename to _onchange_template_image
          self._compute_template_keypoints()

File: sbc_compassion/models/import_letter_line.py (hunk 32 above)
Comment (ecino) [inline sbc_compassion/models/import_letter_line.py]: Ok this makes more sense.

File: sbc_compassion/models/import_letters_history.py (hunk 19 above)
Comment (ecino) [inline sbc_compassion/models/import_letters_history.py]: Ok

File: sbc_compassion/models/import_letters_history.py (hunk 22 above)
Comment (loikki) [inline sbc_compassion/models/import_letters_history.py]: I have created a new method for the whole part about the QR code

File: sbc_compassion/models/import_letters_history.py (hunk 23 above)
Comment (loikki) [inline sbc_compassion/models/import_letters_history.py]: I have created a method for the whole part about the QR code (with the rotation)

File: sbc_compassion/models/sponsorship_correspondence.py (hunk 24 above)
Comment (loikki) [inline sbc_compassion/models/sponsorship_correspondence.py]: Did not think about that ^^'
I removed the readonly

File: sbc_compassion/models/correspondence_template.py (hunk 3 above)
Comment (loikki) [inline sbc_compassion/models/correspondence_template.py]: done

File: sbc_compassion/models/correspondence_template.py (hunk 5 above)
Comment (loikki) [inline sbc_compassion/models/correspondence_template.py]: Now I think I know why, it is due to some function calls in odoo when the field is under compute (I do not have enough control over it when it is a compute)

File: sbc_compassion/models/correspondence_template.py [hunk 34]
```diff
@@ -59,31 +62,35 @@ class CorrespondenceTemplate(models.Model):
...
         help='Maximum Y position of the area in which to look for the QR '
              'code inside the template (given in pixels)')
     pattern_center_x = fields.Float(
-        readonly=True,
+        compute="_compute_template_keypoints",
         help='X coordinate of the center of the pattern. '
         'Used to detect the orientation of the pattern in the image')
     pattern_center_y = fields.Float(
-        readonly=True,
+        compute="_compute_template_keypoints",
```
ecino: Why is bluesquare stored but not pattern_center ? If pattern is not stored, `_compute_template_keypoints` will be called each time a template is opened.

sbc_compassion/models/correspondence_template.py
```diff
@@ -168,7 +175,76 @@ def _set_image(self):
...
+        for template in self:
+            if template.template_image:
+                with tempfile.NamedTemporaryFile() as template_file:
+                    data = template.with_context(bin_size=False).template_image
+                    if not data:
+                        data = template.template_image
+                    template_file.write(base64.b64decode(
+                        data))
+                    template_file.flush()
+                    # Find the pattern inside the template image
```
ecino: Here you say you find the pattern, but method `_compute_img_constant` seems to compute only qr code and page size.

sbc_compassion/models/correspondence_template.py
```diff
@@ -168,7 +175,76 @@ def _set_image(self):
...
+            float(self.env['ir.config_parameter'].get_param('qrcode_y_max'))
+        return img
+
+    @api.depends('template_attachment_id',
+                 'pattern_image', 'template_image')
+    def _compute_template_keypoints(self):
+        """ This method computes all keypoints that can be automatically
+        detected (bluesquare and pattern_center)
+        """
+        for template in self:
```
ecino: You can here add the context :
```
for template in self.with_context(bin_size=False):
```
...

sbc_compassion/models/correspondence_template.py
```diff
@@ -168,7 +175,76 @@ def _set_image(self):
...
+            if (template.template_image and template.pattern_image and
+                    _verify_template(template)):
+                with tempfile.NamedTemporaryFile() as template_file:
+                    template_file.write(base64.b64decode(
+                        template.with_context(bin_size=False).template_image))
```
ecino: ... and remove context from here

sbc_compassion/models/correspondence_template.py
```diff
@@ -168,7 +175,76 @@ def _set_image(self):
...
+        return img
+
+    @api.depends('template_attachment_id',
+                 'pattern_image', 'template_image')
+    def _compute_template_keypoints(self):
+        """ This method computes all keypoints that can be automatically
+        detected (bluesquare and pattern_center)
+        """
+        for template in self:
+            if (template.template_image and template.pattern_image and
```
ecino: Outter parentheses not needed.

sbc_compassion/models/correspondence_template.py
```diff
@@ -168,7 +175,76 @@ def _set_image(self):
...
+                    _verify_template(template)):
+                with tempfile.NamedTemporaryFile() as template_file:
+                    template_file.write(base64.b64decode(
+                        template.with_context(bin_size=False).template_image))
+                    template_file.flush()
+                    # Find the pattern inside the template image
```
ecino: Again `_compute_img_constant` does not find the pattern.

sbc_compassion/models/import_letter_line.py
```diff
@@ -25,8 +25,10 @@ class ImportLetterLine(models.Model):
 
     sponsorship_id = fields.Many2one('recurring.contract', 'Sponsorship',
                                      compute='_set_sponsorship_id')
-    sponsorship_status = fields.Boolean(compute='_set_sponsorship_id',
-                                        readonly=True)
+    sponsorship_found = fields.Boolean(compute='_set_sponsorship_id',
+                                       readonly=True)
```
ecino: readonly not needed for comupted fields (if no inverse function is set)

sbc_compassion/models/import_letter_line.py
```diff
@@ -25,8 +25,10 @@ class ImportLetterLine(models.Model):
...
+    child_partner_found = fields.Boolean(compute='_set_sponsorship_id',
+                                         readonly=True)
```
ecino: readonly not needed

sbc_compassion/models/import_letter_line.py
```diff
@@ -39,10 +41,10 @@ class ImportLetterLine(models.Model):
     letter_image_preview = fields.Binary()
 
     status = fields.Selection([
-        ("lang", _("Error in Language")),
-        ("sponsor", _("Error in Sponsorship")),
-        ("encourager", _("Sponsorship not found")),
-        ("temp", _("Error in Template")),
+        ("no_lang", _("Error in Language")),
```
ecino: "Language not detected"

sbc_compassion/models/import_letter_line.py
```diff
@@ -39,10 +41,10 @@ class ImportLetterLine(models.Model):
...
+        ("no_sponsorship", _("Sponsorship not Found")),
+        ("no_child_partner", _("Partner or Child not Found")),
+        ("no_template", _("Error in Template")),
```
ecino: "Template not detected"

sbc_compassion/models/import_letter_line.py
```diff
@@ -53,36 +55,55 @@ class ImportLetterLine(models.Model):
...
                     ('partner_codega', '=', line.partner_codega),
                     ('is_active', '=', True)], order='end_date desc', limit=1)
-                if len(line.sponsorship_id) == 1:
-                    line.sponsorship_status = None
+                if not line.sponsorship_id:
+                    line.sponsorship_id = line.env[
+                        'recurring.contract'].search(
+                            [('child_id.code', '=', line.child_code),
+                             ('partner_codega', '=', line.partner_codega)],
+                            order='start_date desc', limit=1)
```
ecino: I think you don't have to do two consecutive searches. You can do only one and put in the order clause :
```
order='is_active desc, end_date desc'
```

sbc_compassion/models/import_letter_line.py
```diff
@@ -92,3 +113,28 @@ def _set_name(self):
                 line.name = str(
                     line.sponsorship_id.partner_codega) + " - " + str(
                         line.child_code)
+
+    @api.multi
+    def get_letter_datas(self, mandatory_review=False):
+        """ Create a list of dictionaries in order to create them inside
```
ecino: To create __what__ inside __what__ ? Please clarify the comment.

sbc_compassion/models/correspondence_template.py
loikki: It was an old comment for all the part, not just the one between this comment and the next one. I have updated it

sbc_compassion/models/correspondence_template.py
loikki: Not needed but recommend: http://stackoverflow.com/a/181557

sbc_compassion/models/import_letter_line.py
loikki: I tried without, but if someone is dropping a sponsorship and want to send a last letter, it won't be possible without the second one

ecino: Yes it is possible by removing condition `is_active = True` and put the order I suggested.

sbc_compassion/models/correspondence_template.py
ecino: ok

ecino:
Thanks ! :+1:

PR #140, opened 2015-12-01
Fix views of import letters process

ecino:
- Replace import at save by a button to trigger the import
- Change states colors
- Change states names
- Fix small code issues


sbc_compassion/models/import_letters_history.py
```diff
@@ -251,7 +277,7 @@ def _analyze_attachment(self, file_, filename):
         template, key_img = self._find_template(img)
         lang_id = self._find_language(img, key_img, template)
 
-        letters_line = self.env['import.letter.line'].sudo().create({
+        letters_line = self.env['import.letter.line'].create({
```
loikki: Did you try it without admin right?

sbc_compassion/security/ir.model.access.csv
```diff
@@ -8,4 +8,4 @@
 "access_correspondence_type","Read access on correspondence_type","model_sponsorship_correspondence_type","child_compassion.group_sponsorship",1,0,0,0
 "access_correspondence_lang_checkbox","Read access on correspondence_checkbox","model_sponsorship_correspondence_lang_checkbox","child_compassion.group_sponsorship",1,0,0,0
 "access_import_letters_history","Access on import_letters_history","model_import_letters_history","child_compassion.group_sponsorship",1,1,1,0
-"access_import_letters_line","Access on import_letters_line","model_import_letter_line","child_compassion.group_sponsorship",1,1,0,1
+"access_import_letters_line","Access on import_letters_line","model_import_letter_line","child_compassion.group_sponsorship",1,1,1,1
```
loikki: If I remember correctly, we said that we do not want people creating a letter_line without using the import

sbc_compassion/models/import_letters_history.py
ecino: Yes, I added create right for non-admin users.

ecino:
I corrected the view to prevent user from creating a import_line, even though the access rights allows to do it. We thus don't need a sudo() to import the files but still cannot create from the view.

PR #146, opened 2015-12-01
Add demo data for partner, child, country and contract

loikki:


child_compassion/__openerp__.py
```diff
@@ -56,7 +56,11 @@
         'data/delegate_child_compassion_cron.xml',
         'data/child_compassion_template.xml',
     ],
-    'demo': [],
+    'demo': [
+        'demo/compassion.country.csv',
```
ecino: It is preferrable to put underscores in file names instead of dots.

child_compassion/demo/compassion.child.csv
```diff
@@ -0,0 +1,67 @@
+id,code,name
```
ecino: name is not a mandatory field and is automatically populated by the webservice when record is created, so I think you can remove this column.

child_compassion/demo/install.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+	<data noupdate='1'>
+        <function model="compassion.country" name="_set_demo_data"/>
```
ecino: Fix indentation

child_compassion/model/country_compassion.py
```diff
@@ -72,3 +72,10 @@ def _get_url(self, api_mess, api_value):
         url += ('/ci/v1/countries/' + api_mess + '/' + api_value +
                 '?api_key=' + api_key)
         return url
+
+    @api.model
+    def _set_demo_data(self):
+        """
+        Update the info of the demo countries
+        """
+        self.env[self._name].search([]).update_informations()
```
ecino: self~~.env[self._name]~~.search is enough

child_compassion/demo/compassion.child.csv
loikki: At first, it is what I have done, but it seems that they are not updated (or badly)

ecino: fine

child_compassion/demo/compassion.child.csv
```diff
@@ -0,0 +1,67 @@
...
+child_01,"BR2150576","Marilyn Monroe"
+child_02,"RW1440285","Abraham Lincoln"
```
ecino: You could speed up install by using children of same projects (does not need to fetch data if already exists).
For those who are in the same country (i.e. RW144XXXX, RW530XXXX) you can change and put only RW144XXXX (one project per country).

child_compassion/__openerp__.py
ecino: This file is not needed. When creating children, the project will be automatically created with its country.

child_compassion/__openerp__.py
```diff
@@ -56,7 +56,11 @@
         'data/delegate_child_compassion_cron.xml',
         'data/child_compassion_template.xml',
     ],
-    'demo': [],
+    'demo': [
+        'demo/compassion.country.csv',
+        'demo/install.xml',
```
ecino: This file is not needed. When creating children, the project will be automatically created with its country.

child_compassion/__openerp__.py
loikki: The problem is for the id in the xml.
I need to update them in sbc_compassion (for the languages)

child_compassion/__openerp__.py
loikki: It is not for the project but for the country (see above)

child_compassion/demo/compassion.child.csv
loikki: Maybe I keep just one case where a country has two projects (it will be one more test)

ecino: Ok fine

child_compassion/__openerp__.py
ecino: Right, I didn't remember this.

child_compassion/__openerp__.py
ecino: fine

ecino:
:+1:

PR #147, opened 2015-12-02
Fix #138

### LINKED ISSUE #138: Import lines fields partner_codega and child_code should be references
Change fields `partner_codega` and `child_code` to `partner_id` and `child_id` with link to actual records of related models. This makes easier setting the child and the partner in case the qr code detection didn't work.

### Should adapt method `_find_qrcode` :

* https://github.com/CompassionCH/compassion-modules/blob/sbc-devel/sbc_compassion/models/import_letters_history.py#L246

Instead of returning multiple variables, the function could return the dictionary of values needed to be passed to the `create` called at https://github.com/CompassionCH/compassion-modules/blob/sbc-devel/sbc_compassion/models/import_letters_history.py#L254

* https://github.com/CompassionCH/compassion-modules/blob/sbc-devel/sbc_compassion/models/import_letters_history.py#L285

Variable `data` is never used inside the method, but is overidden and returned. Check what is the purpose of this and correct the code by either removing data from the parameters or making use of it. Also add a comment to the function to document the parameters.

loikki:
This fix is done with the data from my other pull request, maybe I will need to do an other one

### DIFF
--- a/sbc_compassion/models/import_letter_line.py
+++ b/sbc_compassion/models/import_letter_line.py
@@ -27,9 +27,9 @@ class ImportLetterLine(models.Model):
                                      compute='_set_sponsorship_id')
     sponsorship_found = fields.Boolean(compute='_set_sponsorship_id')
     child_partner_found = fields.Boolean(compute='_set_sponsorship_id')
-    partner_codega = fields.Char('Partner')
+    partner_id = fields.Many2one('res.partner', 'Partner')
     name = fields.Char(compute='_set_name')
-    child_code = fields.Char('Child')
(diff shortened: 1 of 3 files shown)


ecino:
Your branch does not match anymore sbc-devel, I guess you created the branch before updating sbc-devel. Please rebase (commit Add demo data should not appear).

loikki:
I have done it but it seems that we have lost all your comments :/

ecino:
By looking again at the code, I see you corrected the comments I made (at least those I remember) :+1:

PR #153, opened 2015-12-09
Move computation outside the model import.letters.history

loikki:
I am preparing everything for my wizard. I think it will be better for future pull request if theses modifications are quickly integrated


sbc_compassion/models/import_letters_history.py
```diff
@@ -62,7 +54,7 @@ class ImportLettersHistory(models.Model):
         'import.letter.line', 'import_id', 'Files to process')
     letters_ids = fields.One2many(
         'sponsorship.correspondence', 'import_id', 'Imported letters',
-        readonly=True)
+        readonly=True, ondelete='cascade')
```
ecino: Not sure for this one. I would have put it for the field import_line_ids but not letters_ids.

loikki: sorry I was a little bit too fast ^^
I wanted to put it to the other one. I think that this one should be kept normal (not cascade) as it is possible to delete everything from the other menu

ecino:
:+1:

PR #155, opened 2015-12-11
Add wizard for reviewing letters imports

### LINKED ISSUE #141: Import letters: Wizard for Import Line Review
We should create a wizard showing the image in order to ask a human being to verify the language and the template (at least for the first few months of use in order to verify the algorithm at a larger scale).

* Browse the import lines that need verification one by one
* Display image and import line data that can be edited

Rules for including a line in the review process :

* Letter does not need translation (to be sure it is really the case and avoid transmitting a letter the child won't understand)
* Default template is selected but `Force default template` was not activated, to be sure it wasn't a valid template not detected
* Any missing data
* ... something else ?

ecino:
* [ADD] wizard for reviewing all imports easily
* [CHANGE] imports are done asynchronously using connectors
* [FIX] small bugs

Fixes #141

### DIFF
--- a/sbc_compassion/tools/import_letter_functions.py
+++ b/sbc_compassion/tools/import_letter_functions.py
@@ -150,10 +150,13 @@ def analyze_attachment(env, file_, filename, force_template, test=False):
     :param str filename: Filename to give in odoo
     :param sponsorship.correspondence.template force_template: Template
     :param bool test: Test import or not
+
+    :returns: Import Line values, IR Attachment values
+    :rtype: dict, dict
     """
-    f = open(file_)
-    data = f.read()
-    f.close()
+    line_vals = {}
+    document_vals = {}
+
     # convert to PNG
     if isPDF(file_) or isTIFF(file_):
         name = os.path.splitext(file_)[0]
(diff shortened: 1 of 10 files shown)


loikki:
Except from the failed build (and the few bugs that you want to remove): :+1:

loikki:
I am trying to use it.
1) I am not able to import the letters, the code is not executing correctly.
2) I think that the state of an import does not take into account the possibility of a failed import

ecino:
Pull request updated. A failed import stays in "Analyzing" state and should be resolved from the "Connector" menu (Requeue job).

loikki:
until the commit 06db4d6 ("Use connectors for importing letters"): :+1:

ecino:
You can now review the whole pull request. I didn't removed count in review wizard, it is needed in the view to display (x/20).

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -485,9 +481,10 @@ def _find_languages(env, img, pattern_center, line_vals, test):
 
     # if pattern has not been detected
     if pattern_center is None:
-        return line_vals
+        return
 
-    template = line_vals['template_id']
+    template = env['sponsorship.correspondence.template'].browse(
```
loikki: I do not see the point if you are not updating the way you use it later

ecino: `line_vals['template_id']` is an integer value.
At line https://github.com/CompassionCH/compassion-modules/pull/155/files#diff-e41465c073f2badef6f5427249b3c86bR499
we use `template` object and need therefore to fetch the record.

loikki: Ok therefore you can merge it

PR #156, opened 2015-12-11
Add a new menu for the admin in order to check the quality of the import

loikki:
In Sponsorship->Configuration->Test Imports admin can do an import
(with an additionnal csv) that will compare the values obtained and
the value in the csv.

### DIFF
--- a/sbc_compassion/tests/test_tools.py
+++ b/sbc_compassion/tests/test_tools.py
@@ -60,8 +60,8 @@ class TestTools(common.TransactionCase):
         self.assertIsNotNone(template)
         self.assertEqual(template.name, 'Test Template 2')
         np.testing.assert_allclose(pattern_center,
-                                   np.array([1273.47, 3298.65]),
-                                   atol=0.1)
+                                   np.array([1273, 3301]),
+                                   atol=1)
 
     def test_pattern_recognition_no_pattern(self):
         """
@@ -121,7 +121,10 @@ class TestTools(common.TransactionCase):
 
     def _pattern_recognition(self, path):
         img = self._read_img(path)
-        return tools.patternrecognition.find_template(img, self.templates, 0.3)
+        # when the parameter test=False (default value)
+        # find_template returns a third parameter equals to None
+        return tools.patternrecognition.find_template(
+            img, self.templates, 0.3)[:2]
 
     @staticmethod
     def _qr_decode(path):
--- a/sbc_compassion/tools/import_letter_functions.py
+++ b/sbc_compassion/tools/import_letter_functions.py
@@ -21,13 +21,121 @@ import patternrecognition as pr
 import checkboxreader as cbr
 import bluecornerfinder as bcf
 import numpy as np
+from math import ceil
 from glob import glob
 from wand.image import Image
+from collections import namedtuple
 
+from openerp import _, exceptions
 
 ##########################################################################
 #                           GENERAL METHODS                              #
 ##########################################################################
+
+
+def testline(env, line_vals, csv_file_ids, doc_name):
+    """ Test each line and verify the data given by the csv.
+    A csv file need to be written in the same way than
+    sbc_compassion/tests/testdata/import/travis_file.csv
+
+    :param env: Odoo environment
+    :param dict line_vals: Dictionnary that will be used to create line
+    :param Recordset('ir.attachment') csv_file_ids: CSV used to check
+    :param str doc_name: Document name
+    :returns: Error detected
+    :rtype: str
+    """
+    # str that will be returned
+    error = ''
+    csv_data = []
+    # creates a list containing all the csv data
+    for csv_ in csv_file_ids:
+        if os.path.splitext(csv_.name)[1] != '.csv':
+            raise exceptions.Warning(
+                _("File '{}' is not a csv file".format(csv_.name)))
+        csv_string = base64.b64decode(csv_.datas).splitlines()
+        csv_data += read_csv(csv_string)
+    # check that the letter is present in the csv file and find it
+    values = {}
+    for csv_line in csv_data:
+        if csv_line['name'] == doc_name:
+            values = csv_line
+            break
+    else:
+        raise exceptions.Warning(
+            _("File '{}' not found in the csv files".format(doc_name)))
+    # Test that the template is the same as given in the csv
+    template_csv = env.ref(
+        'sbc_compassion.'+values['template_id'])
+    if template_csv.id != line_vals['template_id']:
+        error = error + 'template,\n'
+
+    # Test that the barcode detection matches the result given in csv file
+    BarcodeTuple = namedtuple('BarcodeTuple', ['data'])
+    barcode = BarcodeTuple(values['barcode'])
+    partner_detected, child_detected = decodeBarcode(env, barcode)
+    # 1) we check the case where the barcode was not detected and was not
+    # expected
+    if (line_vals['child_id'] != child_detected or
+            line_vals['partner_id'] != partner_detected):
+        error = error + 'barcode,\n'
+
+    # Test that languages detected are the same as given in the csv file
+    lang_ok = True
+    letter_languages_id = line_vals['test_letter_languages_id'].split(',')
+    for lang in letter_languages_id:
+        lang_ok = lang_ok and lang in values['lang']
+    line_nber_lang = len(letter_languages_id)
+    csv_nber_lang = len(values['lang'].split(','))
+    lang_ok = lang_ok and line_nber_lang == csv_nber_lang
+
+    if not lang_ok:
+        error = error + 'lang,\n'
+
+    # remove the two last character (',\n')
+    return error[:-2]
+
+
+def update_stat_text(test_import):
+    """ Update the text for the stats in test_import ("12/34 (35.3%)").
+    Test is used in order to know if the number of passed test needs to be
+    increased.
+    :param test.import.letters.history test_import: Item to check
+    :returns: Updated Item
+    :rtype: test.import.letters.history
+    """
+    test = 0
+    qr = 0
+    tpl = 0
+    lang = 0
+    for line in test_import.test_import_line_ids:
+        if len(line.error) > 0:
+            test += 1
+        if 'template' in line.error:
+            tpl += 1
+        if 'barcode' in line.error:
+            qr += 1
+        if 'lang' in line.error:
+            lang += 1
+
+    test = test_import.nber_test - test
+    tpl = test_import.nber_test - tpl
+    qr = test_import.nber_test - qr
+    lang = test_import.nber_test - lang
+
+    test_import.test_ok = write_text_test(test, test_import.nber_test)
+    test_import.qr_ok = write_text_test(qr, test_import.nber_test)
+    test_import.lang_ok = write_text_test(lang, test_import.nber_test)
+    test_import.template_ok = write_text_test(tpl, test_import.nber_test)
+    return test_import
+
+
+def write_text_test(nber, tot):
+    ratio = float(nber) / float(tot)
+    text = str(nber) + '/' + str(tot) + ' (' + str.format(
+        "{0:.2f}", 100*ratio) + '%)'
+    return text
+
 
 def analyze_attachment(env, file_, filename, force_template, test=False):
     """
@@ -61,18 +169,24 @@ def analyze_attachment(env, file_, filename, force_template, test=False):
         img = cv2.imread(file_png)
         MULTIPAGE = True
         if img is None:
-            raise Warning("The '{}' image cannot be read".format(filename))
+            raise exceptions.Warning(
+                "The '{}' image cannot be read".format(filename))
     # first compute the QR code
     line_vals, data, img = _find_qrcode(env, img, MULTIPAGE, file_, data, test)
     # now try to find the layout
     # loop over all the patterns in the pattern directory
-    line_vals, pattern_center = _find_template(env, img, line_vals)
+    line_vals, pattern_center = _find_template(env, img, line_vals, test)
     if line_vals['template_id'] != env.ref('sbc_compassion.default_template'):
-        line_vals = _find_languages(env, img, pattern_center, line_vals)
+        line_vals = _find_languages(env, img, pattern_center, line_vals, test)
     else:
         line_vals.update({
             'letter_language_id': False
         })
+        if test:
+            line_vals.update({
+                'lang_preview': '',
+                'test_letter_languages_id': ''
+            })
     if force_template and force_template is not None:
         line_vals.update({
             'template_id': force_template.id
@@ -80,29 +194,25 @@ def analyze_attachment(env, file_, filename, force_template, test=False):
 
     line_vals.update({
         'is_encourager': False,
+        'template_id': line_vals['template_id'].id
     })
-    letters_line = env['import.letter.line'].create(line_vals)
 
     file_png_io = open(file_png, "r")
     file_data = file_png_io.read()
     file_png_io.close()
 
-    document_vals = {'name': filename,
-                     'datas': base64.b64encode(data),
-                     'datas_fname': filename,
-                     'res_model': 'import.letter.line',
-                     'res_id': letters_line.id}
-
-    letters_line.letter_image = env[
-        'ir.attachment'].create(document_vals)
-    letters_line.letter_image_preview = base64.b64encode(file_data)
+    document_vals = {
+        'name': filename,
+        'datas': base64.b64encode(data),
+        'datas_fname': filename,
+    }
 
     os.remove(file_png)
     if MULTIPAGE:
         delfiles = glob(name + '*png')
         for file_ in delfiles:
             os.remove(file_)
-    return letters_line
+    return line_vals, document_vals, file_data
 
 
 def _find_qrcode(env, img, MULTIPAGE, file_,
@@ -149,6 +259,11 @@ def _find_qrcode(env, img, MULTIPAGE, file_,
     # first attempt
     qrcode = zx.decode(file_png, try_harder=True, crop=[
         int(left), int(top), int(width), int(height)])
+    if test:
+        right = int(left)+int(width)
+        bottom = int(top) + int(height)
+        test_img = cv2.imread(file_png)[int(top):bottom, int(left):right]
+        test_data = readEncode(test_img)
     # check if found, if not means that page is rotated
     if qrcode is None:
         # rotate image
@@ -158,44 +273,87 @@ def _find_qrcode(env, img, MULTIPAGE, file_,
         qrcode = zx.decode(file_png, try_harder=True, crop=[
             int(left), int(top), int(width), int(height)])
         # if the QR code is found
-        if qrcode is not None:
+        if qrcode is not None and testDirectionQRcode(qrcode):
+            right = int(left)+int(width)
+            bottom = int(top) + int(height)
+            test_img = cv2.imread(file_png)[int(top):bottom, int(left):right]
+            test_data = readEncode(test_img)
             # save it for zxing
             cv2.imwrite(file_png, img)
             # replace the image by the returned one
             data = _save_img(MULTIPAGE, file_, img)
         else:
-            cv2.imwrite(file_png, img[::-1, ::-1])
+            img = img[::-1, ::-1]
+            cv2.imwrite(file_png, img)
             qrcode = zx.decode(file_png, try_harder=True)
-    if qrcode is not None and 'XX' in qrcode.data:
-        partner_id, child_id = qrcode.data.split('XX')
-        child_id = env['compassion.child'].search(
-            [('code', '=', child_id)]).id
-        partner_id = env['res.partner'].search(
-            [('ref', '=', partner_id),
-             ('is_company', '=', False)]).id
 
-    else:
-        partner_id = None
-        child_id = None
+    partner_id, child_id = decodeBarcode(env, qrcode)
 
     dict_odoo = {
         'partner_id': partner_id,
         'child_id': child_id,
     }
     if test:
-        right = int(left)+int(width)
-        bottom = int(top) + int(height)
-        test_img = cv2.imread(file_png)[int(top):bottom, int(left):right]
-        with tempfile.NamedTemporaryFile(
-                suffix='.png') as temp_file:
-            cv2.imwrite(temp_file.name, test_img)
+        dict_odoo.update({
+            'qr_preview': test_data
+        })
+    return dict_odoo, data, img
+
+
+def testDirectionQRcode(barcode):
+    """ Test if the direction of the QR code is correct
+    :param Barcode barcode: Data from zxing
+    :return: True if the qrcode is in the right direction
+    :rtype: bool
+    """
+    # vector between the two uppermost points (index 1 is
+    # the one at the upper leftmost corner)
+    top = np.array(barcode.points[2]) - np.array(barcode.points[1])
+    left = np.array(barcode.points[0]) - np.array(barcode.points[1])
+    diag = top + left
+    if diag[0] > 0 < diag[1]:
+        return True
+    else:
+        return False
+
+
+def decodeBarcode(env, barcode):
+    """ Split the barcode and return the id of the partner and the child.
+    If the partner is not found, return None (same for the child).
+    :param env: Odoo environment
+    :param Barcode barcode:
+    :return: partner.id, Child.id
+    :rtype: int, int
+    """
+    partner_id = None
+    child_id = None
+    if barcode is not None and 'XX' in barcode.data:
+        barcode_split = barcode.data.split('XX')
+        if len(barcode_split) == 2:
+            partner_ref, child_code = barcode_split
+            child_id = env['compassion.child'].search(
+                [('code', '=', child_code)]).id
+            partner_id = env['res.partner'].search(
+                [('ref', '=', partner_ref),
+                 ('is_company', '=', False)]).id
+    return partner_id, child_id
+
+
+def readEncode(img, format_img='png'):
+    """
+    Create a tempfile, write the image on it and return a base64 encoded
+    string
+    :param np.array img: Image
+    :param str format_img: Format image
+    :returns: Data (encoded in base64)
+    :rtype: str
+    """
+    with tempfile.NamedTemporaryFile(suffix='.'+format_img) as temp_file:
+            cv2.imwrite(temp_file.name, img)
             f = open(temp_file.name)
             test_data = f.read()
             f.close()
-        dict_odoo.update({
-            'qr_preview': base64.b64encode(test_data)
-        })
-    return dict_odoo, data, img
+    return base64.b64encode(test_data)
 
 
 def _save_img(MULTIPAGE, file_, img):
@@ -242,12 +400,13 @@ def _save_img(MULTIPAGE, file_, img):
     return data
 
 
-def _find_template(env, img, line_vals):
+def _find_template(env, img, line_vals, test):
     """
     Use pattern recognition to detect which template corresponds to img.
     :param env env: Odoo variable env
     :param img: Image to analyze
     :param dict line_vals: Dictonnary containing the data for a line
+    :param bool test: Enable the test mode (will save some img)
     :returns: line_vals (updated), center position of detected pattern
     :rtype: dict, layout
     """
@@ -255,18 +414,25 @@ def _find_template(env, img, line_vals):
         [('pattern_image', '!=', False)])
     threshold = float(
         env.ref('sbc_compassion.threshold_keypoints_template').value)
-    template, pattern_center = pr.find_template(img, templates,
-                                                threshold=threshold)
+    template, pattern_center, img = pr.find_template(
+        img, templates, threshold=threshold, test=test)
+    if test:
+        img = manyImages2OneImage(img, 1)
+        line_vals.update({
+            'template_preview': img
+        })
+
     if template is None:
         pattern_center = np.array(([0, 0], [0, 0]))
         template = env.ref('sbc_compassion.default_template')
+
     line_vals.update({
         'template_id': template
     })
     return line_vals, pattern_center
 
 
-def _find_languages(env, img, pattern_center, line_vals):
+def _find_languages(env, img, pattern_center, line_vals, test):
     """
     Use the pattern and the blue corner for doing a transformation
     (rotation + scaling + translation) in order to crop a small part
@@ -302,13 +468,26 @@ def _find_languages(env, img, pattern_center, line_vals):
     :param pattern_center: Center position of detected pattern
     :param dict line_vals: Dictonnary containing the data for a line
     (and the template)
+    :param bool test: Enable the test mode (will save some img)
     :returns: line_vals (updated)
     :rtype: dict
     """
-    template = line_vals['template_id']
+    line_vals['letter_language_id'] = False
+    if test:
+        # The color order is BGR
+        lang_color = (0, 97, 232)
+        corner_color = (0, 255, 0)
+
+        line_vals.update({
+            'lang_preview': '',
+            'test_letter_languages_id': ''
+        })
+
     # if pattern has not been detected
     if pattern_center is None:
-        return []
+        return line_vals
+
+    template = line_vals['template_id']
     # get position of the blue corner
     box = [float(
         env['ir.config_parameter'].get_param('bluecorner_x_min')),
@@ -317,7 +496,7 @@ def _find_languages(env, img, pattern_center, line_vals):
     bluecorner = bcf.BlueCornerFinder(img, box=box)
     bluecorner_position = bluecorner.getIndices()
     if bluecorner_position is None:
-        return []
+        return line_vals
 
     # vector between the blue square and the pattern
     diff_ref = np.array(template.get_bluesquare_area() -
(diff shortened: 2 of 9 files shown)


loikki:
No need to review yet, I still want to add something to the preview

sbc_compassion/wizards/import_letters_test.py
```diff
@@ -0,0 +1,107 @@
...
+import base64
+
+from openerp import models, fields, _, api
+from ..tools import import_letter_functions as func
+
+
+class TestImportLetters(models.TransientModel):
+    """ Class testing the letters import by highlighting the keypoints of
+    the import (image used, keypoints found for the templates, showing only
+    the line not well detected, ...)
```
ecino: Avoid ... in comments

sbc_compassion/wizards/import_letters_test.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+    _inherit = 'import.letters.history'
+    _name = 'test.import.letters.history'
+    _description = _("""Easier way to test the import""")
+    csv_file = fields.Many2many(
```
ecino: Can it have many csv attached ? Please respect the standard convention for field names :
Many2one should end by `_id` and Many2many should end with `_ids`

sbc_compassion/wizards/import_letters_test.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+class TestImportLetters(models.TransientModel):
+    """ Class testing the letters import by highlighting the keypoints of
+    the import (image used, keypoints found for the templates, showing only
+    the line not well detected, ...)
+    """
+    _inherit = 'import.letters.history'
+    _name = 'test.import.letters.history'
+    _description = _("""Easier way to test the import""")
```
ecino: This is not a good description. Either remove it or be more explanatory.

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,117 @@
...
+        csv_string = base64.b64decode(csv_.datas).splitlines()
+        csv_data += read_csv(csv_string)
+    # check the letters
+    values = {}
+    test = True
+    for csv_line in csv_data:
+        if csv_line['name'] == doc_name:
+            values = csv_line
+            break
+    if values == {}:
```
ecino: `if not values` is better for testing emptiness of dictionary.

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,117 @@
...
+        raise Warning(
+            _("File '{}' not found in the csv files".format(doc_name)))
+    # first test: template
+    template_id = env.ref(
+        'sbc_compassion.'+values['template_id'])
+    tpl_detected = env['sponsorship.correspondence.template'].search(
+        [('id', '=', line_vals['template_id'])])
+    temp_test = template_id == tpl_detected
```
ecino: I think there is no need to have `temp_test` and `test` variables. You can simply do :
```
if not template_id == tpl_detected:
    error = error + 'template,\n'
...
```

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,117 @@
...
+    # partner
+    partner_temp = env['res.partner'].search([
+        ('id', '=', line_vals['partner_id'])])
+    # child
+    child_temp = env['compassion.child'].search([
+        ('id', '=', line_vals['child_id'])])
+    pcodega = partner_temp.ref
+    ccode = child_temp.code
+    temp_test = True
+    if values['barcode'] == '' and not (pcodega and ccode):
```
ecino: Can you explain the following test and what do you capture in `temp_test` ? I have the feeling this can be greatly simplified (until line 96).

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,117 @@
...
+        temp_test = temp_test and lang in values['lang']
+    line_nber_lang = len(line_vals['test_letter_languages_id'].split(','))
+    csv_nber_lang = len(values['lang'].split(','))
+    temp_test = temp_test and line_nber_lang == csv_nber_lang
+
+    test = temp_test and test
+    if not temp_test:
+        error = error + 'lang,\n'
+
+    return test, error[:-2]
```
ecino: Add a comment explaining you remove end of line in error string. You could avoid to return `test` variable and instead check outside of the func:

```
if error:
    .... (means an error was detected and error string is not empty)
else:
    .... (no error detected)
```

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,117 @@
...
+    # child
+    child_temp = env['compassion.child'].search([
+        ('id', '=', line_vals['child_id'])])
+    pcodega = partner_temp.ref
+    ccode = child_temp.code
+    temp_test = True
+    if values['barcode'] == '' and not (pcodega and ccode):
+        temp_test = True
+    # xor (if both pcodega and ccode are False => barcode not detected)
+    elif bool(pcodega) ^ bool(ccode):
```
ecino: Don't understand what is your test

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,117 @@
...
+    temp_test = temp_test and line_nber_lang == csv_nber_lang
+
+    test = temp_test and test
+    if not temp_test:
+        error = error + 'lang,\n'
+
+    return test, error[:-2]
+
+
+def update_text(text, test):
```
ecino: Not a good method name, it is too generic.

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,117 @@
...
 from wand.image import Image
 
+from openerp import _
+
 
 ##########################################################################
 #                           GENERAL METHODS                              #
 ##########################################################################
+def testline(env, line_vals, csv_file, doc_name):
+    """ Test each line and verify the data given by the csv.
```
ecino: Do you document anywhere what the csv file looks like ? Is it the same as the one used for the tests ? There should be at least somwhere inside the module a explanation about it (if there already is one, you can point to it inside this docstring).

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,117 @@
...
+    If a line does not correspond to the csv, this function erases it.
+    :param env: Odoo environment
+    :param dict line_vals: Dictionnary that will be used to create line
+    :param One2many('test.import.letter.line') csv_file: CSV used to check
```
ecino: This doesn't make sense to me : a param cannot be a One2many. One2many is a field, It can be a Recordset of 'test.import.letter.line' objects. But is it a recordset or a CSV file ?

sbc_compassion/wizards/import_letters_test.py
loikki: And One2many with _ids?

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,135 @@
...
+    temp_test = temp_test and line_nber_lang == csv_nber_lang
+
+    test = temp_test and test
+    if not temp_test:
+        error = error + 'lang,\n'
+
+    return test, error[:-2]
+
+
+def update_text(test_import):
```
ecino: Not a good method name, it is too generic.

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -66,13 +188,18 @@ def analyze_attachment(env, file_, filename, force_template, test=False):
...
     # loop over all the patterns in the pattern directory
-    line_vals, pattern_center = _find_template(env, img, line_vals)
+    line_vals, pattern_center = _find_template(env, img, line_vals, test)
     if line_vals['template_id'] != env.ref('sbc_compassion.default_template'):
-        line_vals = _find_languages(env, img, pattern_center, line_vals)
+        line_vals = _find_languages(env, img, pattern_center, line_vals, test)
     else:
         line_vals.update({
-            'letter_language_id': False
+            'letter_language_id': None
```
ecino: Usually, False reprensents an empty relation (not None).

sbc_compassion/tools/import_letter_functions.py
loikki: I need to return test, therefore I am avoiding to test twice the equality by writing it inside a variable (and it decreases the size of a line)

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -242,12 +383,13 @@ def _save_img(MULTIPAGE, file_, img):
...
 
 
-def _find_template(env, img, line_vals):
+def _find_template(env, img, line_vals, test):
     """
     Use pattern recognition to detect which template corresponds to img.
     :param env env: Odoo variable env
     :param img: Image to analyze
     :param dict line_vals: Dictonnary containing the data for a line
+    :param bool test: Save the image of all the template detections
```
ecino: This is really a bad choice for variable name if it does what it says in this comment.

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -302,13 +454,26 @@ def _find_languages(env, img, pattern_center, line_vals):
     :param pattern_center: Center position of detected pattern
     :param dict line_vals: Dictonnary containing the data for a line
     (and the template)
+    :param bool test: Save the image of all the subimage
```
ecino: Again bad name or bad comment. Comment should be `tells if the import is done in test mode` or variable name should be `save_images`

sbc_compassion/tools/import_letter_functions.py
loikki: I have found a small mistake in it, I am rewriting it

sbc_compassion/wizards/import_letters_test.py
ecino: Yes

sbc_compassion/tools/import_letter_functions.py
ecino: I think you don't need test. See my comment at the return line.

sbc_compassion/tools/patternrecognition.py
```diff
@@ -120,7 +120,7 @@ def findMatches(des1, des2, test=0.8):
     :rtype: list[Keypoints]
     """
     bf = cv2.BFMatcher()
-    matches = bf.knnMatch(des2, des1, k=2)
+    matches = bf.knnMatch(des1, des2, k=2)
```
ecino: Why this swap, can you explain the change?

sbc_compassion/tools/patternrecognition.py
```diff
@@ -82,7 +82,7 @@ def patternRecognition(image, pattern, crop_area=None,
         return kp1, kp2, good
     if len(good) >= threshold:
         # put in a np.array the position of the image's keypoints
-        keypoints = np.array([kp1[i[0].trainIdx].pt for i in good])
+        keypoints = np.array([kp1[i[0].queryIdx].pt for i in good])
```
ecino: Can you explain this change ?

sbc_compassion/tools/patternrecognition.py
```diff
@@ -167,7 +167,7 @@ def keyPointCenter(keypoints):
         return center / N
 
 
-def find_template(img, templates, threshold=0.8):
+def find_template(img, templates, test=False, threshold=0.8):
```
ecino: you forgot to update the method docstring.

sbc_compassion/tools/patternrecognition.py
```diff
@@ -197,5 +215,7 @@ def find_template(img, templates, threshold=0.8):
                 len(tmp_key))/float(template.nber_keypoints)
             key_img = tmp_key
             matching_template = template
-
-    return matching_template, keyPointCenter(key_img)
+    if test:
```
ecino: Why not always return test_img (can be none if not called with test) ?

sbc_compassion/tools/import_letter_functions.py
loikki: Done

sbc_compassion/wizards/import_letters_test.py
loikki: I forgot to answer to the question. Yes it can (it is easy to deal with it in the code and the user will not have to create a new csv every time he wants to mix some test files together)

sbc_compassion/tools/import_letter_functions.py
loikki: I need to check with rz-comp

sbc_compassion/tools/patternrecognition.py
loikki: I had some trouble with the images for the test and usually this order is given (and seems more logical). I do not know exactly the difference, but I think I have done it wrong before this modification

sbc_compassion/tools/patternrecognition.py
loikki: Same as the comment above.
As I have done a switch between des1 and des2, I need to take the other Id (query is for one image and train the other one).

sbc_compassion/tools/patternrecognition.py
loikki: Done

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,134 @@
...
 ##########################################################################
+def testline(env, line_vals, csv_file_ids, doc_name):
+    """ Test each line and verify the data given by the csv.
+    If a line does not correspond to the csv, this function erases it.
+    A csv file need to be written in the same way than
+    sbc_compassion/tests/testdata/import/travis_file.csv
+
+    :param env: Odoo environment
+    :param dict line_vals: Dictionnary that will be used to create line
+    :param Recordset('test.import.letter.line') csv_file_ids: CSV used to check
```
ecino: In fact after checking, csv_file_ids are of type `ir.attachement`, not `test.import.letter.line`

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,134 @@
...
+    # check the letters
+    values = {}
+    for csv_line in csv_data:
+        if csv_line['name'] == doc_name:
+            values = csv_line
+            break
+    if not values:
+        raise Warning(
+            _("File '{}' not found in the csv files".format(doc_name)))
+    # first test: template
```
ecino: `Test that the template is the same as given in the csv`

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,134 @@
...
+    # str that will be returned
+    error = ''
+    csv_data = []
+    # creates a list containing all the csv data
+    for csv_ in csv_file_ids:
+        if os.path.splitext(csv_.name)[1] != '.csv':
+            raise Warning(_("File '{}' is not a csv file".format(csv_.name)))
+        csv_string = base64.b64decode(csv_.datas).splitlines()
+        csv_data += read_csv(csv_string)
+    # check the letters
```
ecino: `Check that the letter is present in the csv file`

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,134 @@
...
+    values = {}
+    for csv_line in csv_data:
+        if csv_line['name'] == doc_name:
+            values = csv_line
+            break
+    if not values:
```
ecino: Instead of this, you can use a `for.. else` loop (check the syntax)

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,134 @@
...
+        raise Warning(
+            _("File '{}' not found in the csv files".format(doc_name)))
+    # first test: template
+    template_id = env.ref(
+        'sbc_compassion.'+values['template_id'])
+    tpl_detected = env['sponsorship.correspondence.template'].search(
+        [('id', '=', line_vals['template_id'])])
+    # test if the template detected is the same as the expected one
```
ecino: You can remove this comment if the previous is updated.

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,134 @@
...
+    values = {}
+    for csv_line in csv_data:
+        if csv_line['name'] == doc_name:
+            values = csv_line
+            break
+    if not values:
+        raise Warning(
+            _("File '{}' not found in the csv files".format(doc_name)))
+    # first test: template
+    template_id = env.ref(
```
ecino: `template_csv`

loikki: template_csv_id?

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,134 @@
...
+        'sbc_compassion.'+values['template_id'])
+    tpl_detected = env['sponsorship.correspondence.template'].search(
+        [('id', '=', line_vals['template_id'])])
+    # test if the template detected is the same as the expected one
+    if template_id != tpl_detected:
```
ecino: I think it is safer to compare their ids instead of the entire objects:
`template_csv.id != tpl_detected.id`

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,134 @@
...
+        error = error + 'template,\n'
+
+    # second: qr codexb
```
ecino: Please write sentences in the comments.
`Test that the barcode detection matches the result given in csv file.`

sbc_compassion/tools/import_letter_functions.py
ecino: Sorry for those rules but `_id` suffix is for field names only (because they end up to be rows in the database storing integers). However, for only variable names, it makes less sense to suffix with `_id` as the variable will be a record and not an integer value.

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,133 @@
...
+    temp_test = None
+    if (line_vals['child_id'] is None and line_vals['partner_id'] is None and
+            values['barcode'] == ""):
+        temp_test = True
+    # 2) we check that the partner and the child are the same for the computed
+    # and the expected values
+    elif (partner_id.id == line_vals['partner_id'] and
+          child_id.id == line_vals['child_id']):
+        temp_test = True
+    # 3) if 2 does not work, it means that something was wrong with the
```
ecino: So you don't have the scenario when only one information is missing (either child or sponsor) ?

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,133 @@
...
+        error = error + 'template,\n'
+
+    # second: qr codexb
+    partner_id = env['res.partner'].search([
+        ('id', '=', line_vals['partner_id'])])
+    child_id = env['compassion.child'].search([
+        ('id', '=', line_vals['child_id'])])
+    # 1) we check the case where the barcode was not detected and was not
+    # expected
+    temp_test = None
```
ecino: You could define a function that decodes the barcode data into partner_ref and child_code values (can be used both in the import process and in this test). Here you could then do something like this, which would be more easy:
```
partner_csv, child_csv = decode_sponsorship_barcode(values['barcode'])
if partner_csv != partner.ref or child_csv != child.code:
    error = error + 'barcode,\n'
```

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,133 @@
...
+    # Test that the template is the same as given in the csv
+    template_id = env.ref(
+        'sbc_compassion.'+values['template_id'])
+    tpl_detected = env['sponsorship.correspondence.template'].search(
+        [('id', '=', line_vals['template_id'])])
+    if template_id != tpl_detected:
+        error = error + 'template,\n'
+
+    # second: qr codexb
+    partner_id = env['res.partner'].search([
```
ecino: rename to `partner_detected`

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,133 @@
...
+        ('id', '=', line_vals['partner_id'])])
+    child_id = env['compassion.child'].search([
```
ecino: `child_detected`

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,133 @@
...
+    # 3) if 2 does not work, it means that something was wrong with the
+    # detection
+    else:
+        temp_test = False
+    if not temp_test:
+        error = error + 'barcode,\n'
+
+    # third: languages
+    temp_test = True
+    for lang in line_vals['test_letter_languages_id'].split(','):
```
ecino: Since you use twice `line_vals['test_letter_languages_id'].split(',')`, you can put in in a variable before.

sbc_compassion/tools/import_letter_functions.py
loikki: If I have a partner missing, partner_id.id == False and line_vals['partner_id'] == False too, therefore the test will be OK

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -159,15 +274,42 @@ def _find_qrcode(env, img, MULTIPAGE, file_,
...
+def decodeBarcode(env, barcode):
+    """ Split the barcode and return the id of the partner and the child.
+    If the partner is not found, return None (same for the child).
+    :param env: Odoo environment
+    :param Barcode barcode:
+    :return: partner.id, Child.id
+    :rtype: int, int
+    """
+    if barcode is not None and 'XX' in barcode.data:
+        partner_id, child_id = barcode.data.split('XX')
```
ecino: If `barcode` contains more than one 'XX', split will return an array with len() > 2 and will break the code. You should test len(split()) == 2.
Also rename `partner_id`and `child_id` to `partner_ref` and `child_code` and later when you search you can use `partner_id` and `child_id`

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -159,15 +274,42 @@ def _find_qrcode(env, img, MULTIPAGE, file_,
...
+            'qr_preview': test_data
+        })
+    return dict_odoo, data, img
+
+
+def decodeBarcode(env, barcode):
+    """ Split the barcode and return the id of the partner and the child.
+    If the partner is not found, return None (same for the child).
+    :param env: Odoo environment
+    :param Barcode barcode:
```
ecino: Why is barcode a Tuple and not a simple string ?

loikki: It is not a Tuple, but a class (see zxing.Barcode).
In the test case, I am just using a namedtuple in order to mimic the behavior of the barcode

sbc_compassion/tests/test_tools.py
```diff
@@ -60,8 +60,8 @@ def test_pattern_recognition(self):
         self.assertIsNotNone(template)
         self.assertEqual(template.name, 'Test Template 2')
         np.testing.assert_allclose(pattern_center,
-                                   np.array([1273.47, 3298.65]),
-                                   atol=0.1)
+                                   np.array([1273, 3301]),
```
ecino: @rz-comp can you validate this change ?

sbc_compassion/tests/test_tools.py
```diff
@@ -121,7 +121,8 @@ def _read_img(self, path):
 
     def _pattern_recognition(self, path):
         img = self._read_img(path)
-        return tools.patternrecognition.find_template(img, self.templates, 0.3)
+        return tools.patternrecognition.find_template(
+            img, self.templates, 0.3)[:2]
```
ecino: I'm not fan of this, as it seems a bit "magic" to cut the result. By reading the code we don't understand what's going on. Can you return all data and just ignore what is not needed in the tests ?

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,122 @@
...
+    """ Test each line and verify the data given by the csv.
+    If a line does not correspond to the csv, this function erases it.
+    A csv file need to be written in the same way than
+    sbc_compassion/tests/testdata/import/travis_file.csv
+
+    :param env: Odoo environment
+    :param dict line_vals: Dictionnary that will be used to create line
+    :param Recordset('ir.attachment') csv_file_ids: CSV used to check
+    :param str doc_name: Document name
+    :returns: test (True if equal to csv), error
```
ecino: Update return doc

sbc_compassion/tests/test_tools.py
loikki: If I am doing it, I will always ignore the last parameter (it will always be None).
Maybe it will be better if I am just explaining in a comment before returning the values

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,122 @@
...
+        if csv_line['name'] == doc_name:
+            values = csv_line
+            break
+    else:
+        raise Warning(
+            _("File '{}' not found in the csv files".format(doc_name)))
+    # Test that the template is the same as given in the csv
+    template_csv = env.ref(
+        'sbc_compassion.'+values['template_id'])
+    tpl_detected = env['sponsorship.correspondence.template'].search(
```
ecino: I think you don't need to search for the template. You can simply do:
`if template_csv.id != line_vals['template_id']:`

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,122 @@
...
+    if template_csv.id != tpl_detected.id:
+        error = error + 'template,\n'
+
+    # Test that the barcode detection matches the result given in csv file
+    BarcodeTuple = namedtuple('BarcodeTuple', ['data'])
+    barcode = BarcodeTuple(values['barcode'])
+    partner_detected, child_detected = decodeBarcode(env, barcode)
+    # 1) we check the case where the barcode was not detected and was not
+    # expected
+    if (line_vals['child_id'] != child_detected or
```
ecino: Will `line_vals` always contain the `child_id` key ? If not, it should be replaced by `line_vals.get('child_id')` to avoid KeyErrors

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,122 @@
...
+            line_vals['partner_id'] != partner_detected):
+        error = error + 'barcode,\n'
+
+    # third: languages
+    temp_test = True
+    letter_languages_id = line_vals['test_letter_languages_id'].split(',')
```
ecino: What if `line_vals['test_letter_languages_id']` is empty ? split will raise an error

sbc_compassion/tools/import_letter_functions.py
loikki: If line_vals comes from my analyze attachment (which is the case actually), I will always have theses keys

sbc_compassion/tools/import_letter_functions.py
loikki: See comments above about the fact that this value exists.
Split does not raise an error when the string is empty.

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,122 @@
...
+    barcode = BarcodeTuple(values['barcode'])
+    partner_detected, child_detected = decodeBarcode(env, barcode)
+    # 1) we check the case where the barcode was not detected and was not
+    # expected
+    if (line_vals['child_id'] != child_detected or
+            line_vals['partner_id'] != partner_detected):
+        error = error + 'barcode,\n'
+
+    # third: languages
+    temp_test = True
```
ecino: rename to `lang_ok`

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,122 @@
...
+    BarcodeTuple = namedtuple('BarcodeTuple', ['data'])
+    barcode = BarcodeTuple(values['barcode'])
+    partner_detected, child_detected = decodeBarcode(env, barcode)
+    # 1) we check the case where the barcode was not detected and was not
+    # expected
+    if (line_vals['child_id'] != child_detected or
+            line_vals['partner_id'] != partner_detected):
+        error = error + 'barcode,\n'
+
+    # third: languages
```
ecino: Test that languages detected are the same as given in the csv file

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,122 @@
...
+        temp_test = temp_test and lang in values['lang']
+    line_nber_lang = len(letter_languages_id)
+    csv_nber_lang = len(values['lang'].split(','))
+    temp_test = temp_test and line_nber_lang == csv_nber_lang
+
+    if not temp_test:
+        error = error + 'lang,\n'
+
+    # remove the two last character (',\n')
+    return error[:-2]
```
ecino: If there are no errors, this will fail

loikki: No it will not

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -158,46 +273,94 @@ def _find_qrcode(env, img, MULTIPAGE, file_,
...
+    """
+    if barcode is not None and 'XX' in barcode.data:
+        barcode_split = barcode.data.split('XX')
+        if len(barcode_split) == 2:
+            partner_ref, child_code = barcode_split
+            child_id = env['compassion.child'].search(
+                [('code', '=', child_code)]).id
+            partner_id = env['res.partner'].search(
+                [('ref', '=', partner_ref),
+                 ('is_company', '=', False)]).id
```
ecino: You can add a condition, `contract_ids != False` to be sure only one partner will be found (sometimes we have duplicates in production (due to employees creating twice the same partner).

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -158,46 +273,94 @@ def _find_qrcode(env, img, MULTIPAGE, file_,
...
+        else:
+            partner_id = None
```
ecino: You can initialize partner_id and child_id to None at the beginning and remove those else parts

sbc_compassion/tests/test_tools.py
ecino: If its always None then you need to update `find_template` method and remove that return.

sbc_compassion/tools/import_letter_functions.py
ecino: Fine

sbc_compassion/tools/import_letter_functions.py
ecino: Ok :wink:

sbc_compassion/tools/import_letter_functions.py
ecino: Fine

sbc_compassion/tools/import_letter_functions.py
loikki: Like this?
```python
partner_id = env['res.partner'].search(
    [('ref', '=', partner_ref),
    ('is_company', '=', False),
    ('contract_ids', '!=', False)]).id
```

sbc_compassion/tests/test_tools.py
loikki: In this case, it is always None (you need to give the parameter test=True to obtain a non-None value)

sbc_compassion/tools/import_letter_functions.py
ecino: Sorry I think this won't work after checking. We could then check it has the "Sponsor" category but this is more complex than one search.

sbc_compassion/tests/test_tools.py
ecino: Ok then add a comment.

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,119 @@
 import checkboxreader as cbr
 import bluecornerfinder as bcf
 import numpy as np
+from math import ceil
 from glob import glob
 from wand.image import Image
+from collections import namedtuple
+
+from openerp import _
```
ecino: Add import exceptions.Warning

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -21,13 +21,119 @@
...
 
 
 ##########################################################################
 #                           GENERAL METHODS                              #
 ##########################################################################
+def testline(env, line_vals, csv_file_ids, doc_name):
+    """ Test each line and verify the data given by the csv.
+    A csv file need to be written in the same way than
+    sbc_compassion/tests/testdata/import/travis_file.csv
```
ecino: No, it is not the same way. In travis_file, there is no 'barcode' column, but in this csv there should be one, otherwise the test fails.

ecino:
Languages cut result should be shown even if no language is detected (if template is detected)

sbc_compassion/tools/import_letter_functions.py
loikki: I have corrected the csv in the test

sbc_compassion/tests/test_tools.py
rz-comp: :+1:

ecino:
:+1:

PR #160, opened 2015-12-16
Doc + Update the csv in order to use the same as in the test menu

loikki:
With a small update of the languages detection


loikki:
Do not merge this pull request before the end of the doc!

loikki:
Ready =)

PR #163, opened 2016-01-06
Update Onramp module

ecino:
Adapt to new information given by U.S., add image retrieval, fix bugs.


ecino:
@rz-comp I know you are working on the continuation of this, so I don't wan't to change those commits. But I can add another commit if you suggest some code corrections before merging to sbc-devel. Then you can rebase your work on this.

I should have corrected the bugs on the wrong message types and adapted to U.S. documentation. I will ask them to send some messages in order to test.

PR #164, opened 2016-01-07
Add field in the view letter

eicher31:
Add a check box in the view letter, check it if is the first letter from the children.

### DIFF
--- a/sbc_compassion/models/sponsorship_correspondence.py
+++ b/sbc_compassion/models/sponsorship_correspondence.py
@@ -99,6 +99,10 @@ class SponsorshipCorrespondence(models.Model):
         ('Sponsor', _('Sponsor')),
         ('Encourager', _('Encourager'))], default='Sponsor')
     mandatory_review = fields.Boolean()
+    is_first_letter = fields.Boolean(
+        compute='_compute_is_first',
+        store=True,
+        readonly=True)
     rework_reason = fields.Char()
     rework_comments = fields.Text()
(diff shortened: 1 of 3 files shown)


sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -179,6 +183,21 @@ def _compute_states(self):
...
+        """ Sets the value true if is the first letter. """
+        for letter in self:
+            if letter.sponsorship_id:
+                count = self.search_count([
+                    ('sponsorship_id', '=', letter.sponsorship_id.id),
+                    ('direction', '=', "Beneficiary To Supporter")
+                    ])
+                if count == 0:
+                    letter.is_first_letter = True
+                    break
```
ecino: I would rather remove this `break` and remove line 199. By default, boolean fields are always set to False. This makes sure you will treat all the letters contained in `self`. With the break statement, you could skip some letters.

sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -179,6 +183,21 @@ def _compute_states(self):
             else:
                 letter.b2s_state = letter.state
 
+    @api.multi
+    @api.depends('sponsorship_id')
+    def _compute_is_first(self):
+        """ Sets the value true if is the first letter. """
```
ecino: `True if it is the first Beneficiary to Supporter letter`

ecino:
:+1:

PR #167, opened 2016-01-13
Download child letters and send emails to sponsors on commkit update

rz-comp:
- download letter on reception of commkit message
- store letter in database and make it accessible at: /b2s_letter?id={uuid}
- add SendGrid module that allows sending emails through the SendGrid web service


ecino:
Why do you create an object `hostedletter` instead of using directly `sponsorship.correspondence` ? `sponsorship.correspondence` has already a field letter_image which points to an `ir.attachment` record. Wouldn't it be easier to store the letter image here ?

onramp_compassion/models/sponsorship_correspondence.py
```diff
@@ -100,6 +119,9 @@ def _commkit_update(self, data, message_id=None):
         else:
             commkit = self.with_context(from_onramp=True).create(commkit_vals)
 
+        # Send email to sponsor
+        commkit.send_email()
```
ecino: I think that you forgot to move this call in switzerland repository, don't you ?

sendgrid/README.rst
```diff
@@ -0,0 +1,43 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Compassion SendGrid
```
ecino: Not sure we should add our name in the module name (we could publish this module if generic enough)

sendgrid/__openerp__.py
```diff
@@ -0,0 +1,45 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+
+{
+    'name': 'Compassion SendGrid',
```
ecino: I think you can remove Compassion from the name of the module

sendgrid/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+#
+#    Copyright (C) 2015 Compassion CH (http://www.compassion.ch)
+#    Releasing children from poverty in Jesus' name
+#    @author: Roman Zoller
+#
+#    The licence is in the file __openerp__.py
+#
+##############################################################################
+
+from . import model
```
ecino: `models` is the new convention

sendgrid/model/email.py
```diff
@@ -0,0 +1,106 @@
...
+##############################################################################
+
+from openerp import models, fields, api, exceptions, _
+from openerp.tools.config import config
+
+import logging
+import sendgrid
+
+
+SUBSTITUTION_PREFIX = '{'
```
ecino: This means someone has to change the code if he wants to change the prefix and postfix. I think it would be better ton include this in the system parameters.

sendgrid/model/email.py
```diff
@@ -0,0 +1,106 @@
...
+            # From SendGrid documentation: All messages to this domain are
+            # accepted for delivery and then immediately deleted.
+            test_address = 'odoo@sink.sendgrid.net'
+
+        sg = sendgrid.SendGridClient(api_key)
+        message = sendgrid.Mail()
+
+        message.set_from(from_address)
+
+        subject = self.subject if self.subject else ' '
```
ecino: `self.subject or ' '`

sendgrid/model/template.py
```diff
@@ -0,0 +1,56 @@
...
+
+    ##########################################################################
+    #                                 FIELDS                                 #
+    ##########################################################################
+    name = fields.Char()
+    remote_id = fields.Char()
+
+    @api.one
+    def update(self):
+        # TODO: remove duplication
```
ecino: you mean `code duplication`?

ecino:
Please fix PEP8 errors

onramp_compassion/controllers/b2s_image.py
```diff
@@ -31,9 +31,11 @@ def handler_b2s_image(self):
         headers = request.httprequest.headers
         self._validate_headers(headers)
         correspondence_obj = request.env['sponsorship.correspondence'].sudo()
-        test = correspondence_obj.browse(10)
-        response = Response(base64.b64decode(test.letter_image.datas),
-                            mimetype='application/pdf')
+        correspondence = correspondence_obj.search([('uuid', '=', id)])
+        correspondence.last_read = fields.Datetime.now()
+        correspondence.read_count += 1
```
ecino: You could use the `write` method instead in order to make one single database query instead of two. Moreover, I think `+=` doesn't work in python.

onramp_compassion/models/sponsorship_correspondence.py
```diff
@@ -93,6 +93,21 @@ def _commkit_update(self, data, message_id=None):
         SponsorshipCorrespondence object. """
         letter_mapping = mapping.new_onramp_mapping(self._name, self.env)
         commkit_vals = letter_mapping.get_vals_from_connect(data)
+
+        # Download and store letter
```
ecino: You should only download letter image of Published letters, because we can receive notifications before the letter is translated. So you can put your code inside :
```
if commkit_vals.get('state') == 'Published to Global Partner':
```

onramp_compassion/models/sponsorship_correspondence.py
ecino: Please remove this call and put it inside switzerland module sbc_email.

sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -314,3 +322,21 @@ def write(self, vals):
         if 'state' in vals:
             vals['status_date'] = fields.Date.today()
         return super(SponsorshipCorrespondence, self).write(vals)
+
+    ##########################################################################
+    #                             PUBLIC METHODS                             #
+    ##########################################################################
+
+    @api.one
+    def send_email(self):
```
ecino: Remove this method and put it only in switzerland module.

rz-comp: rename to process_letter

sendgrid/__openerp__.py
```diff
@@ -0,0 +1,45 @@
...
+#
+##############################################################################
+
+
+{
+    'name': 'SendGrid',
+    'version': '0.1',
+    'category': 'Other',
+    'author': 'Compassion CH',
+    'website': 'http://www.compassion.ch',
```
ecino: Add :
```
'external_dependencies': {
        'python': ['sendgrid'],
    },
```

onramp_compassion/controllers/b2s_image.py
```diff
@@ -31,9 +31,13 @@ def handler_b2s_image(self):
         headers = request.httprequest.headers
         self._validate_headers(headers)
         correspondence_obj = request.env['sponsorship.correspondence'].sudo()
-        test = correspondence_obj.browse(10)
-        response = Response(base64.b64decode(test.letter_image.datas),
-                            mimetype='application/pdf')
+        correspondence = correspondence_obj.search([('uuid', '=', id)])
```
ecino: It would be good to return a 404 error if no correspondence is found, and also return a 400 error if no id is provided

ecino:
:+1:

PR #170, opened 2016-01-29
Make SendGrid email model inherit from mail.mail

rz-comp:


sendgrid/models/email.py
```diff
@@ -26,25 +26,19 @@
...
-    subject = fields.Char()
-    body_html = fields.Text()
     body_text = fields.Text()
     sent_date = fields.Datetime(copy=False)
     substitution_ids = fields.One2many(
         'sendgrid.substitution', 'email_id', copy=True)
     layout_template_id = fields.Many2one('sendgrid.template')
     text_template_id = fields.Many2one('email.template')
+
+    # @TODO integrate with mail.mail
```
ecino: I think mail.mail already defined the state field so you can remove it. What you need to ensure is that state goes to `sent` when using your send method.

PR #171, opened 2016-02-02
Auto filter and menu badge

eicher31:
Add default filter to tree view correspondence letter 
Add menu bagde in menu Correspondence/Letters

### DIFF
--- a/sbc_compassion/models/sponsorship_correspondence.py
+++ b/sbc_compassion/models/sponsorship_correspondence.py
@@ -28,7 +28,7 @@ class SponsorshipCorrespondence(models.Model):
     """
     _name = 'sponsorship.correspondence'
 
-    _inherit = ['mail.thread']
+    _inherit = ['mail.thread', 'ir.needaction_mixin']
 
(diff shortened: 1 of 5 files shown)


ecino:
Currently you filter only on "published child letters". I would want to filter as well only those that has not been read yet (last_read is not set).

sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -343,3 +336,9 @@ def _get_read_url(self):
 
     def _get_uuid(self):
         return str(uuid.uuid4())
+
+    @api.model
+    def _needaction_domain_get(self):
+        domain = ['&', ('direction', '=', 'Beneficiary To Supporter'),
+                  ('state', '=', 'Published to Global Partner')]
```
ecino: You can move this method in onramp_compassion module and add last_read to the domain.
Note : '&' is not necessary, it is the default operator for multiple conditions.

ecino:
:+1:

PR #178, opened 2016-02-26
Mapping page

eicher31:
Add mapping on page when correspondence sending or receive.


onramp_compassion/mappings/correspondence_mapping.py
```diff
@@ -146,3 +136,22 @@ def _process_odoo_data(self, odoo_data):
             del odoo_data['correspondant_id']
             if sponsorship:
                 odoo_data['sponsorship_id'] = sponsorship.id
+        # Replace dict by a tuple for the ORM update/create
+        if 'page_ids' in odoo_data:
```
ecino: I think this is not generic enough, because we will have to do the same transformations for all one2many fields we have.
I would rather move this in `base_mapping` and change the test to check any field ending with `_ids`. This will include `page_ids` and work for any other field.

onramp_compassion/mappings/correspondence_mapping.py
```diff
@@ -146,3 +136,22 @@ def _process_odoo_data(self, odoo_data):
             del odoo_data['correspondant_id']
             if sponsorship:
                 odoo_data['sponsorship_id'] = sponsorship.id
+        # Replace dict by a tuple for the ORM update/create
+        if 'page_ids' in odoo_data:
+            for page in odoo_data['page_ids']:
+                odoo_data['page_ids'].remove(page)
```
ecino: It is a bad practice to modify the list being accessed by the for loop, and can end in unexpected behaviours. I would rather construct the converted value in a separate variable and at the end of the process re-assign the value of the field (`odoo_data['page_ids'] = new_value`)

onramp_compassion/mappings/correspondence_mapping.py
```diff
@@ -146,3 +136,22 @@ def _process_odoo_data(self, odoo_data):
             del odoo_data['correspondant_id']
             if sponsorship:
                 odoo_data['sponsorship_id'] = sponsorship.id
+        # Replace dict by a tuple for the ORM update/create
+        if 'page_ids' in odoo_data:
+            for page in odoo_data['page_ids']:
+                odoo_data['page_ids'].remove(page)
+                # if page_url already exist update it
+                try:
```
ecino: Instead of try...except I would do the search and test the result.
```
page = self.env[...].search(...)
if page:
   ...
else:
   ...
```

onramp_compassion/mappings/page_mapping.py
```diff
@@ -0,0 +1,52 @@
...
+        'OriginalText': None,
+        'EnglishTranslatedText': None,
+        'TranslatedText': None
+    }
+
+    def _process_connect_data(self, connect_data):
+        fields = (
+            'OriginalText', 'EnglishTranslatedText', 'TranslatedText')
+        for field in fields:
+            connect_data[field] = [connect_data[field]]
```
ecino: You can add the verification `if field in connect_data`

sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -56,6 +56,8 @@ class SponsorshipCorrespondence(models.Model):
         'sponsorship_correspondence_type_relation',
         'correspondence_id', 'type_id',
         'Communication type',
+        default=lambda self: [(4, self.env.ref(
+            'sbc_compassion.correspondence_type_supporter').id)],
```
ecino: You can remove those two lines (old version of code)

sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -259,6 +265,15 @@ def _set_partner_review(self):
             if letter.correspondant_id.mandatory_review:
                 letter.mandatory_review = True
 
+    @api.depends('page_ids')
+    def _compute_page(self):
+        self.original_text = ''
```
ecino: Put original_text and translated_text into a variable and assign `self.original_text = original_text` only at the end of the method. This will be more efficient.

sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -89,12 +91,16 @@ class SponsorshipCorrespondence(models.Model):
     destination_language_id = fields.Many2one(
         'res.lang.compassion', compute='_set_destination_language',
         inverse='_change_language', store=True)
-    original_text = fields.Text()
-    translated_text = fields.Text()
+    original_text = fields.Text(compute='_compute_page')
```
ecino: You can add an inverse method (`inverse='_set_page'`) so that we can create a page by typing the text in the view.

eicher31:
Fix comments

sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -11,6 +11,7 @@
 
 import magic
 import base64
+import pdb
```
ecino: Remove pdb

sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -259,6 +265,37 @@ def _set_partner_review(self):
...
                 letter.mandatory_review = True
 
+    @api.depends('page_ids')
+    def _compute_page(self):
+        self.original_text = self._get_original_text('original_text')
+        self.translated_text = self._get_original_text('translated_text')
+        self.nbr_pages = len(self.page_ids)
+
+    def _inverse_page(self):
+        pdb.set_trace()
```
ecino: Remove pdb

sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -259,6 +265,37 @@ def _set_partner_review(self):
...
+        if len(self.page_ids) > 0:
```
ecino: `if self.page_ids` is enough

sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -259,6 +265,37 @@ def _set_partner_review(self):
...
+                'translated_text': self._get_original_text('translated_text')})
+            self.page_ids[1:].unlink()
+        else:
+            self.page_ids.create(
+                {'sponsorship_correspondence_id': self.id,
+                 'original_text': self.original_text,
+                 'translated_text': self.translated_text})
+
+    def _get_original_text(self, source_text):
+        if len(self.page_ids) > 1:
```
ecino: You can simplify this function to this:
```
return '\n\n'.join(self.page_ids.mapped(source_text))
```

sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -259,6 +265,37 @@ def _set_partner_review(self):
             if letter.correspondant_id.mandatory_review:
                 letter.mandatory_review = True
 
+    @api.depends('page_ids')
+    def _compute_page(self):
```
ecino: Would `_compute_texts` be more appropriate?

sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -259,6 +265,37 @@ def _set_partner_review(self):
...
+        self.original_text = self._get_original_text('original_text')
+        self.translated_text = self._get_original_text('translated_text')
+        self.nbr_pages = len(self.page_ids)
+
+    def _inverse_page(self):
```
ecino: Either add `@api.one` decorator or add a loop `for correspondence in self` and compute the result for each letter (self can be a recordset of many letters)

sbc_compassion/models/sponsorship_correspondence.py
ecino: You don't need to keep your tests with len(pages), as the line I gave you should already handle each cases.

ecino:
Ok, I will make some corrections and merge it manually. Thanks :+1:

eicher31:
I understand only now... sorry!

PR #182, opened 2016-03-04
Fix bug, false name of the compute method

eicher31:


ecino:
I already fixed that one ;) You can merge the changes made in CompassionCH:master in your branch.

PR #187, opened 2016-03-14
Import letters on Imports and Done folder on NAS

MickSandoz:
- Import letter updated with multiple selection now possible.
- A module added for the multiple file selection (sync_mail_multi_attach)
- Letters are on "P:\Scan Lettres S2B\Imports" when letter's state is "Analazing"
 At the end of treatment, letters are moved on "P:\Scan Lettres S2B\Done" (state "Open" or "Done")
- Adding possibility to import letters directly from NAS folder ("P:\Scan Lettres S2B\Imports" currently).


ecino:
The changes should be done in `sbc_email` module. We should not copy an existing module.

PR #189, opened 2016-03-22
Fix bug when original text is empty

eicher31:
When you create a letter by the web interface leaving the original text empty, you get an error.

### DIFF
--- a/sbc_compassion/models/sponsorship_correspondence.py
+++ b/sbc_compassion/models/sponsorship_correspondence.py
@@ -280,7 +280,8 @@ class SponsorshipCorrespondence(models.Model):
 
     def _get_original_text(self, source_text):
         """ Gets the desired text (original/translated) from the pages. """
-        return '\n\n'.join(self.page_ids.mapped(source_text))
+        txt = self.page_ids.filtered(source_text).mapped(source_text)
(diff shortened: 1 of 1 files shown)


ecino:
It could be possible that first page doesn't contain any text (think of a picture), and following pages contain some text. In that case you would delete that text from the other pages.
Please describe in the pull request description what bug you are trying to fix and what is the scenario when this bugs happens.

sbc_compassion/models/sponsorship_correspondence.py
```diff
@@ -280,7 +280,13 @@ def _inverse_page(self):
 
     def _get_original_text(self, source_text):
         """ Gets the desired text (original/translated) from the pages. """
-        return '\n\n'.join(self.page_ids.mapped(source_text))
+        txt = self.page_ids.mapped(source_text)
```
ecino: what about ```txt = self.page_ids.filtered(source_text).mapped(source_text)``` ?
https://www.odoo.com/documentation/8.0/reference/orm.html#openerp.models.Model.filtered
You won't need all further checks.

eicher31: Yes! Thank you!

ecino:
:+1:

PR #190, opened 2016-03-31
Import letters improvements

ecino:
- Reduce disk writes
- Allow multiple letters in single document : cut by barcodes


ecino:
@MickSandoz Can you quickly review the code to check nothing bad is hidden in this big pull request?

ecino:
Then you will be able to rebase on this...

MickSandoz:
I checked, and I think it's OK like this

ecino:
Ok, thank you

PR #191, opened 2016-04-06
Refactor sponsorship_correspondence

ecino:
- sponsorship_correspondence becomes correspondence
- new import configuration objects
- correct import tests
- improve the import review wizard

@eicher31 I added the field boolean `translatable` in `res.lang.compassion`, used for telling if we can translate that language locally. Would that help you ?

@MickSandoz, @eicher31 Can you quickly review the code? There is a lot because I renamed a lot of objects remove the `sponsorship` prefix but you can quickly traverse that.


eicher31:
Its OK for me! Thanks for `translatable`!

sbc_compassion/i18n/fr_CH.po
```diff
@@ -159,21 +159,21 @@ msgstr "Correspondant"
...
-#: view:sponsorship.correspondence:sbc_compassion.view_sponsorship_correspondence_tree_details
+#: view:correspondence:sbc_compassion.view_sponsorship_correspondence_form
+#: view:correspondence:sbc_compassion.view_sponsorship_correspondence_tree
+#: view:correspondence:sbc_compassion.view_sponsorship_correspondence_tree_details
 msgid "Correspondence"
 msgstr "Correspondance"
 
 #. module: sbc_compassion
-#: view:sponsorship.crosscheck:sbc_compassion.view_sponsorship_correspondence_crosscheck
+#: view:correspondence.template.crosscheck:sbc_compassion.view_sponsorship_correspondence_crosscheck
```
MickSandoz: I think it's ok but, I expected to "correspondence.crosscheck" instead of "correspondence.**template**.crosscheck

ecino: The model is used to compare the templates and see if one is not too similar with another at the detection, so I think the name is more approriate.

MickSandoz:
Its OK for me too! Thanks for the explanation to my question ;-)

PR #195, opened 2016-04-12
bug fixes and simplify inheritance method redefinition

MickSandoz:
-  bug fixes in correspondence_metadata.py
- change a method in import_letters_history.py to simplify inheritance method redefinition

#CO-15

### DIFF
--- a/sbc_compassion/models/correspondence_metadata.py
+++ b/sbc_compassion/models/correspondence_metadata.py
@@ -42,6 +42,7 @@ class CorrespondenceMetadata(models.AbstractModel):
         """
         self.ensure_one()
         vals = self.read(self.get_fields())[0]
-        vals['template_id'] = vals['template_id'][0]
+        if vals.get('template_id'):
+            vals['template_id'] = vals['template_id'][0]
         del(vals['id'])
         return vals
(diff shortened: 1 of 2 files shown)


sbc_compassion/models/import_letters_history.py
```diff
@@ -244,7 +245,7 @@ def _run_analyze(self):
         self.import_completed = True
         logger.info("Imported letters analysis completed.")
 
-    def _analyze_attachment(self, file_data, file_name):
+    def _analyze_attachment(self, file_data, file_name, is_zipfile=False):
```
ecino: `is_zip_file` is never used inside the function, why adding it?

MickSandoz: I know it's my problem, I don't find the way to simplify method redefinition of analyze_attachment in the inherited code (sbc_emai) wihtout add this params.

Without that, redefinition of _run_analyze()(who call analyze_attachment) in sbc_email becomes bigger

ecino: Ok then...

sbc_compassion/models/correspondence_metadata.py
```diff
@@ -42,6 +42,7 @@ def get_correspondence_metadata(self):
         """
         self.ensure_one()
         vals = self.read(self.get_fields())[0]
-        vals['template_id'] = vals['template_id'][0]
+        if vals.get('template_id'):
```
ecino: Please refine the test to this:
`if isinstance(vals.get('template_id'), tuple):`

ecino: Sorry, you can ignore that. I think the simpler case works fine.

PR #196, opened 2016-04-13
Fix regression bug where template was not set on import lines

ecino:
- Revert language selector view in import review wizard
- Give a name to imports (date of import)
- Recompute import line status after review


MickSandoz:
Ok for me

PR #197, opened 2016-04-14
Letter composition

ecino:
Add method and button for letter composition
For now it adds a page to the letter, in the future it needs ton include translation inside page.

Fix bug of files too big at letter import.


MickSandoz:
Ok for me

PR #198, opened 2016-04-15
CO-16 Fix UpdateCommKits only in incomming message

eicher31:
Add required original language on correspondence's edit view


ecino:
:+1:

PR #201, opened 2016-04-18
Reduce RAM usage at letter import by using pypdf

ecino:
Pypdf is used to split the pdf and load image only one page at a time.
Tiff support is therefore removed !


MickSandoz:
Ok for me

PR #205, opened 2016-04-21
Various bug fixes and improvements

ecino:
CO-96
If supporter doesn't speak child language, set English by default.
CO-81
CO-79


MickSandoz:
Ok for me

PR #218, opened 2016-05-03
Import web letters in Odoo

MickSandoz:


sbc_compassion/models/import_letter_line.py
```diff
@@ -41,6 +41,7 @@ class ImportLetterLine(models.Model):
         ("no_child_partner", _("Partner or Child not Found")),
         ("no_template", _("Template not Detected")),
         ("ok", _("OK"))], compute="check_status", store=True, readonly=True)
+    original_text = fields.Char()
```
ecino: field.Text() is more appropriate

ecino:
:+1:

PR #219, opened 2016-05-10
Field "source" filled in import's letters

MickSandoz:
- move field "source" from correspondence to correspondence_data set by default to "Letter"
- set this field to "Compassion website" in import's letters when letters come from website.


ecino:
:+1:

PR #225, opened 2016-05-26
Update sponsorship_compassion

eicher31:
Move field "send_original" from module sbc_compassion into sponsorship_compassion.


ecino:
:+1:

PR #232, opened 2016-06-06
Disable batch process for gmc messages by default

ecino:
Change view for message center : add one menu for constituent message.
Fix partner must be updated before calling upsert message


ecino:
@eicher31 @MickSandoz Any comments ?

MickSandoz:
Ok for me

eicher31:
ok

PR #233, opened 2016-06-07
Redo and migrate service CreateCommitment and CanccelCommitment

MickSandoz:


message_center_compassion/models/gmc_message.py
```diff
@@ -255,7 +255,7 @@ def _perform_outgoing_action(self):
 
         object_mapping = mapping.new_onramp_mapping(action.model, self.env)
         if action.connect_outgoing_wrapper:
-            # Send multiple objects in a single message to GMC
+            # Send multiple objects in a single message to GMCs
```
ecino: I think this is not useful

sponsorship_compassion/mappings/cancel_sponsorship_mapping.py
```diff
@@ -0,0 +1,50 @@
...
+#    @author: Michael Sandoz <michaelsandoz87@gmail.com>
+#
+#    The licence is in the file __openerp__.py
+#
+##############################################################################
+from openerp.addons.message_center_compassion.mappings.base_mapping \
+    import OnrampMapping
+
+
+class CorrespondenceMapping(OnrampMapping):
```
ecino: Please rename class

sponsorship_compassion/mappings/cancel_sponsorship_mapping.py
```diff
@@ -0,0 +1,50 @@
...
+    """ This class contains the mapping between Odoo fields and GMC field names
+    for delete a sponsorship.
+
+    All fields are described at
+    http://developer.compassion.com/docs/read/compassion_connect2/
```
ecino: Remove link or replace with https://github.com/CompassionTCPT/R4-JSON/blob/0f03f87ba2b00ecd94a3407078ce06f06394e10c/EIM_Reviewed/CI_NS_BeneficiaryCommitmentCancellationRequest_schema.json
(but private github)

ecino:
You defined two different mappings for the same object `recurring.contract`. I am not sure how this will work as the message_center finds the mapping for an object by looking at `ODOO_MODEL` inside the mapping. It will take the first mapping found, so one of your mappings will never be used in my opinion.

Is it possible to merge the two mappings and use the same for both messages ? If not, I'm afraid we have to change how it is done, and instead of using `ODOO_MODEL`, maybe add a field inside the mapping like `GMC_ACTION = 'name of action'` and use that to use the correct mapping for each action.

sponsorship_compassion/models/contracts.py
```diff
@@ -521,6 +523,18 @@ def _on_sponsorship_finished(self):
                 else:
                     contract.signal_workflow('contract_terminated')
 
+            # Create Sponsorship Message
```
ecino: Cancel Sponsorship Message

sponsorship_compassion/models/contracts.py
```diff
@@ -521,6 +523,18 @@ def _on_sponsorship_finished(self):
                 else:
                     contract.signal_workflow('contract_terminated')
 
+            # Create Sponsorship Message
+            message_obj = self.env['gmc.message.pool']
+            action_id = self.env.ref(
+                'sponsorship_compassion.cancel_sponsorship').id
+
+            logger.info("DELETE MESSAGE CREATED")
```
ecino: Not sure why you log this?

sponsorship_compassion/models/contracts.py
```diff
@@ -521,6 +523,18 @@ def _on_sponsorship_finished(self):
...
 
+            # Create Sponsorship Message
+            message_obj = self.env['gmc.message.pool']
+            action_id = self.env.ref(
+                'sponsorship_compassion.cancel_sponsorship').id
+
+            logger.info("DELETE MESSAGE CREATED")
+            message_vals = {
+                'action_id': action_id,
+                'object_id': sponsorship.id
```
ecino: you can add `'partner_id': sponsorship.correspondant_id.id, 'child_id': sponsorship.child_id.id`

sponsorship_compassion/models/contracts.py
```diff
@@ -532,9 +546,23 @@ def contract_active(self):
...
+            partner.upsert_constituent()
+
+            # Create Sponsorship Message
+            message_obj = self.env['gmc.message.pool']
+            action_id = self.env.ref(
+                'sponsorship_compassion.create_sponsorship').id
+
+            messages = message_obj.search([
+                ('partner_id', '=', partner.id),
+                ('state', '=', 'new'),
```
ecino: Change with `('state', 'in', ('new', 'pending'))`. Can you as well change it in the method `upsert_constituent` ?

sponsorship_compassion/models/contracts.py
```diff
@@ -532,9 +546,23 @@ def contract_active(self):
...
-
-            # TODO : Create Sponsorship Message
+            partner.upsert_constituent()
+
+            # Create Sponsorship Message
+            message_obj = self.env['gmc.message.pool']
+            action_id = self.env.ref(
+                'sponsorship_compassion.create_sponsorship').id
+
+            messages = message_obj.search([
```
ecino: I have a better idea for this check : as it is general for each message (we don't wan't to create a new message if one similar does already exist), we can integrate it in the create method of object `gmc_message` and don't worry of that outside. Can you make the change ? We can discuss it if you need more precisions.

sponsorship_compassion/wizards/migration_r4.py
```diff
@@ -68,12 +68,16 @@ def _create_commitments(self, messages):
         Migrate the old CreateCommitment messages
         """
         logger.info("MIGRATION 8.0.3 ----> CreateCommitment")
+        messages.write({'action_id': self.env.ref(
+            'sponsorship_compassion.create_sponsorship').id})
```
ecino: You should remove the `unlink` calls on the `_perform_migration` method, otherwise those messages will be deleted.

message_center_compassion/models/gmc_message.py
MickSandoz: Not watch this change, sorry, i will remove it

sponsorship_compassion/models/contracts.py
MickSandoz: I forgot to remove, it was usefull during phase test

sponsorship_compassion/models/contracts.py
MickSandoz: Yes I do

sponsorship_compassion/models/contracts.py
MickSandoz: It's done now

message_center_compassion/models/gmc_action.py
```diff
@@ -32,6 +32,7 @@ class GmcAction(models.Model):
     """
     _name = 'gmc.action'
 
+    action = fields.Char()
```
ecino: Why not use the field `name` already present ? I don't feel this field is needed.

sponsorship_compassion/models/contracts.py
```diff
@@ -530,11 +545,16 @@ def contract_active(self):
         """
         for contract in self.filtered(
                 lambda c: 'S' in c.type and not c.is_active):
-            # UpsertConstituent Message
-            partner = contract.correspondant_id
-            partner.upser_constituent()
-
-            # TODO : Create Sponsorship Message
+            # Cancel Sponsorship Message
```
ecino: Should be Create Sponsorship

message_center_compassion/mappings/base_mapping.py
```diff
@@ -136,8 +137,12 @@ def get_connect_data(self, odoo_object, fields_to_submit=None):
                 # Field One2Many
                 if field.endswith('ids'):
                     value = list()
-                    sub_mapping = new_onramp_mapping(
-                        field_mapping[1], self.env)
+                    if field_mapping.__len__() == 2:
```
ecino: I don't understand this change and what you are doing here. ?

ecino:
Can you as well remove the code in `upsert_partner` method inside sponsorship_compassion/models/partner_compassion.py where I checked if message already exists? Because it is now done at creation of the message.

MickSandoz:
I don't find  /sponsorship_compassion/models/partner_compassion.py

message_center_compassion/mappings/base_mapping.py
eicher31: It's for relation one to many for specified the action. 
Example:
        'Pages': ('page_ids', 'correspondence.page', 'specified_Action'),
If specified_action is not give, it take by default 'default'

ecino:
Sorry it's res_partner.py

MickSandoz:
It's allready done
https://github.com/MickSandoz/compassion-modules/blob/073409ff2d22d2088f6875e649c89e56f2710479/sponsorship_compassion/models/res_partner.py

message_center_compassion/mappings/base_mapping.py
ecino: Can you then update the comment at line 46 ? And don't forget to update for incoming messages.

message_center_compassion/mappings/base_mapping.py
```diff
@@ -287,13 +297,16 @@ def itersubclasses(cls, _seen=None):
...
 
     :param model: model name of the mapping.
+    :param action: action name (not mandatory, specify when a model has many
+    actions)
     :return: class instance for given model mapping.
     """
     for cls in itersubclasses(OnrampMapping):
         if cls.ODOO_MODEL == model:
-            return cls(env)
+            if cls.ACTION == action:
```
ecino: could be done in the same test `if cls.ODOO_MODEL == model and cls.ACTION == action`

sponsorship_compassion/models/contracts.py
```diff
@@ -530,11 +545,16 @@ def contract_active(self):
         """
         for contract in self.filtered(
                 lambda c: 'S' in c.type and not c.is_active):
-            # UpsertConstituent Message
-            partner = contract.correspondant_id
-            partner.upser_constituent()
```
ecino: No, you should not remove this!

sponsorship_compassion/models/res_partner.py
```diff
@@ -9,6 +9,7 @@
 #
 ##############################################################################
 from openerp import api, fields, models, _
+from functools import reduce
```
ecino: Is this needed ? reduce is normally a builtin function, did you had an error?

MickSandoz: Yes, use on line 112 in method write (not code by me)

ecino:
:+1:

PR #234, opened 2016-06-07
CO-179 Migrate waiting messages UpsertConstituent

eicher31:


ecino:
You should remove the unlink calls on the _perform_migration method, otherwise those messages will be deleted.

eicher31:
Done

PR #236, opened 2016-06-08
Some fixes and improvements for R4

ecino:


ecino:
PEP8 should be fixed in #233

PR #237, opened 2016-06-09
Add global childpool and search message

ecino:


child_compassion/mappings/childpool_search_mapping.py
```diff
@@ -0,0 +1,51 @@
...
+#    The licence is in the file __openerp__.py
+#
+##############################################################################
+from openerp.addons.message_center_compassion.mappings.base_mapping import \
+    OnrampMapping
+
+
+class ChilpoolSearchMapping(OnrampMapping):
+    """ This class contains the mapping between Odoo fields and GMC field
+    names for the model Res Partner.
```
MickSandoz: I think it's not Res Partner instead Child compassion

MickSandoz:
OK

PR #238, opened 2016-06-10
"Beneficiary Rich Mix" message added and functional

MickSandoz:


child_compassion/mappings/childpool_search_mix_mapping.py
```diff
@@ -0,0 +1,33 @@
...
+    ODOO_MODEL = 'compassion.childpool.search'
+    ACTION = "rich_mix"
+
+    CONNECT_MAPPING = {
+        'numberOfBeneficiaries': 'take',
+    }
+
+    FIELDS_TO_SUBMIT = {k: None for k in CONNECT_MAPPING.iterkeys()}
+
+    def _process_connect_data(self, connect_data):
```
ecino: take is always set no ? So this method can be removed I think. Moreover, you could move that class in the same file as the search_mapping don't you think ?

child_compassion/mappings/childpool_search_mapping.py
```diff
@@ -17,6 +17,7 @@ class ChilpoolSearchMapping(OnrampMapping):
     Childpool mapping for searching available children from GMC.
     """
     ODOO_MODEL = 'compassion.childpool.search'
+    ACTION = "profil_search"
```
ecino: would rename `profile_search` for better english

child_compassion/wizards/global_child_search.py
```diff
@@ -92,4 +93,25 @@ def do_search(self):
...
+        # Remove previous search results
+        self.global_child_ids = False
+
+        mapping = base_mapping.new_onramp_mapping(self._name, self.env,
+                                                  "rich_mix")
+        params = mapping.get_connect_data(self)
+        onramp = OnrampConnector()
+        result = onramp.send_message(
+            'beneficiaries/richmix', 'GET', None, params)
+        if result['code'] == 200:
```
ecino: The following code could be refactored as it is almost identical to `do_search`

child_compassion/wizards/global_child_search.py
```diff
@@ -69,20 +69,16 @@ def do_search(self):
...
             self.nb_found = result['content']['NumberOfBeneficiaries']
-            children = result['content']['BeneficiarySearchResponseList']
-            mapping = base_mapping.new_onramp_mapping(
-                'compassion.global.child', self.env)
-            for child_data in children:
-                child_vals = mapping.get_vals_from_connect(child_data)
-                self.global_child_ids += self.env[
-                    'compassion.global.child'].create(child_vals)
+            self.map_and_create_json_oject(result,
+                                           'BeneficiarySearchResponseList')
```
ecino: I'm going in details but why not directly pass `result['content']['BeneficiarySearchResponseList']` to the function ?

child_compassion/wizards/global_child_search.py
```diff
@@ -92,4 +88,29 @@ def do_search(self):
...
+        if result['code'] == 200:
+            self.map_and_create_json_oject(result,
+                                           'BeneficiaryRichMixResponseList')
+        else:
+            raise Warning(
+                result.get('content', result)['Error'])
+
         return True
+
+    def map_and_create_json_oject(self, result, object_name):
```
ecino: could be private (_map_and_create...)

child_compassion/mappings/childpool_search_mapping.py
```diff
@@ -48,3 +49,17 @@ def _process_connect_data(self, connect_data):
...
+    Childpool mapping for searching mix of available children from GMC.
+    """
+    ODOO_MODEL = 'compassion.childpool.search'
+    ACTION = "rich_mix"
+
+    CONNECT_MAPPING = {
+        'numberOfBeneficiaries': 'take',
+    }
+
+    FIELDS_TO_SUBMIT = {k: None for k in CONNECT_MAPPING.iterkeys()}
```
ecino: since you have only one field, is it more simple to put it ? `'numberOfBeneficiaries': None`

PR #249, opened 2016-06-23
Update child descriptions in french for R4

ecino:
* Still project description to do
* Still description for german and italian to do


MickSandoz:
:+1:

PR #253, opened 2016-06-28
Update child and project descriptions + correct mapping for GetBeneficiary Details

ecino:


MickSandoz:
It seems OK, but I think it would be good to check descriptions (on the user interface) with SDS operators.

ecino:
Yes this will be a task because descriptions are not fully good. At least it doesn't fail to generate them.

PR #254, opened 2016-06-29
CO-206 Add Child Reservations

maxime-beck:


ecino:
./child_compassion/models/__init__.py:28:34: W292 no newline at end of file

child_compassion/models/project_compassion.py
```diff
@@ -170,6 +170,9 @@ class CompassionProject(models.Model):
         'icp.sociological.activity', 'icp_socio_ado_act',
         string='Sociological activities (12+)', readonly=True
     )
+    reservation_id = fields.Many2one(
+        'compassion.project.reservation', string='Project Reservation'
```
ecino: You can rename the model to `icp.reservation`

child_compassion/models/project_reservation.py
```diff
@@ -0,0 +1,42 @@
...
+#    The licence is in the file __openerp__.py
+#
+##############################################################################
+
+from openerp import models, fields
+
+
+class ProjectReservation(models.Model):
+
+    _name = 'compassion.reservation'
```
ecino: You can rename to `icp.reservation`

child_compassion/models/project_reservation.py
```diff
@@ -0,0 +1,42 @@
...
+    _description = 'Project Reservation'
+
+    beneficiary_global_id = None
```
ecino: You cannot declare fields like this. Plus you should rename this field (it's not a beneficiary, beneficiary is a child).
```
global_id = fields.Char()
```

child_compassion/models/project_reservation.py
```diff
@@ -0,0 +1,42 @@
...
+    channel_name = fields.Char(string='Channel Name', required=True, size=64)
```
ecino: Remove required property, size may not be useful as well. You are not obliged to put `string='Channel Name'`, as it will be this string by default. It is only helpful when you want to put another name than the field_name.

child_compassion/models/project_reservation.py
```diff
@@ -0,0 +1,42 @@
...
+    global_partner_id = 'CH'
```
ecino: Even if we will always send Switzerland, you cannot declare a field like this. However, we won't need this field in our object so you can completely remove it.

child_compassion/models/project_reservation.py
```diff
@@ -0,0 +1,42 @@
...
+    icp_id = fields.Many2one(
+        'compassion.project', string='Project Compassion', required=True
```
ecino: You don't need to specify the name of the parameter when you follow the order of the optional parameters (see the documentation for fields). So here you can omit `string=`. You can also just name it `Project`

child_compassion/models/project_reservation.py
```diff
@@ -0,0 +1,42 @@
...
+    )
+    campaign_event_identifier = fields.Char(string='Campaign Event '
```
ecino: You can remove all attributes, they are not needed for this field.

child_compassion/models/project_reservation.py
maxime-beck: I didn't meant to declare a field for global_id because according to the [1]specification, this field must always have a value of NULL for an ICP type of reservation.
--
[1] https://github.com/CompassionTCPT/R4-JSON/blob/master/Endpoints/ReadyForTest/GlobalPartners-ICPCreateUpdateReservation_POST.json#L8

ecino: Ah yes so you can completely remove this line. However you will need to add a field for `Reservation_ID`, because you will receive it in the answer and you want to store it.

maxime-beck: Alright !
I guess I can also remove all the static variables I declared for the same purpuse right ?
Meaning : `global_partner_id = 'CH'` `reservation_type = 'ICP'` and `source_code = ''`.

child_compassion/models/project_reservation.py
```diff
@@ -0,0 +1,32 @@
...
+from openerp import models, fields
+
+
+class ProjectReservation(models.Model):
+
+    _name = 'icp.reservation'
+    _description = 'Project Reservation'
+
+    reservation_id = fields.Char()
+    channel_name = fields.Char(required=True)
```
ecino: Remove required

child_compassion/models/project_reservation.py
```diff
@@ -0,0 +1,32 @@
...
+    icp_id = fields.Many2one(
+        'compassion.project', 'Project', required=True
+    )
+    campaign_event_identifier = fields.Char(required=True)
```
ecino: Remove required

child_compassion/models/project_reservation.py
```diff
@@ -0,0 +1,32 @@
...
+    expiration_date = fields.Date(required=True)
+    hold_expiration_date = fields.Date(required=True)
```
ecino: Should be a Datetime

child_compassion/models/project_reservation.py
```diff
@@ -0,0 +1,32 @@
...
+    hold_yield_rate = fields.Integer()
+    is_reservation_auto_approved = fields.Boolean(required=True)
```
ecino: Remove required and put default=True

child_compassion/views/project_reservation_view.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+	<data>
+		<!--Compassion project reservation form view -->
+		<record id="view_compassion_reservation_form"
+                model="ir.ui.view">
+			<field name="name">icp.reservation.form</field>
+			<field name="model">icp.reservation</field>
+			<field name="arch" type="xml">
+				<form string="Reservation">
```
ecino: Wrong indentation

child_compassion/models/project_compassion.py
```diff
@@ -170,6 +170,9 @@ class CompassionProject(models.Model):
         'icp.sociological.activity', 'icp_socio_ado_act',
         string='Sociological activities (12+)', readonly=True
     )
+    reservation_id = fields.Many2one(
```
ecino: You can as well add the field in the view of the project. You can add it like this :
```
<field name="reservation_id" context="{'default_project_id': active_id}"/>
```
so that the project_id field is already set when creating a reservation from the project.

child_compassion/views/project_reservation_view.xml
```diff
@@ -0,0 +1,49 @@
...
+        </record>
+
+        <!-- Action opening the tree view -->
+        <record id="open_view_reservation_form" model="ir.actions.act_window">
+            <field name="name">Reservation</field>
+            <field name="res_model">icp.reservation</field>
+            <field name="view_type">form</field>
+            <field name="view_mode">form</field>
+            <field name="view_id" ref="view_compassion_reservation_form"/>
+            <field name="context">{'search_default_available':1}</field>
```
ecino: You can remove this, it is useful for activating a filter on the view, but we don't need it for reservations.

child_compassion/views/project_reservation_view.xml
```diff
@@ -0,0 +1,49 @@
...
+            </field>
+        </record>
+
+        <!-- Action opening the tree view -->
+        <record id="open_view_reservation_form" model="ir.actions.act_window">
+            <field name="name">Reservation</field>
+            <field name="res_model">icp.reservation</field>
+            <field name="view_type">form</field>
+            <field name="view_mode">form</field>
+            <field name="view_id" ref="view_compassion_reservation_form"/>
```
ecino: You can remove it, it is useful for specifying a view when there are multiple views defined for an object.

child_compassion/views/project_reservation_view.xml
```diff
@@ -0,0 +1,49 @@
...
+                </form>
+            </field>
+        </record>
+
+        <!-- Action opening the tree view -->
+        <record id="open_view_reservation_form" model="ir.actions.act_window">
+            <field name="name">Reservation</field>
+            <field name="res_model">icp.reservation</field>
+            <field name="view_type">form</field>
+            <field name="view_mode">form</field>
```
ecino: `<field name="view_mode">tree,form</field>`
so that it opens the tree view by default.

ecino:
:+1:

child_compassion/models/project_reservation.py
```diff
@@ -0,0 +1,32 @@
...
+    _description = 'Project Reservation'
+
+    reservation_id = fields.Char()
+    channel_name = fields.Char()
+    icp_id = fields.Many2one(
+        'compassion.project', 'Project', required=True
+    )
+    campaign_event_identifier = fields.Char()
+    expiration_date = fields.Date(required=True)
+    hold_expiration_date = fields.DateTime(required=True)
```
MickSandoz: Datetime() !!!

PR #258, opened 2016-07-01
[ADD] Module for partner communication configuration

ecino:
* New module for setting rules for partner communication
* Only e-mails are handled now, printing paper does nothing
* Standardize and move preference communication fields (they will be in compassion-switzerland modules because they concern only switzerland decisions)
* Some bug fixing in handling incoming messages

Before integring changes be careful not to forget to :

* Install module partner_communication
* Upgrade module message_center_compassion


MickSandoz:
:+1:

message_center_compassion/tools/onramp_connector.py
```diff
@@ -142,7 +142,7 @@ def _retrieve_token(self):
             "Expect": "100-continue",
             "Connection": "Keep-Alive"}
         conn = httplib.HTTPSConnection('api2.compassion.com')
-        conn.request("POST", "/pcore/connect/token", params_post, header_post)
+        conn.request("POST", "/core/connect/token", params_post, header_post)
```
MickSandoz: Have we not said we keep Stage configuration on CompassionCH/master?

PR #259, opened 2016-07-05
Add hold exportation date for update a hold when canceling sponsorship

MickSandoz:


contract_compassion/wizard/end_contract_wizard.py
```diff
@@ -65,6 +66,12 @@ def end_contract(self):
         self.ensure_one()
         contract = self.contract_id
 
+        # if not self.has_new_hold:
+        """ Warning maybe to change when R4 will be stable...
+        # Currently Connect refuse sponsorship cancel without
+        expiration hold date"""
```
ecino: This won't change apparently. So the behaviour is to send the current datetime and ignore the hold id returned that will expire very quickly. Here is the answer from the U.S.:
```
When you cancel a commitment, the beneficiary must be put on hold, at least momentarily. This was because most Partners keep the beneficiary on Sponsor Cancel Hold or No Money Hold for some time to try and maintain the sponsorship.

If you don’t wish to use this, set the hold expiration date to the current time. You’ll receive a Hold ID but you should be able to ignore it.
```

contract_compassion/wizard/end_contract_wizard.py
```diff
@@ -32,8 +32,9 @@ class EndContractWizard(models.TransientModel):
     do_transfer = fields.Boolean('I want to transfer the child')
     transfer_country_id = fields.Many2one(
         'compassion.global.partner', 'Country')
+    has_new_hold = fields.Boolean(default=True)
```
ecino: `has_new_hold = fields.Boolean('Keep child on hold', default=fields.Datetime.now())`
in order to give a more comprehensible name in the view and to set by default to the current time.

ecino: Can you as well change the default value of field `end_date`? `fields.Datetime.now()` is better than using `datetime` library.

sponsorship_compassion/mappings/cancel_sponsorship_mapping.py
```diff
@@ -31,10 +30,69 @@ class CancelSponsorship(BaseSponsorshipMapping):
...
+                              relation_search=None):
+        if connect_name and connect_name in ['Beneficiary_GlobalID',
+                                             'HoldID']:
+            self.params[connect_name] = value
+
+        return super(CancelSponsorship, self)._convert_connect_data(
+            connect_name, value_mapping, value, relation_search=None)
+
+    def _process_odoo_data(self, connect_data):
+        if 'Beneficiary_GlobalID' in self.params:
```
ecino: My proposition is to :

1. Before sending the message, when the wizard is executed to cancel the sponsorship, verify if the user wants to put a hold. If no, directly set the hold on the child (which is the hold that was set before the child is sponsored to inactive). If yes, keep the hold active.
2. Change base_mapping of field 'HoldId' to `'HoldID': 'hold_id'` (even if the field does not exist in recurring_contract), just that you can retrieve the value in function `_process_odoo_data`
3. remove `self.params` and change that test to 
```
if 'child_id' in connect_data:
    child = self.env['compassion.child'].browse(connect_data['child_id'])
    if child.hold_id.active:
        # Update new hold
        child.hold_id.hold_id = connect_data['hold_id']
        # Don't write hold_id in contract
        del connect_data['hold_id']
    else:
        child.active = False
```

contract_compassion/wizard/end_contract_wizard.py
```diff
@@ -64,6 +63,16 @@ def fields_view_get(self, cr, user, view_id=None, view_type='form',
     def end_contract(self):
         self.ensure_one()
         contract = self.contract_id
+        child = contract.child_id
+        hold = child.hold_id
+
+        if not self.has_new_hold:
+            # child will be inactive after this choice
+            self.hold_expiration_date = fields.Datetime.now()
```
ecino: This is not very useful because it is already the default value of the field...

sponsorship_compassion/mappings/cancel_sponsorship_mapping.py
```diff
@@ -31,10 +30,47 @@ class CancelSponsorship(BaseSponsorshipMapping):
...
+        self.CONNECT_MAPPING['HoldID'] = 'hold_id'
+        super(CancelSponsorship, self).__init__(env)
+
     def _process_connect_data(self, connect_data):
         # Set end date to correct format for Connect
-        if 'HoldExpirationDate' in connect_data:
-            endDateStr = connect_data.get('HoldExpirationDate')
-            endDate = datetime.strptime(endDateStr, "%Y-%m-%d %H:%M:%S")
-            connect_data['HoldExpirationDate'] = endDate.strftime(
+        if 'HoldExpirationDate' in connect_data and connect_data[
```
ecino: You can simply do `if connect_data.get('HoldExpirationDate')`

ecino: But this will always be true no ?

contract_compassion/wizard/end_contract_wizard.py
MickSandoz: Yes but it's in the case if user change date and after uncheck has_new_hold

sponsorship_compassion/mappings/cancel_sponsorship_mapping.py
MickSandoz: Yes expiration_date is mandatory so i can to remove this check

contract_compassion/wizard/end_contract_wizard.py
ecino: Good point :+1:

ecino:
:+1:

PR #260, opened 2016-07-06
CO-228 ADD MajorRevision incoming message

ecino:
- Change gmc_action to allow any method call when incoming message is received


MickSandoz:
:+1:

PR #265, opened 2016-07-08
Add jobs for print or send e-mail

ecino:


partner_communication/models/communication_job.py
```diff
@@ -0,0 +1,177 @@
...
+    @api.multi
+    def _compute_html(self):
+        for job in self:
+            if job.object_id and job.email_template_id:
+                job.body_html = self.env[
+                    'mail.compose.message'].get_generated_html(
+                    job.email_template_id, [job.object_id])
+
+    @api.multi
+    def _inverse_generation(self):
```
MickSandoz: Is really usefull to have a method which returns True?

MickSandoz:
Finish to review this part, just one question asked before

partner_communication/models/communication_job.py
ecino: It is stupid but it's the way to allow writing in computed fields: we must define an `inverse` method which could do something when we update the field. In our case we have nothing to do when writing in these fields so we can just return true.

But if I remove the methode, the fields become readonly.

PR #269, opened 2016-07-12
CO-239 Add wizard for hold creation

philippe89:


ecino:
You have a PEP8 error which makes the build fail :
./child_compassion/wizards/__init__.py:16:38: W292 no newline at end of file

Before a pull request you can run flake8 command to check your code against PEP8 standards

child_compassion/views/global_childpool_view.xml
```diff
@@ -130,6 +128,31 @@
...
+                            <field name="type"/>
+                            <field name="hold_expiration_date"/>
+                            <field name="primary_owner"/>
+                            <field name="secondary_owner"/>
+                            <field name="no_money_yield_rate"/>
+                            <field name="yield_rate"/>
+                        </group>
+                        <footer>
+                            <button name="send" string="Validate" type="object" class="oe_highlight"/>
+                            or
```
ecino: New guidelines for odoo views is to avoid cancel button, because the user can simply close the popup and that acts as a cancel. There is no need to add a button for that.

child_compassion/wizards/child_search_attributes.py
```diff
@@ -0,0 +1,124 @@
...
+#
+#    The licence is in the file __openerp__.py
+#
+##############################################################################
+from openerp import models, api, fields, _
+
+
+class ImportReview(models.TransientModel):
+
+    _name = 'child.search.attributes'
```
ecino: This name is not very good... what about `child.hold.wizard`? Because these are not search attributes, but rather hold parameters. If you find another name you can as well propose it. Rename as well the references to the object in the views and the code.

child_compassion/wizards/child_search_attributes.py
```diff
@@ -0,0 +1,124 @@
...
+#    Releasing children from poverty in Jesus' name
+#    @author: Emanuel Cino
+#
+#    The licence is in the file __openerp__.py
+#
+##############################################################################
+from openerp import models, api, fields, _
+
+
+class ImportReview(models.TransientModel):
```
ecino: The classe name should be similar to the object name

child_compassion/wizards/child_search_attributes.py
```diff
@@ -0,0 +1,124 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (C) 2015 Compassion CH (http://www.compassion.ch)
+#    Releasing children from poverty in Jesus' name
+#    @author: Emanuel Cino
```
ecino: You can put your name instead

child_compassion/wizards/child_search_attributes.py
```diff
@@ -0,0 +1,124 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (C) 2015 Compassion CH (http://www.compassion.ch)
```
ecino: 2016

child_compassion/wizards/child_search_attributes.py
```diff
@@ -0,0 +1,124 @@
...
+    ##########################################################################
+    #                             FIELDS METHODS                             #
+    ##########################################################################
+
+    @api.multi
+    def send(self):
+
+        holds = self.env['compassion.hold']
+        messages = self.env['gmc.message.pool']
+        global_child = self.env['compassion.childpool.search'].browse(
```
ecino: Rename this variable to `child_search` or `global_child_search`

child_compassion/wizards/child_search_attributes.py
```diff
@@ -0,0 +1,124 @@
...
+            self.env.context.get('active_id'))
+
+        for child in global_child.global_child_ids:
+            # Save children form global children to compassion children
+
+            child_vals = {
+                'global_id': child.global_id,
```
ecino: instead of copying all values, you could use method `read` of an object to get that dictionary. See https://www.odoo.com/documentation/8.0/reference/orm.html#openerp.models.Model.read

child_compassion/wizards/child_search_attributes.py
```diff
@@ -0,0 +1,124 @@
...
+            hold_vals = {
+                'name': self.name,
+                'child_id': child_comp.id,
+                'type': self.type,
+                'expiration_date': self.hold_expiration_date,
+                'primary_owner': self.primary_owner,
+                'secondary_owner': self.secondary_owner,
+                'no_money_yield_rate': self.no_money_yield_rate,
+                'yield_rate': self.yield_rate,
+                'channel': '',
```
ecino: Add as well channel and source code as fields in the wizard, like you did for the others. Don't forget to add them in the view.

child_compassion/wizards/child_search_attributes.py
```diff
@@ -0,0 +1,124 @@
...
+                    'state': 'N',
+                }
+                child_to_update.write(child_vals)
+            else:
+                # delete child if no hold_id received
+                child_to_update.unlink()
+                hold.unlink()
+
+        return {
+            'name': _(''),
```
ecino: You can put a name, like `Created holds`

child_compassion/wizards/child_search_attributes.py
```diff
@@ -0,0 +1,124 @@
...
+            'type': 'ir.actions.act_window',
+            'view_type': 'form',
+            'view_mode': 'form,tree',
```
ecino: if you put tree at first I think it displays the tree mode at first

child_compassion/wizards/child_search_attributes.py
```diff
@@ -0,0 +1,124 @@
...
+            'res_model': 'compassion.hold',
+            'target': 'current',
```
ecino: If you add
```
'domain': [('id', 'in', holds.ids)]
````
It will only display the freshly created holds rather than all existing holds.

ecino:
:+1:

PR #271, opened 2016-07-15
Hotfixes

ecino:
- Clean invoices FIX for suspension/reactivation and contract termination
- #BOX# problems in child and sponsor letters translations
- Put analytic account of users into partner origins
- Put analytic account of origins into generated invoices


ecino:
Some travis errors because of Legacy Web Services not responding... will merge anyway.

PR #274, opened 2016-07-18
C0-138 Add Compassion Logging Module

maxime-beck:
I started taking too much time on trying to call the subscribe function of auditlog through my logging_compassion module.
After testing, I noticed that it actually works fine without calling it (action_id fields stay at null).
Let me know if you judge it necessary to invest more time on this issue.

### DIFF
--- a/logging_compassion/__openerp__.py
+++ b/logging_compassion/__openerp__.py
@@ -1,0 +1,43 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#       ______ Releasing children from poverty      _
+#      / ____/___  ____ ___  ____  ____ ___________(_)___  ____
+#     / /   / __ \/ __ `__ \/ __ \/ __ `/ ___/ ___/ / __ \/ __ \
+#    / /___/ /_/ / / / / / / /_/ / /_/ (__  |__  ) / /_/ / / / /
+#    \____/\____/_/ /_/ /_/ .___/\__,_/____/____/_/\____/_/ /_/
+#                        /_/
+#                            in Jesus' name
+#
+#    Copyright (C) 2016 Compassion CH (http://www.compassion.ch)
+#    @author: Maxime Beck <mbcompte@gmail.com>
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+
+{
+    'name': "Compassion Logging",
+    'version': '8.0.3.0',
+    'category': 'Audit and Logging',
+    'author': 'Compassion CH',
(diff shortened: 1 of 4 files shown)


maxime-beck:
./logging_compassion/__openerp__.py has wrong header ?

ecino:
__openerp__.py should follow the structure given in file https://github.com/CompassionCH/compassion-modules/blob/master/travis/compassion_openerp_header.txt

logging_compassion/__openerp__.py
```diff
@@ -0,0 +1,43 @@
...
+##############################################################################
+
+
+{
+    'name': "Compassion Logging",
+    'version': '8.0.3.0',
+    'category': 'Audit and Logging',
+    'author': 'Compassion CH',
+    'website': 'http://www.compassion.ch',
+    'depends': ['base'],
```
ecino: Should depend on module `auditlog`

logging_compassion/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_logging_compassion_logging_compassion,logging_compassion.logging_compassion,model_logging_compassion_logging_compassion,,1,0,0,0
```
ecino: This security file is useless because you didn't create any new models.

logging_compassion/__openerp__.py
ecino: Should also depend on module `sbc_compassion`

logging_compassion/data/compassion_auditlog_rules.xml
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data>
+        <record id="correspondence_rule" model="auditlog.rule">
+            <field name="name">Correspondence Rule</field>
+            <field name="model_id" ref="sbc_compassion.model_correspondence" />
+            <field name="log_type">full</field>
+            <field name="log_read">False</field>
+            <field name="log_write">Create</field>
```
ecino: Shouldn't be `True` instead of `Create` ?

logging_compassion/data/compassion_auditlog_rules.xml
```diff
@@ -0,0 +1,65 @@
...
+            <field name="log_unlink">Create</field>
+            <field name="log_create">Create</field>
+            <field name="state">subscribed</field>
+        </record>
+        <record id="res_partner_rule" model="auditlog.rule">
+            <field name="name">Res Partner Rule</field>
+            <field name="model_id" ref="crm_compassion.model_res_partner" />
```
ecino: ref could be `base.model_res_partner`

logging_compassion/data/compassion_auditlog_rules.xml
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data>
+        <record id="correspondence_rule" model="auditlog.rule">
+            <field name="name">Correspondence Rule</field>
+            <field name="model_id" ref="sbc_compassion.model_correspondence" />
+            <field name="log_type">full</field>
+            <field name="log_read">False</field>
```
ecino: You should either set to `false` without capital F or `0`

ecino:
Calling subscribe is useful to create the `action` so that the logs can be shown directly from a record. If you spent too much time on it I will take care of adding it.

ecino:
You can also remove `auditlog` dependency on module `child_compassion` (because it is not a real dependency).

maxime-beck:
I didn't try to implement the call to the subscribe function on this fix so we can merge the module and still show it at the sprint review.

ecino:
Yes, it's good.

ecino:
:+1:

PR #275, opened 2016-07-18
Fix for  CO-286, CO-287, CO-288, CO-290, CO-289

maxime-beck:


child_compassion/models/household.py
```diff
@@ -30,6 +30,9 @@ class Household(models.Model):
         readonly=True
     )
 
+    _sql_constraints = [('household_uniq', 'unique(household_id)',
+                        ('An Household with the same ID already exists.'))]
+
```
maxime-beck: I added the unique constraint but couldn't reproduce the bug (new Household is created with the same ID when updating an existing one).
The update doesn't seem to be the issue.

ecino: This is in case of a major revision but take a look at what happens when calling the "Child Details" message, because this might create Households as well.

child_compassion/models/compassion_hold.py
```diff
@@ -70,6 +70,13 @@ def check_hold_validity(self):
             ('expiration_date', '<',
              fields.Datetime.now())
         ])
+
+        for expired_hold in expired_holds:
```
ecino: Why looping twice in the same lists when you can do all operations in the same loop?

child_compassion/models/compassion_hold.py
```diff
@@ -70,8 +70,14 @@ def check_hold_validity(self):
             ('expiration_date', '<',
              fields.Datetime.now())
         ])
-        for hold in expired_holds:
-            hold.active = False
+
+        for expired_hold in expired_holds:
+            child = self.env['compassion.child'].search([
```
ecino: not useful to search, simply `child = expired_hold.child_id` or directly `expired_hold.child_id.active = False`

ecino:
:+1:

PR #287, opened 2016-07-28
Disaster

ecino:
- Log notes when new disaster impact on child, and when new notes.
- Add buttons to open impacted children and icps from disaster alert view
- Change a bit the child view
- Subscribe to auditlog rules

### DIFF
--- a/child_compassion/models/child_note.py
+++ b/child_compassion/models/child_note.py
@@ -31,6 +31,14 @@ class ChildNote(models.Model):
     source_code = fields.Char()
 
     @api.model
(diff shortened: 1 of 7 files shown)


maxime-beck:
I think there may be an indent issue on [1]child_compassion_view.xml
___
[1] https://github.com/ecino/compassion-modules/blob/fafba9859cda1c611ee03116cc7a061d0b9bf442/child_compassion/views/child_compassion_view.xml#L4

maxime-beck:
:+1:

PR #288, opened 2016-07-28
CO-319 & CO-189

philippe89:


child_compassion/wizards/child_hold_wizard.py
```diff
@@ -35,7 +35,9 @@ class ChildHoldWizard(models.TransientModel):
         ('Sponsored', _('Sponsored')),
         ('Sub Child Hold', _('Sub Child Hold'))], required=True)
     hold_expiration_date = fields.Datetime(required=True)
-    primary_owner = fields.Char(required=True)
+    primary_owner = fields.Char(required=True,
+                                default=lambda self:
+                                self.default_primary_owner())
```
ecino: You could directly put `lambda self: self.env.user.name` instead of doing a function for this.

sbc_compassion/models/correspondence.py
```diff
@@ -54,6 +54,10 @@ class Correspondence(models.Model):
     name = fields.Char(compute='_set_name')
     correspondant_id = fields.Many2one(
         related='sponsorship_id.correspondant_id', store=True)
+    redactor_id = fields.Many2one(
+        'res.partner', 'Partner', store=True,
```
ecino: Don't call the field Partner but `Encourager`

sbc_compassion/models/correspondence.py
```diff
@@ -54,6 +54,10 @@ class Correspondence(models.Model):
...
+        readonly=False, compute=lambda self:
```
ecino: Should not be a computed field. The user can simply select an encourager if needed.

sbc_compassion/views/correspondence_view.xml
```diff
@@ -69,6 +69,7 @@
                                 <field name="communication_type_ids" widget="many2many_checkboxes"/>
                             </group>
                             <group>
+                                <field name="redactor_id" string="Redactor name"/>
```
ecino: put the field after `relationship` and make it visible only if `relationship` is set to `Encourager`

ecino:
Waiting some clarifications from US about redactors, as we couldn't send other redactors when trying.

ecino:
So we learned that `redactor` is not available anymore. We therefore won't add this functionality. Can you remove commit CO-189 ? You can remove it with `git rebase -i upstream/master` if you remove the line where the commit is listed (force push is then required)

ecino:
:+1:

PR #289, opened 2016-07-28
CO-267 Add support for outcoming GlobalPartnerWeeklyDemandRequest

maxime-beck:


crm_compassion/data/gmc_action.xml
```diff
@@ -0,0 +1,17 @@
...
+    <data>
+        <!-- Weekly Planning Message Creation -->
+        <record id="create_demand_planning" model="gmc.action">
+            <field name="direction">out</field>
+            <field name="name">DemandPlanning</field>
+            <field name="mapping_name">create_demand_planning</field>
+            <field name="model">demand.planning</field>
+            <field name="description">This will send a new set of Weekly Plannings to GMC.</field>
+            <field name="connect_service">globalpartners/demandplanning</field>
+            <field name="connect_outgoing_wrapper"></field>
```
ecino: You can now remove this field

crm_compassion/model/demand_planning.py
```diff
@@ -73,3 +73,39 @@ class DemandPlanning(models.Model):
     weekly_demand_ids = fields.Many2many(
         'demand.weekly.demand', string='Weekly Demands'
     )
+
+    ##########################################################################
+    #                             ORM METHODS                                #
+    ##########################################################################
+    @api.model
+    def create(self, vals):
```
ecino: I wouldn't send the message directly at creation. Add instead a `send` button in the view that would trigger the send to GMC.

You can also add a `state` field to know if a planning was sent or not (by default state would be draft, when successfully sent it goes to sent, and if an error occurs you can put state error)

maxime-beck: Should I add an `active` field as well to handle the unlink operation ?

ecino: No, I don't think it is necessary

crm_compassion/model/demand_planning.py
```diff
@@ -33,43 +34,84 @@ class WeeklyDemand(models.Model):
...
+        res = super(DemandPlanning, self).create(vals)
+        res.state = 'draft'
```
ecino: You don't need to override create method, you can simply add parameter `default='draft'` to the field definition

crm_compassion/model/demand_planning.py
```diff
@@ -33,43 +34,84 @@ class WeeklyDemand(models.Model):
...
 
     name = fields.Char(required=True)
     date = fields.Date()
     sent_date = fields.Datetime()
     state = fields.Selection([
-        ('new', 'New'),
+        ('draft', 'Draft'),
         ('sent', 'Sent'),
         ('error', 'Error'),
     ])
```
ecino: add `default='draft'` here

crm_compassion/model/demand_planning.py
```diff
@@ -33,43 +34,84 @@ class WeeklyDemand(models.Model):
...
+    def create(self, vals):
+        res = super(DemandPlanning, self).create(vals)
+        res.state = 'draft'
+        return res
+
+    @api.multi
+    def write(self, vals):
+        res = super(DemandPlanning, self).write(vals)
+        if not isinstance(res, bool):
+            res.state = 'draft'
```
ecino: This is bad, you will put state to draft each time we write on the object. Remove this method...

crm_compassion/model/demand_planning.py
```diff
@@ -33,43 +34,84 @@ class WeeklyDemand(models.Model):
...
+
+        action_id = self.env.ref(
+            'crm_compassion.create_demand_planning').id
+        object_id = self.id
+        message_vals = {
+            'action_id': action_id,
+            'object_id': object_id
+        }
+        pool = message_obj.with_context(async_mode=False,
+                                        creating=True).create(message_vals)
```
ecino: what is creating=True?

crm_compassion/model/demand_planning.py
```diff
@@ -33,43 +34,84 @@ class WeeklyDemand(models.Model):
...
+
+        if pool.failure_reason:
+            self.state = 'error'
+            self.env.cr.commit()
+            raise Warning("Error", pool.failure_reason)
+        self.state = 'sent'
```
ecino: Add `return True` at the end. Button methods should always return something.

crm_compassion/model/demand_planning.py
maxime-beck: `creating=True` is not relevant anymore since I added the button that triggers the message sending. I forgot the remove it.

crm_compassion/model/demand_planning.py
maxime-beck: I figured after editing a message in `sent` or `error` state, we should always put it back to `draft` until the user try to resend it again to GMC. Keeping another one of these states after editing would be incorrect.

ecino:
Merged with corrections :+1:

PR #292, opened 2016-07-29
ICP suspension preparation

ecino:
- New module for testing incoming messages
- ICP suspension : hold letters and reactivate letters
- Some corrections in ICP lifecycle message


philippe89:
(thumbsup)

PR #293, opened 2016-08-02
CO-308 minor changes on event parameters

philippe89:


child_compassion/models/child_compassion.py
```diff
@@ -300,7 +300,7 @@ class CompassionChild(models.Model):
 
     # Delegation
     ############
-    delegated_to = fields.Many2one('res.partner', 'Delegated to')
+    delegated_to = fields.Many2one('res.users', 'Delegated to')
```
ecino: This is not good, because it will prevent us to delegate to any sponsor... 
A user is linked to a partner object, so you can fetch it by doing `user.partner_id`.

crm_compassion/model/event_compassion.py
```diff
@@ -13,6 +13,14 @@
 from datetime import datetime
 
 
+class CompassionChild(models.Model):
+    _inherit = 'compassion.child'
+
+    @api.multi
+    def child_available(self):
+        self.state = 'D'
```
ecino: If you do like this, all new children will be delegated which is not what we want. We want only children taken for events that become delegated.

crm_compassion/model/event_compassion.py
```diff
@@ -56,6 +64,19 @@ def create_hold_vals(self, child_comp):
...
 
     @api.multi
+    def create_child_vals(self, child):
+        child_vals = super(ChildHoldWizard, self).create_child_vals(child)
+
+        if self.env.context.get('event_id') is not None:
+            child_vals.update({
+                'date_delegation': fields.date.today(),
+                'delegated_comment': self.env.context.get('event_name'),
+                'delegated_to': self.env.context.get('user_id')
```
ecino: Maybe here you can add `'state': 'D'` ?

crm_compassion/model/event_compassion.py
```diff
@@ -456,6 +477,8 @@ def allocate_children(self):
             'target': 'current',
             'context': self.with_context({
                 'number_allocate_children': self.number_allocate_children,
-                'event_id': self.id
+                'event_id': self.id,
+                'user_id': self.user_id.id,
```
ecino: you can put `partner` with `self.user_id.partner_id.id`

ecino:
Some conflicts were introduced with the previous merges

ecino:
:+1:

PR #295, opened 2016-08-04
Disaster Alert and Reservation to Hold

ecino:


child_compassion/__openerp__.py
```diff
@@ -83,6 +83,7 @@
         'data/icp.sociological.activity.csv',
         'data/icp.spiritual.activity.csv',
         'data/fo.high.risk.csv',
+        'data/fo.disaster.loss.csv',
```
philippe89: test

child_compassion/__openerp__.py
```diff
@@ -83,6 +83,7 @@
         'data/icp.sociological.activity.csv',
         'data/icp.spiritual.activity.csv',
```
philippe89: test

child_compassion/models/compassion_hold.py
```diff
@@ -128,3 +131,20 @@ def hold_sent(self, vals):
...
+
+    @api.model
+    def reservation_to_hold(self, commkit_data):
+        """ Called when a reservation gots converted to a hold. """
+        mapping = ReservationToHoldMapping(self.env)
+        hold_data = commkit_data.get(
+            'GlobalPartnerBeneficiaryReservationToHoldNotification')
+        child_global_id = hold_data and hold_data.get('Beneficiary_GlobalID')
+        if child_global_id:
+            child = self.env['compassion.child'].create(
```
philippe89: Maybe this is faster:
hold = self.env['compassion.hold'].create(
              mapping.get_vals_from_connect(hold_data))
child = self.env['compassion.child'].create({
              'global_id': child_global_id,
              'hold_id': hold
})

ecino: Yes I will correct it

ecino: In fact it is the same because at the end I have to add `hold.child_id = child` with your solution.
It just inverse the order of creation between child and hold.

PR #299, opened 2016-08-09
CO-295 CRON for automatically send demand planning

ecino:
- automatically compute next 78 weeks at
  demand planning creation
- automatically compute revisions


philippe89:
:+1:

PR #303, opened 2016-08-11
CO-325 & CO-340 & CO-342 & CO-343

philippe89:


child_compassion/mappings/project_reservation_mapping.py
```diff
@@ -21,7 +21,7 @@ class ProjetReservationCreateMapping(OnrampMapping):
     MAPPING_NAME = "create_reservation"
 
     CONNECT_MAPPING = {
-        'Beneficiary_GlobalID': None,
+        'Beneficiary_GlobalID': 'global_id',
```
ecino: You should keep `None`. global_id is not a field in model `icp.reservation`

child_compassion/models/project_reservation.py
```diff
@@ -23,6 +23,7 @@ class ProjectReservation(models.Model):
     channel_name = fields.Char()
     icp_id = fields.Many2one(
         'compassion.project', 'Project', required=True)
+    global_id = fields.Char(default='')
```
ecino: Instead of global_id, you should link to the child directly `child_id = fields.Many2one(...`

child_compassion/mappings/project_reservation_mapping.py
ecino: Sorry you can ignore my previous comment, but you can change to `'child_id.global_id'` so that you can link the reservation to the child.

contract_compassion/model/contracts.py
```diff
@@ -334,6 +334,9 @@ def contract_active(self):
         if not self.is_active:
             vals.update({'activation_date': datetime.today().strftime(DF)})
         self.write(vals)
+        self.env['compassion.hold'].browse(self.child_id.hold_id.id).active\
```
ecino: `self.child_id_hold_id.active = False` is enough

contract_compassion/wizard/end_contract_wizard.py
```diff
@@ -64,22 +64,25 @@ def end_contract(self):
...
-            # child will be inactive after this choice
-            self.hold_expiration_date = fields.Datetime.now()
-            child.write({'active': False})
-
-        # in all case current hold became inactive
-        hold.write({'active': False})
+        if self.has_new_hold:
+            self.env['icp.reservation'].create({
+                'name': 'single beneficiary reservation',
+                'expiration_date': self.end_date,
```
ecino: should be set to `self.hold_expiration_date` as well. Otherwise the reservation will expire too quickly (`end_date` is most probably today)

contract_compassion/wizard/end_contract_wizard.py
```diff
@@ -64,22 +64,25 @@ def end_contract(self):
...
-        hold = child.hold_id
 
-        if not self.has_new_hold:
-            # child will be inactive after this choice
-            self.hold_expiration_date = fields.Datetime.now()
-            child.write({'active': False})
-
-        # in all case current hold became inactive
-        hold.write({'active': False})
+        if self.has_new_hold:
```
ecino: I think you forgot the case where the sponsorship was not yet activated. In that scenario the child may still have a valid hold active. So in that case :
1. If user wants to keep the child on hold - do nothing
2. If user doesn't want to keep the hold - inactive it, remove it, and inactive the child (+ state = 'F')

contract_compassion/wizard/end_contract_wizard.py
philippe89: in the wizard you can chose both dates freely so then it becomes useless to be able to change both dates?

ecino: In any case the reservation should not expire when the sponsorship is terminated. Otherwise we will never have it on hold.

I think end_date of contract is useless. You can modify it but the contract will be terminated immediately.

ecino:
:+1:

PR #310, opened 2016-08-25
Add new mass_mailing module with Sendgrid

ecino:
Will be used to send the Christmas mailing to our sponsors

- Adapt sendgrid events to call hooks that can be inherited for doing more than just writing in e-mail tracking object
- Correct sendgrid templates to use v3 API


philippe89:
:+1:

PR #313, opened 2016-08-30
CO-386 Fix demand planning CRON

maxime-beck:


ecino:
You should as well correct method `_get_default_weekly_demands()` because it creates prevision for next week, but it should be for the week after.

maxime-beck:
Doesn't `_get_default_weekly_demands()` creates/updates the previsions for all the weeks after the current one ?
https://github.com/maxime-beck/compassion-modules/blob/master/crm_compassion/models/demand_planning.py#L54

ecino:
Yes this is good but it starts from next sunday instead of sunday in two weeks :
https://github.com/maxime-beck/compassion-modules/blob/master/crm_compassion/models/demand_planning.py#L51

ecino:
:+1:

PR #314, opened 2016-08-30
CO-392 Add child_id field on Reservation View

maxime-beck:


child_compassion/views/project_reservation_view.xml
```diff
@@ -21,6 +21,7 @@
                         <group>
                             <group>
                                 <field name="channel_name" />
+                                <field name="child_id" />
```
ecino: Put it after icp_id

ecino: You can add attributes `attrs="{'invisible': [('child_id', '=', False)]}"` to hide it if not set.

ecino:
:+1:

PR #316, opened 2016-08-30
CO-390 Fix web resupply compute

maxime-beck:


crm_compassion/models/weekly_demand.py
```diff
@@ -115,7 +115,11 @@ def _default_unsponsored_web(self):
         start_date = datetime.today() - timedelta(weeks=52)
         web_sponsored = self.env['recurring.contract'].search_count([
             ('channel', '=', 'internet'),
-            ('start_date', '>=', fields.Date.to_string(start_date))
+            ('start_date', '>=', fields.Date.to_string(start_date)),
+            '&',
```
ecino: This is not useful, '&' is the default operator between the conditions.

crm_compassion/models/weekly_demand.py
```diff
@@ -115,7 +115,11 @@ def _default_unsponsored_web(self):
...
+            ('origin_id.type', '!=', 'partner'),
```
ecino: You can have in one line `('origin_id.type', 'not in', ('partner', 'event', 'sub'))`

ecino:
:+1:

PR #319, opened 2016-09-01
CO-349

philippe89:


child_compassion/controllers/web_children_hold.py
```diff
@@ -66,11 +66,13 @@ def hold_children_job(session, research_id):
         active_id=research_id).sudo()
     expiration_date = datetime.now() + timedelta(minutes=15)
 
+    user_id = session.env['res.users'].search([('name', '=', 'Roma Reber')])
```
ecino: Add `.id` at the end, to get the id instead of ther record. Name should be `Reber Rose-Marie` to find it.

child_compassion/wizards/child_hold_wizard.py
```diff
@@ -22,8 +22,7 @@ class ChildHoldWizard(models.TransientModel):
     type = fields.Selection(
         selection='_get_hold_types', required=True, default='Consignment Hold')
     hold_expiration_date = fields.Datetime(required=True)
-    primary_owner = fields.Char(required=True,
-                                default=lambda self: self.env.user.name)
+    primary_owner = fields.Many2one('res.users', required=True)
```
ecino: Leave default method `lambda self: self.env.uid`

contract_compassion/wizard/end_contract_wizard.py
```diff
@@ -74,7 +74,7 @@ def end_contract(self):
                     'expiration_date': self.hold_expiration_date,
                     'hold_expiration_date': self.hold_expiration_date,
                     'number_of_beneficiaries': '1',
-                    'primary_owner': self.env.user.name,
+                    'primary_owner': self.env.user,
```
ecino: `self.env.uid` to get the id instead of the record.

sponsorship_compassion/mappings/cancel_sponsorship_mapping.py
```diff
@@ -53,7 +53,7 @@ def _process_connect_data(self, connect_data):
                 'child_id': child.id,
                 'type': 'Consignment Hold',
                 'expiration_date': end_date,
-                'primary_owner': 'Rose-Marie Reber',
+                'primary_owner': self.env.user,
```
ecino: `self.env.uid`

ecino:
:+1:

PR #323, opened 2016-09-05
CO-417 Add Hold Start Date field on Events

maxime-beck:


crm_compassion/models/event_compassion.py
```diff
@@ -138,7 +139,8 @@ class event_compassion(models.Model):
         track_visibility='onchange',
         required=True)
     planned_sponsorships = fields.Integer(
-        'Expected sponsorships', track_visibility='onchange', required=True)
+        'Expected sponsorships', compute='_compute_planned_sponsorships',
```
ecino: No, this field must not be computed. It's the user who must estimate how many sponsorships we will win from this event.

The computation you made must be for the demand planning when we compute the demand for the events.

crm_compassion/models/event_compassion.py
```diff
@@ -151,6 +153,12 @@ class event_compassion(models.Model):
     parent_copy = fields.Many2one(
         'account.analytic.account', related='parent_id')
     calendar_event_id = fields.Many2one('calendar.event')
+    hold_start_date = fields.Date(
+        'Hold Start Date',
+        required=True,
+        default=lambda self: int(self.env['ir.config_parameter'].get_param(
```
ecino: This is wrong because you assign an integer to a date field. Instead of using this default attribute I propose you to add a `@api.onchange('start_date')` method, so that the field is set when the start_date is selected.

see https://www.odoo.com/documentation/8.0/reference/orm.html#module-openerp.api

crm_compassion/models/event_compassion.py
```diff
@@ -261,6 +269,11 @@ def write(self, vals):
         """ Push values to linked objects. """
         super(event_compassion, self).write(vals)
 
+        if self.hold_start_date > self.start_date:
```
ecino: Instead of checking in the write method, use a `@api.constrains` method.
See https://www.odoo.com/documentation/8.0/reference/orm.html#module-openerp.api

crm_compassion/models/weekly_demand.py
```diff
@@ -80,19 +80,22 @@ def _compute_period_locked(self):
...
+                ('hold_start_date', '>=', week.week_start_date),
+                ('start_date', '<=', week.week_end_date)
             ])
-            allocate = sum(events.mapped('number_allocate_children'))
-            resupply = allocate - sum(events.mapped('planned_sponsorships'))
+            allocate = 0
+            weeks_before_event = 0
+            for event in events:
+                weeks_before_event = (
+                     fields.Date.from_string(event.hold_start_date) -
```
ecino: It should be `start_date - hold_start_date` instead of `hold_start_date - start_date`

crm_compassion/models/weekly_demand.py
```diff
@@ -80,19 +80,22 @@ def _compute_period_locked(self):
...
+                     fields.Date.from_string(event.start_date)
+                 ).days / 7
+                allocate += event.number_allocate_children
+            resupply = allocate
```
ecino: This is wrong. You should compute allocate and resupply for each event.
```
for event in event:
    event_allocate = event.number_allocate_children / weeks_before_event (if >1)
    allocate += event_allocate
```
Resupply is computed in another way. I will explain it another comment.

ecino:
For event resupply :

1. Add a setting for defining how many days we want to keep the children on hold after the event. Use 10 days by default.
https://github.com/CompassionCH/compassion-modules/blob/master/crm_compassion/wizards/demand_settings.py#L22
2. When selecting children from the event, put the hold expiration_date by default to the end_date + the setting : 
https://github.com/CompassionCH/compassion-modules/blob/master/crm_compassion/models/event_compassion.py#L472
3. For the event resupply computation : do it separately from the allocate computation. You must search events that have week_start_date >= (end_date + setting) <= week_end_date. We resupply all the children at once (= event_planned_sponsorships - event_children_allocate)

crm_compassion/models/event_compassion.py
```diff
@@ -151,10 +152,25 @@ class event_compassion(models.Model):
...
     #                             FIELDS METHODS                             #
     ##########################################################################
+    @api.model
+    def _default_hold_start_date(self):
+        days_allocate_before_event = int(
+            self.env['ir.config_parameter'].get_param(
+                'crm_compassion.days_allocate_before_event'
+            )
+        )
+        self.hold_start_date = fields.date.today() - timedelta(
```
ecino: `fields.date.today()` returns a string. So you should use `datetime.today()` instead to have a date object.
Moreover you should not assign the result to `self` in a `api.model` method. Return the result instead. It should also be converted to a string before returned (using `fields.Date.to_string()`)

crm_compassion/models/event_compassion.py
```diff
@@ -450,11 +473,18 @@ def _get_calendar_vals(self):
 
     @api.multi
     def allocate_children(self):
-        no_money_yield = float(
-            self.planned_sponsorships) / self.number_allocate_children
+        no_money_yield = float(self.planned_sponsorships)
```
ecino: Previous version was correct, except that we need to multiply by 100 (number should be 0-100)

ecino: Sorry I didn't see that you divide later. You can ignore this.

crm_compassion/models/event_compassion.py
```diff
@@ -450,11 +473,18 @@ def _get_calendar_vals(self):
...
     def allocate_children(self):
-        no_money_yield = float(
-            self.planned_sponsorships) / self.number_allocate_children
+        no_money_yield = float(self.planned_sponsorships)
         yield_rate = float(
             self.number_allocate_children - self.planned_sponsorships
-        ) / float(self.number_allocate_children)
+        )
+        if self.number_allocate_children > 1:
+            no_money_yield /= self.number_allocate_children
```
ecino: Add `* 100` because we need percentages between 0-100

crm_compassion/models/event_compassion.py
```diff
@@ -450,11 +473,18 @@ def _get_calendar_vals(self):
...
+            yield_rate /= float(self.number_allocate_children)
```
ecino: Add `* 100`

crm_compassion/models/weekly_demand.py
```diff
@@ -94,24 +94,32 @@ def _compute_period_locked(self):
...
         for week in self.filtered('week_start_date').filtered('week_end_date'):
-            start_date = fields.Date.from_string(week.week_start_date) + \
-                timedelta(days=days_before_event)
-            end_date = fields.Date.from_string(week.week_end_date) + \
-                timedelta(days=days_before_event)
             events = self.env['crm.event.compassion'].search([
-                ('start_date', '>=', fields.Date.to_string(start_date)),
-                ('start_date', '<=', fields.Date.to_string(end_date)),
+                ('hold_start_date', '>=', week.week_start_date),
+                ('start_date', '<=', week.week_end_date)
```
ecino: Should be `hold_start_date` instead of `start_date`

crm_compassion/models/weekly_demand.py
```diff
@@ -94,24 +94,32 @@ def _compute_period_locked(self):
...
+                ).days / 7
+                event_allocate = event.number_allocate_children
+                if weeks_before_event > 0:
+                    event_allocate = event.number_allocate_children / \
+                                 weeks_before_event
+                allocate += event_allocate
+            setting = int(self.env['ir.config_parameter'].get_param(
+                'crm_compassion.days_for_hold'))
+            events = self.env['crm.event.compassion'].search([
+                ('week_start_date', '>=', ('end_date' + setting)),
```
ecino: events have no `week_start_date` and `week_end_date` fields. I think you reversed the tuples in the criterias.

ecino:
Add as well `days_for_hold` in the settings view

maxime-beck:
I think I already added

ecino:
This is the model, not the view... the view is in the xml file

ecino:
Included in #332

PR #330, opened 2016-09-08
CO-434

philippe89:


child_compassion/models/project_reservation.py
```diff
@@ -88,6 +88,21 @@ def cancel_reservation(self):
...
 
+    @api.multi
+    def show_reserved_children(self):
+
+        self.ensure_one()
+        return {
+            'type': 'ir.actions.act_window',
+            'view_type': 'form',
+            'view_mode': 'form',
+            'res_model': 'project.reservation.wizard',
```
ecino: You don't need to create a new wizard for that. You can search the children directly inside this method and change this return with these values :
```
'res_model' : 'compassion.child',
'domain': [('id', 'in', found_children.ids)],
'view_mode': ['tree', 'form'],
```

child_compassion/views/project_reservation_view.xml
```diff
@@ -33,6 +33,7 @@
                                 <field name="hold_yield_rate" />
                                 <field name="is_reservation_auto_approved" />
                                 <field name="secondary_owner" />
+                                <button name="show_reserved_children" type="object" string="Show reserved children" />
```
ecino: It's better to have a nice icon button at the upper right corner of the view. See how it's done for disaster alerts : 

https://github.com/CompassionCH/compassion-modules/blob/master/child_compassion/views/field_office_disaster_alert_view.xml#L20
http://fontawesome.io/icons/

child_compassion/models/project_reservation.py
```diff
@@ -88,6 +88,27 @@ def cancel_reservation(self):
...
         }
 
+    @api.multi
+    def show_reserved_children(self):
+        self.ensure_one()
+
+        holds = self.env['compassion.hold'].search([
+            ('reservation_id', '=', self.id)
+        ])
+        children_ids = list()
```
ecino: You can directly do : `child_ids = holds.mapped('child_id.id')`

ecino:
:+1:

PR #348, opened 2016-10-11
CO - 473, 474, 478, 479, 480

emgirardin:
CO-478 & CO-479: Bug solved for gift message mapping of "Id" and "ID" & add mapping of FieldOfficeNoteFieldForGlobalPartner
CO-480 Added Partner and Child in message tree view
CO-474: The wizard for downloading child pictures display information about missing/invalid image url
CO-473: "update picture" is working now.

### DIFF
--- a/child_compassion/models/child_pictures.py
+++ b/child_compassion/models/child_pictures.py
@@ -27,7 +27,7 @@ class child_pictures(models.Model):
     """
 
     _name = 'compassion.child.pictures'
-    _order = 'date desc'
+    _order = 'date desc, id desc'
 
     ##########################################################################
     #                                 FIELDS                                 #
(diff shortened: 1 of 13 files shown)


child_compassion/models/child_pictures.py
```diff
@@ -43,21 +43,27 @@ class child_pictures(models.Model):
...
-            picturename = data.datas_fname
-            picturename = picturename.split('.')
-            picturename = picturename[0]
-            if picturename == 'Fullshot':
-                self.fullshot = data.datas
-            elif picturename == 'Headshot':
-                self.headshot = data.datas
+        # We sort the images by date of creation from newest to oldest
+        attachments = attachments.sorted(
+            lambda key: key.create_date, reverse=True)
```
ecino: you can directly add the desired order in the search function at line 51, by adding parameter `order='create_date desc'`

message_center_compassion/models/gmc_message.py
```diff
@@ -70,6 +70,7 @@ class GmcMessagePool(models.Model):
                                 'several records. (ids separated by commas)')
     res_name = fields.Char(compute='_compute_res_name', store=True)
     partner_id = fields.Many2one('res.partner', 'Partner')
+    child_id = fields.Many2one('compassion.child', 'Child_Code')
```
ecino: This is not needed. It is already added in module `child_compassion`

message_center_compassion/view/gmc_message_view.xml
```diff
@@ -29,6 +29,8 @@
             <field name="arch" type="xml">
                 <tree string="Message Pool" colors="blue:state == 'new';black:state == 'pending';gray:state == 'success';red:state == 'failure'" create="false">
                     <field name="action_id" />
+                    <field name="partner_id" />
+                    <field name="child_id" />
```
ecino: Remove this field from the view, because module `message_center` does not depend on `child_compassion`. You should add it only in module `child_compassion`

PR #369, opened 2016-11-10
CO-551 Intervention CreateHold and CancelHold messages

ecino:
- Intervention Hold messages
- Make intervention readonly fields
- Some access rights fixed
- Error from message center fixed


intervention_compassion/mappings/hold_mapping.py
```diff
@@ -0,0 +1,81 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (C) 2016 Compassion CH (http://www.compassion.ch)
+#    Releasing children from poverty in Jesus' name
+#    @author: Emmanuel Girardin <emmanuel.girardin@outlook.com>
```
emgirardin: This is not from me

ecino: Haha ok, I will change it.

PR #371, opened 2016-11-10
3 bug fixes about messages with gmc

emgirardin:


sbc_compassion/mappings/correspondence_mapping.py
```diff
@@ -21,6 +21,7 @@ class CorrespondenceMapping(OnrampMapping):
     Communication_Kit_Fields_and_Sample_JSON
     """
     ODOO_MODEL = 'correspondence'
+    MAPPING_NAME = 'CreateLetter'
```
ecino: Instead of doing this, you should remove the mapping_name in the file gmc_action.xml

PR #379, opened 2016-11-21
Zbar, pdfminer and Cie

emgirardin:


ecino:
Code seems ok. Waiting tests on devel

ecino:
👍

PR #435, opened 2017-03-08
Various fixes and improvements

ecino:


ecino:
To fix number of sponsorships

PR #457, opened 2017-07-17
CO-907 Reduce console output when installing opencv

maxime-beck:
- Redirect stdout to opencv_log and print stderr
- To verify if opencv is installed : pkg-config --modversion opencv


ecino (changes requested):
Please exclude folder ./opencv-3.2.0/ from `travis_header.sh` check as it is producing errors in the tests.

ecino (approved):

PR #458, opened 2017-07-24
Add Compassion Dashboard Module

maxime-beck:


partner_communication/static/src/less/communication_dashboard.less
```diff
@@ -0,0 +1,223 @@
+.o_kanban_view.o_kanban_dashboard.o_partner_communication_kanban {
```
ecino: Is this .css file useful? Isn't it possible to reuse one from the core modules?

partner_communication/views/partner_communication.xml
```diff
@@ -0,0 +1,18 @@
...
+    <data>
+        <template id="assets_backend" name="partner_communication assets" inherit_id="web.assets_backend">
+            <xpath expr="." position="inside">
+                <link rel="stylesheet" href="/partner_communication/static/src/less/communication_dashboard.less"/>
+                <script type="text/javascript" src="/partner_communication/static/src/js/phone_button_widget.js"></script>
+                <script type="text/javascript" src="/partner_communication/static/src/js/communication_dashboard.js"></script>
+            </xpath>
+        </template>
+
+        <record id="action_communication_tree" model="ir.actions.act_window">
```
ecino: This should be in `communication_job_view.xml`

compassion_dashboard/README.rst
```diff
@@ -0,0 +1,27 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
ecino, line 1: I'm not sure, is the module still needed?

ecino (changes requested):

partner_communication/static/src/less/communication_dashboard.less
maxime-beck: The css related to the ribbon is a copy of the one specific to the sales_team dashboard and the one related to the cards is defined in the core module "web" but overrited so the margins are similar to the account_journal dashboard.

Because these stylesheets are not defined as is in core modules this less file is useful. Althought, for future dashboards, we could maybe move this file in the newly created "Compassion Dashboard Modules" and use more generic names for classes ?

compassion_dashboard/README.rst
maxime-beck: I took this header from other modules, it just adds this image to the module description, do you think I should delete it ?

partner_communication/static/src/less/communication_dashboard.less
```diff
@@ -0,0 +1,223 @@
...
+    @communication-table-spacing: 10px;
+
+    .o_kanban_record {
```
ecino: This class doesn't need to be redefined, no? I assume you can use it from the core modules without to have to define it again, unless you did some changes?

partner_communication/static/src/less/communication_dashboard.less
```diff
@@ -0,0 +1,223 @@
...
+                }
+            }
+        }
+
+        .o_invisible .o_kanban_graph_section {
+            display: none; // This fix a bug on firefox
+        }
+    }
+
+    .o_communication_dashboard {
```
ecino: Can't you use another existing class and add your custom class which will just override what you need to change?
Then in your template your dashboard will have two classes : `o_existing_dashboard` (replace existing by some real class), and `my_custom_communication_class`.

compassion_dashboard/README.rst
ecino: No I mean is the whole module `compassion_dashboard` necessary? The module `partner_communication` is not even depending on it.

maxime-beck: Currently, this module only provides a widget that eases the creation of line and bar graph within dashboard cards. Therefore, partner_communication does not depend on it yet but we could use it to generalize other dashboard related elements as we go. What do you think?

ecino: Ok so we can keep it.

compassion_dashboard/README.rst
```diff
@@ -0,0 +1,27 @@
...
+    :alt: License: AGPL-3
+
+Partner Communication
```
ecino: Please change and put a useful module title instead

compassion_dashboard/__init__.py
```diff
@@ -0,0 +1,10 @@
+# -*- encoding: utf-8 -*-
```
ecino: The new standard is `coding` instead of `encoding`. Please change all your headers.

compassion_dashboard/static/src/js/compassion_dashboard.js
```diff
@@ -0,0 +1,100 @@
...
+
+var CompassionDashboardGraph = kanban_widgets.AbstractField.extend({
+    start: function() {
+        this.graph_type = this.$node.attr('graph_type');
+        this.data = JSON.parse(this.field.raw_value);
+        this.display_graph();
+        return this._super();
+    },
+
+    display_graph : function() {
```
ecino, line 15: Please comment from where you copied the code.

partner_communication/models/communication_dashboard.py
```diff
@@ -0,0 +1,76 @@
...
+
+class CommunicationDashboard(models.Model):
+
+    _inherit = 'partner.communication.config'
+    _description = 'Communication Dashboard'
+
+    ##########################################################################
+    #                                 FIELDS                                 #
+    ##########################################################################
+    @api.one
```
ecino: api.one is deprecated. Use multi instead with a loop
```
for config in self:
   config.kanban_dashboard = ...
```

partner_communication/models/communication_dashboard.py
```diff
@@ -0,0 +1,76 @@
...
+    def _kanban_dashboard(self):
+        self.kanban_dashboard = json.dumps(self.get_communication_dashboard_datas())
+
+    @api.one
```
ecino: Change to api.multi

ecino (changes requested):

ecino (approved):

PR #461, opened 2017-08-03
CO-1016 | CO-1026 | CO-1027 Gift_compassion and Partner_communication update

nfluckiger:
CO-1016 Update gift_compassion / Add some field to the list view, make correction in french translate, add a field to the search

CO-1026 | CO-1027 Update partner_communication / Make the send now button become print now in both main view and pdf preview

### DIFF
--- a/gift_compassion/models/sponsorship_gift.py
+++ b/gift_compassion/models/sponsorship_gift.py
@@ -76,12 +76,14 @@ class SponsorshipGift(models.Model):
     currency_id = fields.Many2one('res.currency', default=lambda s:
                                   s.env.user.company_id.currency_id)
     currency_usd = fields.Many2one('res.currency', compute='_compute_usd')
-    exchange_rate = fields.Float(readonly=True, copy=False)
+    exchange_rate = fields.Float(readonly=True, copy=False, digits=(12, 6))
     amount_us_dollars = fields.Float('Amount due', readonly=True, copy=False)
     instructions = fields.Char()
-    gift_type = fields.Selection('get_gift_type_selection', required=True)
+    gift_type = fields.Selection('get_gift_type_selection', required=True,
+                                 string=_("Gift for"))
     attribution = fields.Selection('get_gift_attributions', required=True)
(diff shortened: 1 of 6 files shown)


gift_compassion/models/sponsorship_gift.py
```diff
@@ -109,9 +111,9 @@ class SponsorshipGift(models.Model):
     @api.model
     def get_gift_type_selection(self):
         return [
-            ('Project Gift', _('Project Gift')),
-            ('Family Gift', _('Family Gift')),
-            ('Beneficiary Gift', _('Beneficiary Gift')),
+            ('Project', _('Project')),
```
ecino: Only change the right part of the tuples. The left part is the value in the database and is used as well for the mapping when receiving messages, so this will break the code if we modify it.

gift_compassion/views/gift_view.xml
```diff
@@ -14,13 +14,14 @@
                                          green:state == 'verify';
                                          red:state == 'error';
                                          grey:state == 'fund_delivered'">
+                <field name="id"/>
```
ecino: This field is not needed. It's the id in the database but we don't care of it.

partner_communication/wizards/pdf_wizard.py
```diff
@@ -31,9 +31,15 @@ class PdfPreviewWizard(models.TransientModel):
         'partner.communication.job', required=True)
     preview = fields.Binary(compute='_compute_pdf')
     state = fields.Char(compute='_compute_state')
+    send_state = fields.Char(compute='_compute_send_state')
```
ecino: Instead of a compute field, you can use a related field (you can change as well field state. This will simply become 
```
send_state = fields.Char(related='communication_id.state')
```
and you can remove the compute method.

ecino (changes requested):

ecino:
It's still missing ability to search on gmc_gift_id and float precision to 6 decimals.

ecino (approved):

PR #463, opened 2017-08-07
CO-1023 Bug correction with the composition process message

nfluckiger:
Add a lock to prevent two workers to work on the same items at the same time

### DIFF
--- a/sbc_compassion/models/correspondence.py
+++ b/sbc_compassion/models/correspondence.py
@@ -12,6 +12,7 @@ import base64
 import logging
 import re
 import uuid
+import threading
 
 from odoo import fields, models, api, _
 from odoo.exceptions import UserError
@@ -158,6 +159,9 @@ class Correspondence(models.Model):
          'unique(kit_identifier)',
          _('The kit id already exists in database.'))
     ]
+    # Lock
+    #######
+    process_lock = threading.Lock()
 
     ##########################################################################
     #                             FIELDS METHODS                             #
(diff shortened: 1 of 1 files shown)


sbc_compassion/models/correspondence.py
```diff
@@ -614,33 +618,38 @@ def get_chars(t):
...
+                published_state = 'Published to Global Partner'
+                is_published = vals.get('state') == published_state
+
+                # Write/update letter
+                kit_identifier = vals.get('kit_identifier')
+                letter = self.search([('kit_identifier', '=', kit_identifier)])
+                if letter:
+                    # Avoid to publish twice a same letter
+                    is_published = is_published and \
+                                   letter.state == published_state
```
ecino: Please keep the previous definition of `is_published`, otherwise you break the code at line 645 which will not test the same case anymore. `is_published = is_published and letter.state != published_state`. 
Adapt your test at line 638 with `if is_published or letter.state != published_state:`

ecino (approved):

sbc_compassion/models/correspondence.py
```diff
@@ -614,33 +618,38 @@ def get_chars(t):
...
+                is_published = vals.get('state') == published_state
+
+                # Write/update letter
+                kit_identifier = vals.get('kit_identifier')
+                letter = self.search([('kit_identifier', '=', kit_identifier)])
+                if letter:
+                    # Avoid to publish twice a same letter
+                    is_published = is_published and \
+                                   letter.state != published_state
+                    if is_published:
```
ecino: No, change to `if is_published or letter.state != published_state:`

PR #465, opened 2017-08-08
CO-1059 Create the send method in preview wizard

nfluckiger:


partner_communication/wizards/pdf_wizard.py
```diff
@@ -58,3 +58,8 @@ def _compute_html(self):
             self.body_html = body_html
 
         self.body_html = comm.body_html
+
+    @api.multi
+    def send(self):
+        self.communication_id.send()
```
ecino: return self.communication_id.send()

PR #471, opened 2017-08-16
CO-1036 - Implement Suspended State for Sponsorship Gifts

maxime-beck:


ecino:
please fix Travis  build (see pylint errors in the build result)

child_compassion/models/project_lifecycle_event.py
```diff
@@ -112,7 +112,7 @@ def create(self, vals):
         if event.type == 'Reactivation':
             if fund_suspended:
                 project.reactivate_project()
-            if hold_gifts and not event.hold_gifts():
+            if hold_gifts and not event.hold_gifts:
```
ecino, line 115: This test is nonsense, it will always be false. You should have renamed method hold_gifts and keep the call here.

gift_compassion/models/contracts.py
```diff
@@ -77,15 +78,42 @@ def open_gifts(self):
...
     def hold_gifts(self):
-        """ Postpone open gifts. """
+        for contract in self:
+            today = datetime.date.today()
+            first_day = today.replace(day=1)
+            last_day = today.replace(day=calendar.monthrange(today.year,
+                                                             today.month)[1])
+            suspended_gifts = self.env['sponsorship.gift'].search([
+                ('child_id.project_id', '=', contract.project_id.id),
+                ('state', '=', 'Delivered'),
```
ecino: I think you should rather look for pending gifts instead of delivered.

gift_compassion/models/contracts.py
```diff
@@ -77,15 +78,42 @@ def open_gifts(self):
...
+                ('create_date', '>=', str(first_day)),
+                ('create_date', '<=', str(last_day))
+            ])
+            suspended_gifts.action_suspended()
+
+            related_move = self.env['account.move'].browse(
```
ecino: You don't need to use the browse method. You can simply use suspended_gifts.mapped('payment_I'd)

gift_compassion/models/sponsorship_gift.py
```diff
@@ -516,6 +517,18 @@ def action_verify(self):
         return True
 
     @api.multi
+    def action_in_progress(self):
+        self.write({'state': 'In Progress'})
+        self.mapped('message_id').write({'state': 'postponed'})
```
ecino: State should be pending

gift_compassion/models/sponsorship_gift.py
```diff
@@ -516,6 +517,18 @@ def action_verify(self):
...
+        return True
+
+    @api.multi
+    def action_suspended(self):
+        self.write({'state': 'suspended'})
+        self.mapped('message_id').write({'state': 'pending'})
```
ecino: State should be suspended

ecino: *postponed

ecino (changes requested):

child_compassion/models/project_lifecycle_event.py
maxime-beck: It won't always be False, `event.hold_gifts` refers to its value in the incoming message while `hold_gifts` refers to its current value. The test prevents the method from being called if the incoming message doesn't change the `hold_gift`'s current value. I believe the parentheses were a typo.

ecino: Ok seems good

gift_compassion/models/contracts.py
```diff
@@ -77,15 +79,40 @@ def open_gifts(self):
...
     def hold_gifts(self):
-        """ Postpone open gifts. """
+        for contract in self:
+            today = datetime.date.today()
+            first_day = today.replace(day=1)
+            last_day = today.replace(day=calendar.monthrange(today.year,
+                                                             today.month)[1])
+            suspended_gifts = self.env['sponsorship.gift'].search([
+                ('child_id.project_id', '=', contract.project_id.id),
+                ('state', '=', 'open'),
```
ecino: Still not sure it will do what we want... The gifts in open state are those sent but not yet received by GMC. There should not be a lot of them. You should instead search for "In Progress" gifts (or both)

gift_compassion/models/sponsorship_gift.py
```diff
@@ -516,6 +517,18 @@ def action_verify(self):
         return True
 
     @api.multi
+    def action_in_progress(self):
+        self.write({'state': 'In Progress'})
+        self.mapped('message_id').write({'state': 'pending'})
```
ecino: I don't think you should change the state of the message. It's already done elsewhere

gift_compassion/models/sponsorship_gift.py
```diff
@@ -516,6 +517,18 @@ def action_verify(self):
...
+        return True
+
+    @api.multi
+    def action_suspended(self):
+        self.write({'state': 'suspended'})
+        self.mapped('message_id').write({'state': 'postponed'})
```
ecino: I don't think you should change the state of the message. It's already done elsewhere

ecino (changes requested):

gift_compassion/models/contracts.py
```diff
@@ -77,15 +79,43 @@ def open_gifts(self):
...
 
     def reactivate_gifts(self):
-        """ Put again gifts in OK state. """
+        for contract in self:
+            suspended_gifts = self.env['sponsorship.gift'].search([
+                ('child_id.project_id', '=', contract.project_id.id),
+                ('state', '=', 'suspended')
+            ])
+            suspended_gifts.action_in_progress()
+            suspended_gifts.action_send()
```
ecino, line 116: We should remove this call

gift_compassion/models/sponsorship_gift.py
```diff
@@ -516,6 +517,16 @@ def action_verify(self):
         return True
 
     @api.multi
+    def action_in_progress(self):
+        self.write({'state': 'In Progress'})
```
ecino, line 521: We should recreate the moves at that point

ecino (changes requested):
Still some changes to do. I will fix that before merging.

gift_compassion/models/sponsorship_gift.py
davidwul: Actually we just discussed that in the sprint review. It's wrong to send the message a second time. Then an easier solution is to put the move in draft on suspension and just confirm it on reactivation.
See the jira task was modified accordingly...

ecino:
Ok great!

PR #475, opened 2017-10-02
Improve transactions and error handling

ecino:
https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#never-commit-the-transaction

Add error messages when some may be catched and not seen.


eicher31 (approved):
Looks good.

PR #482, opened 2017-10-17
CO-1147 The name for the icon of the button was wrong

cmtCousteau:


sponsorship_compassion/views/res_partner_view.xml
```diff
@@ -52,7 +52,7 @@
                                 <field name="months_paid"/>
                                 <field name="total_amount" />
                                 <field name="state" />
-                                <button name="open_contract" type="object" icon="fa-open" />
+                                <button name="open_contract" type="object" icon="fa-external-link"  />CO-1147
```
ecino: You accidentally typed the issue code in the view :laughing:

sbc_compassion/models/correspondence_template.py
```diff
@@ -135,15 +135,15 @@ def get_gmc_layouts(self):
     def _get_default_checkboxes(self):
         return [
             (0, False, {'language_id': self.env.ref(
-                'child_compassion.lang_compassion_french').id}),
+                'child_switzerland.lang_compassion_french').id}),
```
ecino: In fact, the module `sbc_compassion` has no access to module `child_switzerland`, because `child_switzerland` is a child module in the dependency list.

Therefore, you should remove this _default_checkboxes from here (and remove the call to the function in the field definition), and move it in the module `sbc_switzerland`, while you should add `child_switzerland` in the dependencies of the module `sbc_switzerland`.

ecino (changes requested):

sbc_compassion/models/correspondence_template.py
```diff
@@ -98,9 +98,6 @@ class CorrespondenceTemplate(models.Model):
     pattern_y_max = fields.Integer(
         help='Maximum Y position of the area in which to look for the '
              'pattern inside the template (given in pixels)')
-    checkbox_ids = fields.One2many(
-        'correspondence.lang.checkbox', 'template_id',
-        default=lambda self: self._get_default_checkboxes(), copy=True)
```
ecino: Sorry it was not clear, in fact you must keep the field, but remove the `default` parameter.

ecino (changes requested):

PR #484, opened 2017-10-18
CO-1167 FIX - Remove redondant workflow activity "act_cancelled"

eicher31:


ecino (approved):

ecino:
In fact there is a problem in line https://github.com/eicher31/compassion-modules/blob/5e351a1a78819ecc7c85602d58a6367c2219b4eb/sponsorship_tracking/workflow/sds_workflow.xml#L102 and below where you must change reference to `act_cancelled` by `contract_compassion.act_cancelled`

sbc_compassion/models/import_letter_line.py
```diff
@@ -25,9 +25,9 @@ class ImportLetterLine(models.Model):
     ##########################################################################
 
     sponsorship_id = fields.Many2one('recurring.contract', 'Sponsorship',
-                                     compute='_set_sponsorship_id')
+                                     compute='_compute_set_sponsorship_id')
```
ecino: compute_sponsorship would be better

sbc_compassion/models/import_letter_line.py
```diff
@@ -25,9 +25,9 @@ class ImportLetterLine(models.Model):
...
     partner_id = fields.Many2one('res.partner', 'Partner')
-    name = fields.Char(compute='_set_name')
+    name = fields.Char(compute='_compute_set_name')
```
ecino: _compute_name would be better

sbc_compassion/models/import_letters_history.py
```diff
@@ -55,7 +55,7 @@ class ImportLettersHistory(models.Model):
         track_visibility='onchange')
     import_completed = fields.Boolean()
     nber_letters = fields.Integer(
-        "Number of files", readonly=True, compute="_count_nber_letters")
+        "Number of files", readonly=True, compute="_compute_count_nber_letters")
```
ecino: _compute_nber_letters would be better

sbc_compassion/wizards/import_review.py
```diff
@@ -23,12 +23,12 @@ class ImportReview(models.TransientModel):
     ##########################################################################
     #                                 FIELDS                                 #
     ##########################################################################
-    progress = fields.Float(compute='_get_current_line', store=True)
+    progress = fields.Float(compute='_compute_get_current_line', store=True)
```
ecino: _compute_current_line

ecino (approved):
Small propositions not mandatory...

sbc_compassion/wizards/import_letters_test.py
```diff
@@ -61,7 +61,7 @@ def _analyze_attachment(self, file_, filename):
             })
             letters_line.letter_image = self.env[
                 'ir.attachment'].create(document_vals[i])
-            letters_line.check_status()
+            letters_line._compute_check_status()
```
eicher31, line 64: I'm not sure if its the right way that access to a private methode

sbc_compassion/wizards/import_review.py
eicher31: I agree

sbc_compassion/wizards/import_letters_test.py
ecino: It's not a problem from Odoo code to access private methods of Odoo objects. Private methods won't be accessible remotely with XML-RPC.

ecino:
:+1:

PR #498, opened 2017-11-17
CO-1248 - The wrong field was used to retrieve the sub

cmtCousteau:


crm_compassion/models/weekly_demand.py
```diff
@@ -158,7 +158,7 @@ def _default_unsponsored_ambassador(self):
         """
         start_date = datetime.today() - timedelta(weeks=STATS_DURATION)
         ambass_sponsored = self.env['recurring.contract'].search_count([
-            ('origin_id.type', '=', 'partner'),
+            ('parent_id', '!=', False),
```
ecino: Here the fix is wrong! it's only the type 'sub' that we want to correct, so please remove this change.

ecino (changes requested):
One change to do. When you correct your commits, please use the commit --amend feature, to avoid duplicating your commits.

It will edit your commit instead of creating a new one. The you can rebase on compassion.

ecino:
crm_compassion/models/weekly_revision.py:96:17: W503 line break before binary operator

ecino (approved):

ecino:
Please rebase. Waiting as well for the "@job" fix in compassion-switzerland before merging this one

PR #502, opened 2017-11-24
CO-891 CREATE - Test for the module sbc_compassion,

eicher31:
test specifically the import of scan.


sbc_compassion/demo/correspondence_template_postman.xml
```diff
@@ -0,0 +1,133 @@
+<odoo>
```
ecino: maybe remove postman from file name as you define several templates in this file

ecino (approved):

ecino:
Seems thas the wrong opencv is installed on travis. You can edit the travis.yml file in order to fix that (we did the same on your machine I think).

ecino:
./sbc_compassion/tests/test_sbc_compassion.py has wrong header
It's maybe only this that fails the tests

PR #507, opened 2017-12-05
CO-586 CREATE - add a link in the intervention's view for going to

eicher31:
expense, income, contract and partner.

### DIFF
--- a/intervention_compassion/__manifest__.py
+++ b/intervention_compassion/__manifest__.py
@@ -30,7 +30,7 @@
 # pylint: disable=C8101
 {
     'name': 'Compassion Interventions',
-    'version': '10.0.1.1.0',
+    'version': '10.0.1.2.0',
     'category': 'Other',
     'author': 'Compassion CH',
     'license': 'AGPL-3',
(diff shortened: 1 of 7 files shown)


intervention_compassion/__manifest__.py
```diff
@@ -30,7 +30,7 @@
 # pylint: disable=C8101
 {
     'name': 'Compassion Interventions',
-    'version': '10.0.1.1.0',
+    'version': '10.0.1.2.0',
```
eicher31, line 33: On my db the migration doesn't work because the intervention inside doesn't exist anymore in US db.

intervention_compassion/models/compassion_intervention.py
```diff
@@ -59,6 +58,9 @@ class CompassionIntervention(models.Model):
         'compassion.project', 'icp_interventions', 'intervention_id', 'icp_id',
         string='ICPs', readonly=True
     )
+    product_id = fields.Many2one('product.template',
```
ecino: I would name this field `product_template_id` since it is not a product but a product template.

ecino (approved):

PR #531, opened 2018-01-24
CO-1344 UPDATE - Add the computed field "PDF size" in the partner com…

eicher31:
…munication view

### DIFF
--- a/partner_communication/models/communication_job.py
+++ b/partner_communication/models/communication_job.py
@@ -97,8 +97,7 @@ class CommunicationJob(models.Model):
                                    readonly=True)
     body_html = fields.Html()
     pdf_page_count = fields.Integer(string='PDF size',
-                                    compute='_compute_pdf_page',
-                                    store=True,)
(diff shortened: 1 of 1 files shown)


partner_communication/models/communication_job.py
```diff
@@ -260,7 +256,12 @@ def write(self, vals):
         if vals.get('need_call'):
             vals['state'] = 'call'
 
-        return super(CommunicationJob, self).write(vals)
+        super(CommunicationJob, self).write(vals)
+
+        if vals.get('body_html') or vals.get('send_mode'):
```
ecino: I would do ` or vals.get('send_mode') == 'physical'` to only count if we change to print the letter.

ecino (changes requested):

partner_communication/models/communication_job.py
eicher31: That's true! Thanks

ecino (approved):

PR #546, opened 2018-02-07
 CO-619 - The unknown messages are now listed

cmtCousteau:
1.

### DIFF
--- a/child_compassion/models/compassion_reservation.py
+++ b/child_compassion/models/compassion_reservation.py
@@ -167,7 +167,7 @@ class CompassionReservation(models.Model):
         """ Called when the reservation has failed"""
         self.message_post(
             subject='Reservation failed',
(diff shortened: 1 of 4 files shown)


child_compassion/models/compassion_reservation.py
```diff
@@ -184,10 +184,6 @@ def send_reservation(self):
         messages = self.with_context(async_mode=False).handle_reservation()
         for i in range(0, len(messages)):
             if messages[i].state != 'success':
-                self[i].message_post(
-                    messages[i].failure_reason,
-                    _("Reservation failed")
-                )
```
ecino: You can remove all the for loop which is not useful anymore. You can simply end the function with :
```
return self.with_context(async_mode=False).handle_reservation()
```

ecino (changes requested):

ecino (approved):

PR #573, opened 2018-02-27
hr_attendance_calendar fixes

samuel-fringeli:


hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -181,7 +184,26 @@ def _compute_break_total(self):
     @api.depends('worked_hours', 'due_hours')
     def _compute_extra_hours(self):
         for att_day in self.filtered('worked_hours'):
-            att_day.extra_hours = att_day.worked_hours - att_day.due_hours
+            extra_hours = att_day.worked_hours - att_day.due_hours
+            max_extra_hours = float(self.env['ir.config_parameter'].get_param(
+                'hr_attendance_calendar.max_extra_hours'))
+
+            if extra_hours > max_extra_hours:
```
eicher31: The max_extra_hours is for the sum of all extra hours, this comparison should be done in the employee.

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -181,7 +184,26 @@ def _compute_break_total(self):
...
+                att_day.extra_hours = max_extra_hours
+            else:
+                att_day.extra_hours = extra_hours
```
eicher31: extra_hours should not be altered even if the limit had exceeded

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -181,7 +184,26 @@ def _compute_break_total(self):
...
+
+    @api.multi
+    @api.depends('extra_hours')
+    def _compute_extra_hours_lost(self):
+        for att_day in self.filtered('worked_hours'):
+            extra_hours = att_day.worked_hours - att_day.due_hours
```
eicher31: already compute in att_day.extra_hours

eicher31 (changes requested):
Do not hesitate to ask me if you don't understand some of my requests.

hr_attendance_calendar/models/hr_attendance_day.py
samuel-fringeli: No, it's the setting we can configure in attendances settings (hr_attendance_calendar.max_extra_hours)

hr_attendance_calendar/models/hr_attendance_day.py
samuel-fringeli: extra_hours is altered, but the difference goes in extra_hours_lost

hr_attendance_calendar/models/hr_attendance_day.py
samuel-fringeli: yes, but extra_hours can already have the value of max_extra_hours at this state of the execution, so it's necessary to recompute this value

ecino:
Conflicts with devel, the branch should be rebased.

ecino (approved):

PR #586, opened 2018-03-05
Hr Attendance and more

ecino:


hr_attendance_calendar/models/hr_attendance.py
```diff
@@ -1,184 +1,75 @@
...
-                    for working_hour in att_day.employee_id.contract_id.working_hours.attendance_ids:
-                        if working_hour.dayofweek == str(fields.Date.from_string(att_day.name).weekday()):
-                            due_hours += working_hour.due_hours 
-
-                    att_day.calendar_due_hours = due_hours
-
+    ##########################################################################
+    #                                 FIELDS                                 #
+    ##########################################################################
+    # New fields --------------------------------------------------------------
```
ecino: I would remove this comment. The others are already very visible.

hr_attendance_calendar/models/hr_attendance.py
```diff
@@ -1,184 +1,75 @@
...
+    ##########################################################################
+    #                               ORM METHODS                              #
+    ##########################################################################
+    @api.model
+    def create(self, vals):
+        new_record = super(HrAttendance, self).create(vals)
+        if isinstance(vals['check_in'], basestring):
+            att_check_in_date = fields.Date.from_string(vals['check_in'])
+        else:
+            att_check_in_date = vals['check_in']
```
ecino: Do you really have the case where the value is passed as a date and not a string? This should never be the case. If yes, it's maybe somwhere else in the code that you wrongly create an attendance?

hr_attendance_calendar/models/hr_attendance.py
```diff
@@ -1,184 +1,75 @@
...
+        attendance_day = self.env['hr.attendance.day'].search([
+            ('employee_id', '=', vals['employee_id']),
+            ('date', '=', att_check_in_date)
```
ecino: In the search method (think of equivalent as a SQL query), the date parameters are always passed as strings and not as dates. So you should not convert check_in in date object for that search.

hr_attendance_calendar/models/hr_attendance.py
```diff
@@ -1,184 +1,75 @@
...
+        ])
+        if attendance_day:
+            new_record.attendance_day_id = attendance_day
+            # todo: find better solution
```
ecino, line 41: I think a good solution for the computation of the breaks would be to use some `base.action.rule` records and define all the rules that would trigger recomputation there. This way we would clean the code and all checks at write/create which would be more understandable when reading the code. We have an example in our module `sponsorship_tracking`. But I propose to add a task in JIRA for next sprint and we don't do the change for now as we want to merge as quick as possible.

hr_attendance_calendar/models/hr_attendance.py
```diff
@@ -1,184 +1,75 @@
...
+    attendance_day_id = fields.Many2one(comodel_name='hr.attendance.day',
+                                        string='Attendance day',
+                                        readonly=True)
+
+    ##########################################################################
+    #                               ORM METHODS                              #
+    ##########################################################################
+    @api.model
+    def create(self, vals):
+        new_record = super(HrAttendance, self).create(vals)
```
ecino: Add a comment (docstring) to explain you create the attendance day if not existent and that you recompute the breaks of the day.

hr_attendance_calendar/models/hr_attendance.py
```diff
@@ -1,184 +1,75 @@
...
-    calendar_due_hours = fields.Float(string="Due hours (based on calendar)", compute=_get_due_hours)
-    total_break_hours = fields.Float(string="Total break", compute=_get_total_break_time)
-    total_calendar_breaks = fields.Float(string="Total break (calendar)",compute=_get_total_calendar_break_time)
-    total_bonus_malus = fields.Float(string="Bonus/Malus", compute=_get_bonus_malus)
-
+    def write(self, vals):
+        for att in self:
+            # Check if only the date of check_in has change and thus change the
+            # attendance_day
+            check_in_date = fields.Date.from_string(att.check_in)
```
ecino, line 56: I would rename this variable `old_check_in`

hr_attendance_calendar/models/hr_attendance.py
```diff
@@ -1,184 +1,75 @@
...
+
+            if 'check_in' in vals:
+                check_in = fields.Date.from_string(vals['check_in'])
```
ecino, line 59: I would rename `new_check_in`

hr_attendance_calendar/models/hr_attendance.py
```diff
@@ -1,184 +1,75 @@
...
+                if check_in_date != check_in:
+                    attendance_day = self.env['hr.attendance.day'].search([
+                        ('employee_id', '=', att.employee_id.id),
+                        ('date', '=', check_in)])
+                    vals['attendance_day_id'] = attendance_day.id
+
+            ret = super(HrAttendance, self).write(vals)
+
+            if 'check_in' in vals or 'check_out' in vals:
+                if 'no_compute_break' not in vals:
```
ecino, line 70: `no_compute_break` is not a field in `hr_attendance`. You should use the context to add those additional variables instead of puting them in the values of the record creation!

ecino: Here we could as well use `base.action.rule` instead of all this write method!

hr_attendance_calendar/models/hr_attendance_break.py
```diff
@@ -0,0 +1,75 @@
...
+class HrAttendanceBreak(models.Model):
+    _name = "hr.attendance.break"
+
+    ##########################################################################
+    #                                 FIELDS                                 #
+    ##########################################################################
+    employee_id = fields.Many2one('hr.employee', "Employee")
+    attendance_day_id = fields.Many2one('hr.attendance.day', "Attendance day",
+                                        readonly=True,
+                                        ondelete="cascade")
```
ecino: `required=True`?

hr_attendance_calendar/models/hr_attendance_break.py
```diff
@@ -0,0 +1,75 @@
...
+    duration = fields.Float('Duration', compute='_compute_duration',
+                            store=True, readonly=True)
+    start = fields.Datetime("Start", compute='_compute_start_stop', store=True)
```
ecino: `readonly`?

hr_attendance_calendar/models/hr_attendance_break.py
```diff
@@ -0,0 +1,75 @@
...
+    stop = fields.Datetime("Stop", compute='_compute_start_stop', store=True)
```
ecino: `readonly`?

hr_attendance_calendar/models/hr_attendance_break.py
```diff
@@ -0,0 +1,75 @@
...
+
+    previous_attendance = fields.Many2one('hr.attendance')
```
ecino: `previous_attendance_id`

hr_attendance_calendar/models/hr_attendance_break.py
```diff
@@ -0,0 +1,75 @@
...
+    next_attendance = fields.Many2one('hr.attendance')
```
ecino, line 39: `next_attendance_id`

hr_attendance_calendar/models/hr_attendance_break.py
```diff
@@ -0,0 +1,75 @@
...
+
+    system_modified = fields.Boolean('Modified by the system', default=False)
+
+    ##########################################################################
+    #                             FIELDS METHODS                             #
+    ##########################################################################
+    @api.multi
+    @api.depends('previous_attendance',)
```
ecino: It should depend as well on next_attendance_id

hr_attendance_calendar/models/hr_attendance_break.py
```diff
@@ -0,0 +1,75 @@
...
+    def _compute_start_stop(self):
+        for rd in self:
+            rd.start = rd.previous_attendance.check_out
+            rd.stop = rd.next_attendance.check_in
+
+    @api.multi
+    @api.depends('start', 'stop')
```
ecino, line 71: Since start and stop are updated at the same time, I would remove one depends to avoid calling twice the compute method.

hr_attendance_calendar/models/hr_attendance_break.py
```diff
@@ -0,0 +1,75 @@
...
+                att_break.duration = delta.total_seconds() / 3600.0
+
+    ##########################################################################
+    #                               ORM METHODS                              #
+    ##########################################################################
+
+    @api.multi
+    def write(self, vals):
+        for this in self:
+            if 'start' in vals:
```
ecino: Instead of doing this in the `write` method, use the method `inverse` on compute fields.

hr_attendance_calendar/models/hr_attendance_break.py
```diff
@@ -0,0 +1,75 @@
...
+                this.previous_attendance.write({
+                    'check_in': vals['start'],
+                    'no_compute_break': True,
+                })
+            if 'stop' in vals:
```
ecino: Instead of doing this in the `write` method, use the method `inverse` on compute fields.

hr_attendance_calendar/models/hr_attendance_break.py
```diff
@@ -0,0 +1,75 @@
...
+                this.next_attendance.write({
+                    'check_in': vals['stop'],
+                    'no_compute_break': True,
+                })
+            if 'system_modified' not in vals:
+                vals['system_modified'] = False
```
ecino: This is useless as it's already the default value.

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -0,0 +1,370 @@
...
+
+class HrAttendanceDay(models.Model):
+    """
+    The instances of hr.attendance.day is created either at the first
+    attendance of the day or by the method
+    hr.employee._cron_create_attendance() called by a cron everyday.
+    """
+    _name = "hr.attendance.day"
+    _order = 'date DESC'
+    _sql_constraints = [('unique_product', 'unique(date, employee_id)',
```
ecino, line 26: `unique_product` is a bad name for the constraint. `unique_date_employee` instead?

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -0,0 +1,370 @@
...
+                         'This "Attendance day" already exist for this '
+                         'employee')]
+
+    ##########################################################################
+    #                                 FIELDS                                 #
+    ##########################################################################
+    date = fields.Date("Date",
```
ecino, line 33: The string parameter is optional and not needed when it's the same name as the field. You can remove it here.

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -0,0 +1,370 @@
...
+                       # readonly=True, todo: uncomment
```
ecino, line 34: Uncomment?

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -0,0 +1,370 @@
...
+                       required=True)
+    employee_id = fields.Many2one('hr.employee', "Employee",
+                                  ondelete="cascade", required=True)
+
+    # Working schedule
+    cal_att_ids = fields.Many2many('resource.calendar.attendance',
+                                   string="Working schedule",
+                                   required=True)
+    working_day = fields.Char("Working day", compute='_compute_working_day',
```
ecino, line 43: The string parameter is not needed here.

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -0,0 +1,370 @@
...
+                              readonly=True, store=True)
+    name = fields.Char(compute='_compute_name', store=True)
```
ecino, line 45: readonly?

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -0,0 +1,370 @@
...
+
+    # Leaves
+    leave_ids = fields.Many2many('hr.holidays', readonly=True, string='Leaves')
+    in_leave = fields.Boolean('In leave', compute='_compute_in_leave',
```
ecino, line 49: The string parameter is not needed here.

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -0,0 +1,370 @@
...
+                              store=True)
+    public_holiday_id = fields.Many2one('hr.holidays.public.line',
+                                        'Public holidays')
```
ecino, line 52: It's a many2one, so the name should be Public holiday?

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -0,0 +1,370 @@
...
+
+    # Due hours
+    due_hours = fields.Float('Due hours', compute='_compute_due_hours',
```
ecino, line 55: The string parameter is not needed here.

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -0,0 +1,370 @@
...
+    worked_hours = fields.Float('Worked hours',
+                                compute='_compute_worked_hours', store=True,
+                                readonly=True)
+    coefficient = fields.Float(readonly=True, string='Coefficient',
+                               help='Worked hours coefficient')
+
+    # Break
+    due_break_min = fields.Float('Minimum break due',
+                                 compute='_compute_due_break')
+    due_break_total = fields.Float('Total break due',
```
ecino, line 72: I would give another name as `total`. Here you use that for the long break in case the employee worked more than 9 hours if I'm correct? Or why not having only one field named `due_break` that would take the long break or the small break depending on the worked hours? This would be more easy to understand in my opinion.

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -0,0 +1,370 @@
...
+                                   compute='_compute_due_break')
+    break_ids = fields.One2many('hr.attendance.break',
+                                'attendance_day_id',
+                                'Breaks',
+                                readonly=True)
+    break_max = fields.Float('Longest break',
```
ecino, line 78: What is the difference between `break_max` and `break_total`? I'm feeling we can a bit simplify here.

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -0,0 +1,370 @@
...
+                             compute='_compute_break_max',
+                             store=True, )
+    break_total = fields.Float('Total break',
+                               compute='_compute_break_total',
+                               store=True, )
+    rule_id = fields.Many2one('hr.attendance.rules', 'Rules',
```
ecino, line 84: The string should be Rule, not Rules.

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -0,0 +1,370 @@
...
+    due_break_total = fields.Float('Total break due',
+                                   compute='_compute_due_break')
+    break_ids = fields.One2many('hr.attendance.break',
+                                'attendance_day_id',
+                                'Breaks',
+                                readonly=True)
+    break_max = fields.Float('Longest break',
+                             compute='_compute_break_max',
+                             store=True, )
+    break_total = fields.Float('Total break',
```
ecino, line 81: In general (please take a look everywhere in your code), you should name your field like it appears in the view when possible, otherwise it is confusing. Here I would name it `total_break` and remove the string parameter.

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -0,0 +1,370 @@
...
+                               compute='_compute_break_total',
+                               store=True, )
+    rule_id = fields.Many2one('hr.attendance.rules', 'Rules',
+                              compute='_compute_rule_id')
+
+    # Extra hours
+    extra_hours = fields.Float("Extra hours",
```
ecino, line 88: String parameter is not needed.

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -0,0 +1,370 @@
...
+                att_day.date).strftime('%A')
+
+    @api.multi
+    @api.depends('working_day')
+    def _compute_name(self):
+        for rd in self.filtered('working_day'):
+            rd.name = rd.working_day + ' ' + rd.date
+
+    @api.multi
+    @api.depends('leave_ids')
```
ecino, line 111: It should depend on public_holiday_id as well

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -0,0 +1,370 @@
...
+    def _compute_in_leave(self):
+        for att_day in self:
+            if att_day.leave_ids:
```
ecino, line 114: Here you should do:
```
if att_day.leave_ids or att_day.public_holiday_id:
    att_day.in_leave = True
else:
    att_day.in_leave = False
```

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -0,0 +1,370 @@
...
+                att_day.in_leave = True
+
+    @api.multi
+    @api.depends('cal_att_ids', 'cal_att_ids.due_hours', 'leave_ids.state',
+                 'public_holiday_id')
+    def _compute_due_hours(self):
+        """First search the due hours based on the contract and after remove
+        somme hours if they are vacation"""
+        for att_day in self:
+            due_hours = 0
```
ecino: For taking the contract hours, you can do it in one line with :  `due_hours = sum(att_day.mapped('cal_att_ids.due_hours'))`

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -0,0 +1,370 @@
...
+            # Contract
+            for cal_att in att_day.cal_att_ids:
+                due_hours += cal_att.due_hours
+            # Leaves
+            if att_day.leave_ids:
+                for leave in att_day.leave_ids:
```
ecino: `for leave in att_day.leave_ids.filtered(lambda l: l.state in ('validate', 'validate1')):`

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -0,0 +1,370 @@
...
+                    if leave.state not in ['validate', 'validate1']:
+                        continue
```
ecino: Remove this

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -0,0 +1,370 @@
...
+                    if not leave.holiday_status_id.remove_from_due_hours:
+                        working_day = self.env[
+                            'base.config.settings'].get_work_day_duration()
+                        due_hours -= leave.compute_work_day(
+                            att_day.date) * working_day
+            # Public holidays
+            if att_day.public_holiday_id:
+                due_hours = 0
```
ecino: You could avoid all the above computation on the contract when on holiday, if you put this at first!

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -0,0 +1,370 @@
...
+    @api.depends('worked_hours', 'due_hours', 'coefficient')
+    def _compute_extra_hours(self):
+        for att_day in self.filtered('coefficient'):
+            extra_hours = att_day.worked_hours - att_day.due_hours
+            coefficient = att_day.coefficient
+            att_day.extra_hours = extra_hours * coefficient
+
+    @api.multi
+    def _breaks_validation(self):
+        for att_day in self:
```
ecino: Add a docstring and explain this method please.

PR #593, opened 2018-03-08
CO-1500 : bug fix with recurring contracts cancel

samuel-fringeli:
Note : another change is also in compassion-switzerland


ecino (changes requested):
This is not a good fix. The method `_check_need_sub` should only be called for sponsorships and not for other contracts, this would be the correct fix.

ecino (approved):

PR #594, opened 2018-03-08
FIX bad performance with event income and expense

ecino:


crm_compassion/models/account_invoice_line.py
```diff
@@ -21,6 +21,10 @@ class AccountInvoiceLine(models.Model):
     currency_id = fields.Many2one(
         'res.currency', 'Currency', related='invoice_id.currency_id',
         store=True)
+    event_id = fields.Many2one(
+        'crm.event.compassion', 'Event',
+        related='account_analytic_id.event_id', store=True, readonly=True
+    )
```
eicher31, line 27: Why use store=True in this case?

gift_compassion/models/sponsorship_gift.py
```diff
@@ -187,11 +187,27 @@ def _compute_usd(self):
     @api.model
     def create(self, vals):
         """ Try to find existing gifts before creating a new one. """
+        gift_date = vals.get('gift_date')
+        if not gift_date:
+            invl = self.env['account.invoice.line']
+            dates = []
+            default = fields.Date.today()
+            for invl_write in vals.get('invoice_line_ids', [[3]]):
+                if invl_write[0] == 0:
```
eicher31, line 196: Why [[3]]?

eicher31 (approved):

crm_compassion/models/account_invoice_line.py
ecino: To put it in the database and be able to search from the object directly

gift_compassion/models/sponsorship_gift.py
ecino: So that it goes in the else case when no values are set.

PR #598, opened 2018-03-13
CO-1523 Allow to add OMR code to letters

popod:


ecino:
Flake8 errors:
```
partner_communication/models/communication_attachment.py:68:51: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
partner_communication/models/ir_attachment.py:23:9: E128 continuation line under-indented for visual indent
partner_communication/models/communication_job.py:566:17: E303 too many blank lines (2)
partner_communication/models/communication_job.py:647:63: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
hr_attendance_calendar/reports/extra_hours_evolution_day_report.py:40:78: W291 trailing whitespace
```

partner_communication/models/communication_attachment.py
```diff
@@ -64,7 +64,23 @@ def create(self, vals):
...
+        count_attachment_with_omr = 1
         for attachment in self:
+            # add omr to pdf if needed
+            if attachment.communication_id.omr_enable_marks \
+                    and not attachment.attachment_id.disable_omr:
+                is_latest_document = \
+                    count_attachment_with_omr >= total_attachment_with_omr
+                to_print = self.env['partner.communication.job'] \
+                    .add_omr_marks(attachment.data, is_latest_document,
+                                   attachment.communication_id)
```
ecino: Instead of putting the communication in the parameters of the function, you should do
`to_print = attachment.communication_id.add_omr_marks(attachment.data, is_latest_document)`
And the function `add_omr_marks` in partner.communication should be decorated with `@api.multi` and then `self` would be used inside and hold the communication record.

partner_communication/models/communication_job.py
```diff
@@ -475,6 +478,114 @@ def message_post(self, **kwargs):
             })
         return message.id
 
+    @api.multi
+    def add_omr_marks(self, pdf_data, is_latest_document, communication):
```
ecino: Instead of communication parameter, you will use `self`. You can start your method with `self.ensure_one()` to show that you expect the method to be called on only one record at the time.

partner_communication/models/communication_job.py
```diff
@@ -475,6 +478,114 @@ def message_post(self, **kwargs):
...
+
+                # # number of pages in binary (MSB first with sequence:
+                # # 00, 01, 10, 11, 00, 01, ...)
+                # p.line(x1, y_position, x2, y_position)
+                # y_position -= y_step
+                #
+                # p.line(x1, y_position, x2, y_position)
+                # y_position -= y_step
+
+                # partiy mark (total number of marks should be pair)
```
ecino: parity

partner_communication/models/communication_job.py
```diff
@@ -475,6 +478,114 @@ def message_post(self, **kwargs):
...
+
+                # add the omr marks to the page
+                page = existing_pdf.getPage(page_number)
+                page.mergePage(omr_pdf.getPage(0))
+            else:
+                page = existing_pdf.getPage(page_number)
+
+            output.addPage(page)
+
+        # pdf_buffer.close()
```
ecino: You can remove if it's not needed

partner_communication/models/communication_job.py
```diff
@@ -522,10 +633,32 @@ def _print_report(self):
...
+                    job.subject or ''),
+                must_skip_send_to_printer=True
             ).get_pdf(job.ids, job.report_id.report_name)
+
+            # add omr to pdf if needed
+            if job.omr_enable_marks:
+                is_latest_document = False if \
+                    len(self.attachment_ids) > 0 \
+                    and len(self.attachment_ids.filtered(
+                        lambda r: r.attachment_id.disable_omr == False)) > 0 \
```
ecino: Here you could simplify `is_latest_document = not self.attachment_ids.filtered('enable_orm')`
and I propose that you change the field `disable_orm` to `enable_orm` so that you can use the filtered function with just the name of the field.

partner_communication/models/ir_attachment.py
```diff
@@ -19,3 +19,5 @@ class IrAttachment(models.Model):
         'ir.actions.report.xml', 'Print configuration',
         domain=[('property_printing_action_id.action_type', '=', 'server')]
     )
+    disable_omr = fields.Boolean(string="Disable OMR", help="If checked, "
```
ecino: I propose that you do instead `enable_omr = fields.Boolean(..., default=True)`

ecino (changes requested):

partner_communication/models/communication_job.py
popod: I change the name to `enable_omr` and the filter should be `.filtered('attachment_id.enable_omr')`

ecino (approved):

PR #625, opened 2018-04-04
C0-1594 & Pylint fixes

vwvw:


sponsorship_compassion/models/contracts.py
```diff
@@ -1,937 +1,927 @@
-# -*- coding: utf-8 -*-
```
ecino: Why is everything rewritten in this file?

sponsorship_compassion/models/res_partner.py
```diff
@@ -1,250 +1,263 @@
-# -*- coding: utf-8 -*-
```
ecino: Maybe an encoding issue? Please check why the whole file are rewrittent. It's not good for keeping track of the code changes...

vwvw:
Fixed. CRLF vs LF problem

ecino:
Doesn't seem to be fixed. The diffs are still on the entire files.

sponsorship_compassion/models/contracts.py
```diff
@@ -591,10 +591,6 @@ def contract_active(self):
                 'contract_active')
 
         # Update sponsorships on partner
-        partners = self.mapped('partner_id') | self.mapped('correspondant_id')
```
ecino: You can remove the above comment as well

sponsorship_compassion/models/res_partner.py
```diff
@@ -40,7 +40,8 @@ class ResPartner(models.Model):
     unrec_items = fields.Integer(compute='_compute_count_items')
     receivable_items = fields.Integer(compute='_compute_count_items')
     has_sponsorships = fields.Boolean()
-    number_sponsorships = fields.Integer()
+    number_sponsorships = fields.Integer(
+        compute='_compute_number_sponsorships')
```
ecino: The field should be stored with store=True. The problem then is that then we don't know what we can put in @api.depends to recompute at the good time...

sponsorship_compassion/models/res_partner.py
```diff
@@ -101,6 +102,18 @@ def _compute_children(self):
...
+    def _compute_number_sponsorships(self):
+        for partner in self:
+            partner.number_sponsorships = self.env[
+                'recurring.contract'].search_count([
+                    '|',
+                    ('partner_id', '=', partner.id),
+                    ('correspondant_id', '=', partner.id),
+                    ('is_active', '=', True),
+                    ('child_id', '!=', False)])
+            partner.has_sponsorships = partner.number_sponsorships > 0
```
ecino: We already saw that we cannot set other fields in compute methods

ecino (changes requested):

ecino:
Tests are broken

vwvw:
Fixed :-)

sponsorship_compassion/models/contracts.py
```diff
@@ -220,8 +220,12 @@ def write(self, vals):
         if 'child_id' in vals:
             self._on_change_child_id(vals)
 
+        if 'partner_id' in vals:
+            old_partner = self.partner_id
```
ecino: By doing this you prevent the write method to be called on multiple records at the time. You should not do `self.partner_id` but you should check for all possible records in self (add a loop maybe)

sponsorship_compassion/models/contracts.py
```diff
@@ -589,17 +601,20 @@ def contract_active(self):
             gift_contract_lines.mapped('contract_id').signal_workflow(
                 'contract_active')
 
-        # Update sponsorships on partner
-        partners = self.mapped('partner_id') | self.mapped('correspondant_id')
-        for partner in partners:
-            partner.number_sponsorships += 1
-            partner.has_sponsorships = partner.number_sponsorships
+        self.partner_id._compute_number_sponsorships()
```
ecino: Should be `self.mapped('partner_id')` to take all records in consideration

sponsorship_compassion/models/contracts.py
```diff
@@ -589,17 +601,20 @@ def contract_active(self):
...
                 'contract_active')
 
-        # Update sponsorships on partner
-        partners = self.mapped('partner_id') | self.mapped('correspondant_id')
-        for partner in partners:
-            partner.number_sponsorships += 1
-            partner.has_sponsorships = partner.number_sponsorships
+        self.partner_id._compute_number_sponsorships()
+        if not self.fully_managed:
+            self.correspondant_id._compute_number_sponsorships()
```
ecino: Use self.mapped

sponsorship_compassion/models/contracts.py
```diff
@@ -589,17 +601,20 @@ def contract_active(self):
...
+
         return True
 
     @api.multi
     def contract_cancelled(self):
         res = super(SponsorshipContract, self).contract_cancelled()
 
+        self.partner_id._compute_number_sponsorships()
```
ecino: mapped

sponsorship_compassion/models/contracts.py
```diff
@@ -589,17 +601,20 @@ def contract_active(self):
             gift_contract_lines.mapped('contract_id').signal_workflow(
                 'contract_active')
 
-        # Update sponsorships on partner
-        partners = self.mapped('partner_id') | self.mapped('correspondant_id')
-        for partner in partners:
-            partner.number_sponsorships += 1
-            partner.has_sponsorships = partner.number_sponsorships
+        self.partner_id._compute_number_sponsorships()
+        if not self.fully_managed:
```
ecino: mapped as well

sponsorship_compassion/models/contracts.py
```diff
@@ -589,17 +601,20 @@ def contract_active(self):
...
+            self.correspondant_id._compute_number_sponsorships()
+
         return True
 
     @api.multi
     def contract_cancelled(self):
         res = super(SponsorshipContract, self).contract_cancelled()
 
+        self.partner_id._compute_number_sponsorships()
+        if not self.fully_managed:
```
ecino: mapped

sponsorship_compassion/models/contracts.py
```diff
@@ -589,17 +601,20 @@ def contract_active(self):
...
+            self.correspondant_id._compute_number_sponsorships()
```
ecino: mapped

sponsorship_compassion/models/contracts.py
```diff
@@ -608,6 +623,10 @@ def contract_cancelled(self):
     def contract_terminated(self):
         res = super(SponsorshipContract, self).contract_terminated()
 
+        self.partner_id._compute_number_sponsorships()
```
ecino: mapped... you get the idea ;)

sponsorship_compassion/models/res_partner.py
```diff
@@ -39,8 +39,9 @@ class ResPartner(models.Model):
         string='Other contracts')
     unrec_items = fields.Integer(compute='_compute_count_items')
     receivable_items = fields.Integer(compute='_compute_count_items')
-    has_sponsorships = fields.Boolean()
-    number_sponsorships = fields.Integer()
+    has_sponsorships = fields.Boolean(compute='_compute_has_sponsorships',
+                                      store=True)
+    number_sponsorships = fields.Integer(store=True)
```
ecino: store parameter is useless for non computed field

ecino (changes requested):

sponsorship_compassion/models/contracts.py
```diff
@@ -589,17 +601,20 @@ def contract_active(self):
             gift_contract_lines.mapped('contract_id').signal_workflow(
                 'contract_active')
 
-        # Update sponsorships on partner
-        partners = self.mapped('partner_id') | self.mapped('correspondant_id')
-        for partner in partners:
-            partner.number_sponsorships += 1
-            partner.has_sponsorships = partner.number_sponsorships
+        self.mapped('partner_id')._compute_number_sponsorships()
+        self.filtered(lambda c: c.fully_managed).mapped(
```
ecino: the filter should be `not c.fully_managed`

sponsorship_compassion/models/contracts.py
```diff
@@ -589,17 +601,20 @@ def contract_active(self):
...
+            'correspondant_id')._compute_number_sponsorships()
+
         return True
 
     @api.multi
     def contract_cancelled(self):
         res = super(SponsorshipContract, self).contract_cancelled()
 
+        self.mapped('partner_id')._compute_number_sponsorships()
+        self.filtered(lambda c: c.fully_managed).mapped(
```
ecino: the filter should be `not c.fully_managed`

sponsorship_compassion/models/contracts.py
```diff
@@ -608,6 +623,10 @@ def contract_cancelled(self):
     def contract_terminated(self):
         res = super(SponsorshipContract, self).contract_terminated()
 
+        self.mapped('partner_id')._compute_number_sponsorships()
+        self.filtered(lambda c: c.fully_managed).mapped(
```
ecino: the filter should be `not c.fully_managed`

sponsorship_compassion/models/contracts.py
```diff
@@ -805,12 +824,6 @@ def _on_sponsorship_finished(self):
                     ('object_id', '=', sponsorship.id),
                 ]).unlink()
 
-        # Update sponsorships on partner
```
ecino: Instead of having the same code twice in the `contract_cancelled` and `contract_terminated` methods, you can move your computation here in the `_on_sponsorship_finished` method.

sponsorship_compassion/models/res_partner.py
ecino: The encoding issue seems to be back on this file...

ecino (changes requested):
Some last few changes before we can merge

ecino (approved):

PR #636, opened 2018-04-12
CO-1548 : exchange of working days in attendances - beta

samuel-fringeli:
view for forced hours and manager messages not created yet

### DIFF
--- a/hr_attendance_calendar/models/hr_attendance_day.py
+++ b/hr_attendance_calendar/models/hr_attendance_day.py
@@ -83,7 +83,7 @@ class HrAttendanceDay(models.Model):
                                 readonly=True)
     break_total = fields.Float('Total break',
                                compute='_compute_break_total',
-                               store=True, )
+                               store=True)
     rule_id = fields.Many2one('hr.attendance.rules', 'Rules',
                               compute='_compute_rule_id')
 
@@ -125,6 +125,17 @@ class HrAttendanceDay(models.Model):
         """First search the due hours based on the contract and after remove
         some hours if they are vacation"""
         for att_day in self:
+
+            # Forced due hours (when an user changes work days)
+            res = self.env['hr.forced.due.hours'].search([
+                ('employee_id', '=', att_day.employee_id.id),
+                ('date', '=', att_day.date),
+            ])
+            # res.forced_due_hours returns False if exists and has value 0.0
+            # that's why we write len(res) == 1
+            if len(res) == 1:
+                att_day.due_hours = res.forced_due_hours
+                continue
 
             # Public holidays
             if att_day.public_holiday_id:
--- a/hr_attendance_calendar/models/hr_attendance_day.py
+++ b/hr_attendance_calendar/models/hr_attendance_day.py
@@ -58,6 +58,11 @@ class HrAttendanceDay(models.Model):
     # Attendances
     attendance_ids = fields.One2many('hr.attendance', 'attendance_day_id',
                                      'Attendances', readonly=True)
+    has_logged_hours = fields.Boolean('Has logged hours',
+                                      compute='_compute_has_logged_hours',
+                                      store=True)
+    has_linked_change_day_request = fields.Boolean(
+        compute='_compute_has_linked_change_day_request', store=True)
 
     # Worked
     paid_hours = fields.Float(
@@ -97,6 +102,27 @@ class HrAttendanceDay(models.Model):
     #                             FIELDS METHODS                             #
     ##########################################################################
     @api.multi
+    @api.depends('attendance_ids')
+    def _compute_has_logged_hours(self):
+        for att_day in self:
+            # > 1 because when we calculate this, we're creating an attendance
+            att_day.has_logged_hours = (len(att_day.attendance_ids) > 1)
+
+    @api.multi
+    def get_related_forced_due_hours(self):
+        self.ensure_one()
+        return self.env['hr.forced.due.hours'].search([
+            ('employee_id', '=', self.employee_id.id),
+            ('date', '=', self.date)])
+
+    @api.multi
+    @api.depends('due_hours')
+    def _compute_has_linked_change_day_request(self):
+        for att_day in self:
+            res = att_day.get_related_forced_due_hours()
+            att_day.has_linked_change_day_request = len(res) == 1
+
+    @api.multi
     @api.depends('date')
     def _compute_working_day(self):
         for att_day in self.filtered('date'):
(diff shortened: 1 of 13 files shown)


hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -58,6 +58,10 @@ class HrAttendanceDay(models.Model):
     # Attendances
     attendance_ids = fields.One2many('hr.attendance', 'attendance_day_id',
                                      'Attendances', readonly=True)
+    has_logged_hours = fields.Boolean('Has logged hours',
+                                      compute='_compute_has_logged_hours')
+    has_linked_change_day_request = fields.Boolean(
+        default=False, compute='_compute_has_linked_change_day_request')
```
ecino: default is useless for a compute field

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -96,6 +100,27 @@ class HrAttendanceDay(models.Model):
     ##########################################################################
     #                             FIELDS METHODS                             #
     ##########################################################################
+    @api.multi
+    @api.depends('attendance_ids')
+    def _compute_has_logged_hours(self):
+        for att_day in self:
+            # > 1 because when we calculate this, we're creating an attendance
```
ecino: Are you sure this is still the case? Normally the compute method is called when the field is accessed for the first time (when the field is not stored).

ecino: I feel this field is useless, when needed you can simply do `if att_day.attendance_ids` to check if there are logged hours

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -96,6 +100,27 @@ class HrAttendanceDay(models.Model):
...
+
+    @api.multi
+    def get_related_forced_due_hours(self):
+        self.ensure_one()
+        return self.env['hr.forced.due.hours'].search([
+            ('employee_id', '=', self.employee_id.id),
+            ('date', '=', self.date)])
+
+    @api.multi
+    @api.depends('due_hours')
```
ecino, line 110: depends is useless when compute field is not stored

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -126,6 +151,14 @@ def _compute_due_hours(self):
         some hours if they are vacation"""
         for att_day in self:
 
+            # Forced due hours (when an user changes work days)
+            res = att_day.get_related_forced_due_hours()
```
ecino: forced_hours would be better name than res

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -126,6 +151,14 @@ def _compute_due_hours(self):
...
+            # res.forced_due_hours returns False if exists and has value 0.0
+            # that's why we use len(res) == 1
+            if len(res) == 1:
```
ecino: Your comment is not correct. Here you can simply do `if forced_hours`

ecino (changes requested):
I've not had time to do the full review. Here are some first comments.

hr_attendance_calendar/models/hr_forced_due_hours.py
```diff
@@ -0,0 +1,119 @@
...
+    date = fields.Date('Date', required=True)
+    forced_due_hours = fields.Float('Forced due hours', required=True)
+
+    @api.model
+    def recompute_due_hours(self, employee_id, date):
+        return self.env['hr.attendance.day'].search(
+            [('employee_id', '=', employee_id),
+             ('date', '=', date)]).recompute_due_hours()
+
+    @api.one
```
ecino: api.one is deprecated and should not be used, especially not for write methods

hr_attendance_calendar/models/hr_forced_due_hours.py
```diff
@@ -0,0 +1,119 @@
...
+    def write(self, vals):
+        old_vals = dict()
+        for field in ['employee_id', 'date', 'forced_due_hours']:
+            old_vals[field] = getattr(self, field)
+
+        res = super(HrForcedDueHours, self).write(vals)
+
+        self.recompute_due_hours(
+            old_vals['employee_id'].id, old_vals['date'])
```
ecino: old_vals['employee_id'] should already contain an id and not a record, so remove `.id`

hr_attendance_calendar/models/hr_attendance_day.py
samuel-fringeli: The field is not used in hr.attendance.day model, but a related field of this is needed in hr.attendance, to show the wizard only if there are already logged hours.
And for len > 1, the field depends on attendance_ids, so it's recomputed each time an attendance_day has a new attendance

hr_attendance_calendar/models/hr_forced_due_hours.py
samuel-fringeli: No, because `old_vals['employee_id'] = self.employee_id` that is an `hr.employee()` object

ecino: If you do that it's wrong! The write method never takes records as parameters in the dictionary. It should always be ids.

hr_attendance_calendar/wizards/hr_change_day_wizard.py
```diff
@@ -0,0 +1,92 @@
...
+        for index, date in enumerate(dates):
+            #  only get days with due hours
+            if due_hours[index] != 0:
+                data[date] = due_hours[index]
+
+        #  delete created att_days
+        unlinked = [x.unlink() for x in att_day_created].count(False) == 0
+
+        if unlinked:
+            return data
```
ecino: Why not returning data in any cases?

ecino:
Please rebase on devel and fix any pylint and flake8 issues

hr_attendance_calendar/models/hr_forced_due_hours.py
```diff
@@ -0,0 +1,118 @@
...
+            super(HrForcedDueHours, record).write(vals)
+
+            record.recompute_due_hours(
+                old_vals['employee_id'].id, old_vals['date'])
+            record.recompute_due_hours(record.employee_id.id, record.date)
+
+    @api.multi
+    def unlink(self):
+        length = len(self)
+        employee_ids = [0]*length
```
ecino: Better, you can do `employee_ids = self.mapped('employee_id.id')` and dates = `self.mapped('date')`

ecino (changes requested):

eicher31 (approved):

ecino (approved):

PR #642, opened 2018-04-18
CO-1635 Fixing pylint issues

vwvw:
One warning is left. The csv is used. Tried to disable the warning with a pylintrc without success.


hr_planning/models/resource_calendar.py
```diff
@@ -12,27 +12,28 @@
...
+        res = super(ResourceCalendar, self).write(vals)
 
         if ('attendance_ids' in vals):
             self._generate()
         return res
 
-    @api.one
+    @api.multi
     def _generate(self):
+        self.ensure_one()
```
ecino: You should better add a loop through the records in self. I see at line 31 that the method could be called on several records.

sponsorship_tracking/wizards/install_sds_tracking.py
```diff
@@ -1,136 +1,138 @@
-# -*- coding: utf-8 -*-
```
ecino: CLRF issue?

ecino (changes requested):

sponsorship_tracking/wizards/install_sds_tracking.py
vwvw: Yup

vwvw:
Should be fixed now

ecino (approved):

PR #648, opened 2018-04-20
Fixing call button raising an error and failing request to an icon

vwvw:
Icon is not present in odoo any more.
The click2dial function take 4 instead of 5 arguments. 
(self, phone_number, click2dial_model, click2dial_id)

### DIFF
--- a/partner_communication/models/communication_job.py
+++ b/partner_communication/models/communication_job.py
@@ -13,14 +13,13 @@ import logging
 from HTMLParser import HTMLParser
 
 from odoo.addons.base_phone import fields as phone_fields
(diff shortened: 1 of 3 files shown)


vwvw:
One flake8 that I'll correct soon

vwvw:
I had to remove the call to the BaseController because the http.request environment is only set if the method is called as a web requests. Since it is not the case, the method would crash. Replaced it with a call to self.env[...]

The wrong number of argument corrected in the first commit is due to the switch to odoo 10.0 which removed the request argument.

ecino (approved):

PR #649, opened 2018-04-23
CO-1548 : improvements

samuel-fringeli:


ecino (changes requested):
Field employee_id should be required.

ecino:
You will be able to fix it in a new commit, I merge for putting it in production.

PR #650, opened 2018-04-23
Small adjustemens and correction for partner_communication module

vwvw:


ecino (approved):
Great changes!

vwvw:
I was wondering if we should test if the address is present in the partner before printing. We should also extend [this] (https://github.com/CompassionCH/compassion-modules/pull/650/files#diff-4829b8a5d165e20993a42ee774c54e5aR226) test for the case where the send_mode='both'. Would you approve these changes?

PR #664, opened 2018-04-27
HR attendance refactor

ecino:
Since the name of the module is changed, I added a symlink for the migration, we can remove it afterwards.


hr_attendance_calendar/models/hr_attendance.py
```diff
@@ -37,43 +35,51 @@ def create(self, vals):
...
+            ])
+            if not attendance_day:
+                attendance_day = self.env['hr.attendance.day'].create({
+                    'employee_id': employee_id,
+                    'date': date
+                })
+            else:
+                # A modified attendance should update related breaks
+                attendance_day.compute_breaks()
+            attendance_days |= attendance_day
```
vwvw, line 84: attendance_days.update(attendance_day) seems clearer to me.

hr_attendance_calendar/models/hr_attendance_day.py
```diff
@@ -154,21 +151,11 @@ def _compute_due_hours(self):
...
-                for leave in att_day.leave_ids.filtered(
-                        lambda l: l.state in ('validate', 'validate1')):
-                    if not leave.holiday_status_id.remove_from_due_hours:
-                        working_day = self.env[
-                            'base.config.settings'].get_work_day_duration()
-                        due_hours -= leave.compute_work_day(
-                            att_day.date) * working_day
+            for leave in att_day.leave_ids.filtered(
+                    lambda l: l.state == 'validate' and not l.keep_due_hours):
+                due_hours -= leave.compute_leave_time(att_day.date)
```
vwvw, line 158: compute_leave_time is in range [0-24]. Are you sure that you want to deduce 24 hours of work here? 
In particular, if I leave at noon, leave time will return 12 but due hours should probably be reduced only by half. no?

ecino: We discussed with David and choose to let it like that. This scenario will almost never happen and if it happens, the only consequence is the employee will have a bit more extra hours. The manager can correct this manually if needed (or can catch it before approving the leave request).

hr_attendance_calendar/models/hr_attendance.py
ecino: Not producing the same result. Update method on recordset is something else...

ecino:
I will add the tests directly in this PR

hr_attendance_calendar/models/hr_attendance_day.py
vwvw: Yup sounds good to me then !

eicher31 (approved):

PR #674, opened 2018-05-23
CO-1695 [ADD] Module asterisk_autolog

eicher31:


hr_attendance_calendar/static/src/js/attendance.js
```diff
@@ -124,6 +124,13 @@ odoo.define('hr_switzerland.exchange_days_wizard', function(require) {
                 window.find_parent_by_tag_name($('#select_change_day_message'), 'table').css('margin-bottom', '-8px');
             };
 
+            if (this.attendance.check_out) {
+                this.do_action('hr_attendance_extra_hours.check_out_xivo');
```
ecino: I don't think defining an action and a new model is required. The model `attendance.js.log` does not receive any information more than "someone has checked_in". I think it's better to add the method `on_check_in` and `on_check_out` in the model `hr.employee`, make it api.multi, and call it from python by overriding this function already defined [here in hr_employee](https://github.com/odoo/odoo/blob/10.0/addons/hr_attendance/models/hr_employee.py#L102).

This way, you can remove the javascript, the server action and the new AbstractModel.

ecino (changes requested):

ecino (approved):

PR #675, opened 2018-05-29
CO-1681 Privacy data

eicher31:


ecino (approved):
Ok like this, but we will add a compute field (with inverse method) to manually control when the partner created will have an agreement or not.

PR #680, opened 2018-06-01
CO-1681 [ADD] Personal data agreement

eicher31:


sponsorship_compassion/models/res_partner.py
```diff
@@ -98,6 +99,28 @@ def _compute_count_items(self):
                 ('partner_id', '=', partner.id),
                 ('account_id.code', '=', '1050')])
 
+    @api.multi
+    def set_privacy_statement(self, origin):
+        for partner in self:
+            if partner.signed_privacy_statement:
```
ecino: `partner.signed_privacy_statement` is not defined. You should add this field I guess or do something else.

ecino (changes requested):
Tests are failing because you have an error in your code (variable reference before assignment).
Flake8 :
```
sponsorship_compassion/models/res_partner.py:54:1: W293 blank line contains whitespace```

PR #693, opened 2018-06-27
CO-1790: changing HTML field into text field with communication chang…

samuel-fringeli:
…e wizard


ecino (approved):
I will test on devel

PR #706, opened 2018-07-09
CO-1639 Tests for gift_compassion

quentingigon:


gift_compassion/models/sponsorship_gift.py
```diff
@@ -194,7 +194,7 @@ def create(self, vals):
             default = fields.Date.today()
             for invl_write in vals.get('invoice_line_ids', [[3]]):
                 if invl_write[0] == 0:
-                    dates.append(invl_write[2].get('due_date'), default)
+                    dates.append(invl_write[2].get('due_date'))
```
ecino: Here the correct fix is to move the `, default` inside the get method. That was the intent in case the invoice line doesn't contain a `due_date`, it takes today as the date.

ecino (changes requested):

ecino (approved):

ecino:
Thanks, again I squashed a bit those commits to have a proper history in devel. You might need to rebase before your next pull request. :+1:

PR #707, opened 2018-07-09
Co 1638 tests for partner communication

nbornand:
add tests for partner communication

### DIFF
--- a/partner_communication_revision/data/demo.xml
+++ b/partner_communication_revision/data/demo.xml
@@ -14,10 +14,10 @@
                     ${partner.salutation},
                     <br/>
                     <br/>
-                    We thank you for the donation of CHF .
-                    % if partner.partner:
+                    We thank you for the donation of CHF
+                    % if partner.email:
                         you made by email.
-                    % else
+                    % else:
                         you made.
                     % endif
                     We appreciate your generosity.
--- a/partner_communication_revision/data/demo.xml
+++ b/partner_communication_revision/data/demo.xml
@@ -1,0 +1,38 @@
+<odoo>
+    <data noupdate="1">
+        <record id="template" model="mail.template">
+            <field name="name">Donation - Thank You Letter</field>
+            <field name="model_id" ref="partner_communication.model_partner_communication_job"/>
+            <field name="use_default_to" eval="True"/>
+            <field name="subject">Donation Receipt</field>
+            <field name="body_html" type="html">
+                <div>
+                    % set invoice_lines = object.get_objects()
+                </div>
+                <div>
+                    ${partner.salutation},
+                    <br/>
+                    <br/>
+                    We thank you for the donation of CHF ${donations[0]}.-
+                    % if donations[1]:
+                        you made for ${donations[1]}.
+                    % else
+                        you made.
+                    % endif
+                    We appreciate your generosity.
+                    <br/>
+                    <br/>
+                    Yours sincerely
+                </div>
+            </field>
+        </record>
+        <record id="test_communication2" model="partner.communication.config">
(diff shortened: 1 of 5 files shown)


ecino:
These are the pylint errors you are concerned about. You can ignore the other warnings coming from the other modules:
```
************* Module partner_communication_revision
partner_communication_revision/data/demo.xml:1: [W7908(missing-newline-extrafiles), ]  Missing newline
************* Module partner_communication_revision.tests.test_partner_communication
partner_communication_revision/tests/test_partner_communication.py:75: [W1401(anomalous-backslash-in-string), ] Anomalous backslash in string: '\s'. String constant might be missing an r prefix.
partner_communication_revision/tests/test_partner_communication.py:76: [W1401(anomalous-backslash-in-string), ] Anomalous backslash in string: '\s'. String constant might be missing an r prefix.
```

partner_communication_revision/data/demo.xml
```diff
@@ -0,0 +1,39 @@
...
+<odoo>
+    <data>
+        <record id="template" model="mail.template">
+            <field name="name">Donation - Thank You Letter</field>
+            <field name="model_id" ref="partner_communication.model_partner_communication_job"/>
+            <field name="use_default_to" eval="True"/>
+            <field name="subject">Donation Receipt</field>
+            <field name="body_html" type="html">
+                <div>
+                    % set invoice_lines = object.get_objects()
```
ecino: get_objects() is a method in `partner.communication.job` that returns the related objects. In this case, since in your config you put the related model as `res_partner`, the method will return a `res.partner` recordset. Since you don't use invoice_lines later in the template, it's not a real problem but you could rename this variable for better clarity.

partner_communication_revision/data/demo.xml
```diff
@@ -0,0 +1,39 @@
...
+                </div>
+                <div>
+                    ${partner.salutation},
```
ecino: This will fail when generating a letter because `partner` is not a variable declared in the template. Please correct it because some people may use the demo data to try the functionality of the module in an empty database instance, so we want this to work.

partner_communication_revision/data/demo.xml
```diff
@@ -0,0 +1,39 @@
...
+                    <br/>
+                    <br/>
+                    We thank you for the donation of CHF ${donations[0]}.-
+                    % if donations[1]:
```
ecino: `donations` will fail at template rendering since the variable is not declared. You can remove this part for simplicity.

ecino (changes requested):

partner_communication_revision/data/demo.xml
nbornand: I indeed gathered random parts of templates without checking that the whole thing was valid :( Should be better now.

partner_communication_revision/data/demo.xml
```diff
@@ -0,0 +1,39 @@
...
+            <field name="body_html" type="html">
+                <div>
+                    % set partner = object.get_objects()
+                </div>
+                <div>
+                    ${partner.salutation},
+                    <br/>
+                    <br/>
+                    We thank you for the donation of CHF .
+                    % if partner.partner:
```
ecino: I guess you wanted to put `if partner.email` ? I think `partner` is not a valid field inside the `res.partner` object.

partner_communication_revision/data/demo.xml
```diff
@@ -0,0 +1,39 @@
...
+                        you made by email.
+                    % else
```
ecino: Missing the `:` (`% else:` )

ecino (approved):

PR #716, opened 2018-07-12
AP-16 add dedicated endpoint to return pdfs

nbornand:


mobile_app_connector/controllers/mobile_app_controller.py
```diff
@@ -14,6 +14,7 @@
 from werkzeug.exceptions import NotFound, MethodNotAllowed
 
 
+@http.route('/mobile-app-api')
```
ecino: I never saw routes at the class-level, what is that for?

ecino (changes requested):
Just one pylint error:
mobile_app_connector/controllers/mobile_app_controller.py:48:59: W292 no newline at end of file

mobile_app_connector/controllers/mobile_app_controller.py
nbornand: most frameworks let you specify routes at both the class and method level (they get concatenated) so I was curious to see if it worked ;) will get rid of it

ecino (approved):

PR #720, opened 2018-07-17
Co 1909 sms model

nbornand:


sms_sponsorship/models/compassion_hold.py
```diff
@@ -0,0 +1,57 @@
...
+        self.booked_by_phone_number = None
+        self.booked_by_sms_at = None
+
+    @api.multi
+    def generate_url_of_next_sms_sponsoring_step(self):
+        # todo the exact URLs will be changed
+        self.ensure_one()
+        partner = self.env['res.partner'] \
+            .search([('phone', '=', self.booked_by_phone_number)])
+        child_id = self.env['compassion.child'] \
```
ecino: `self.child_id` is a valid field in the hold and is directly pointing to the correct child.

sms_sponsorship/models/compassion_hold.py
```diff
@@ -0,0 +1,57 @@
...
+        self.ensure_one()
+        self.booked_by_phone_number = None
+        self.booked_by_sms_at = None
+
+    @api.multi
+    def generate_url_of_next_sms_sponsoring_step(self):
+        # todo the exact URLs will be changed
+        self.ensure_one()
+        partner = self.env['res.partner'] \
+            .search([('phone', '=', self.booked_by_phone_number)])
```
ecino: I would first search on the `mobile` field and if no partner is found then search on the `phone` field

sms_sponsorship/models/compassion_hold.py
```diff
@@ -0,0 +1,57 @@
...
+        child_id = self.env['compassion.child'] \
+            .search([('hold_id', '=', self.id)]).id
+        if partner:
+            return 'sms-sponsorship/{}/partner/{}' \
```
ecino: I think it would be better to have a link related to the hold instead of the  child and the partner. Then the controller of step 1 can by itself lookup for the partner or the phone number.

sms_sponsorship/__manifest__.py
```diff
@@ -16,7 +16,10 @@
     'license': 'AGPL-3',
     'website': 'http://www.compassion.ch',
     'depends': ['crm_compassion', 'cms_form'],
-    'data': [],
+    'data': [
+        'data/sms_hold_release.xml',
+        'data/event_compassion_view.xml'
```
ecino: views should be placed inside a `views` folder for consistency. You should add as well a view for displaying the new fields you added in the compassion.hold object.

ecino (changes requested):

sms_sponsorship/models/compassion_hold.py
nbornand: ok, maybe we can simply have an or clause with the two filters?

sms_sponsorship/models/compassion_hold.py
nbornand: Sounds better indeed

sms_sponsorship/models/compassion_hold.py
ecino: Yes that seems sufficient. Or can be done like this : `['|', (cond1), (cond2)]`

sms_sponsorship/views/event_compassion_view.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+   <record id="view_events_form" model="ir.ui.view">
+       <field name="name">crm.event.compassion.form</field>
+       <field name="model">crm.event.compassion</field>
+       <field name="inherit_id" ref="crm_compassion.view_events_form"/>
+       <field name="arch" type="xml">
+           <xpath expr="//field[@name='planned_sponsorships']" position="after">
```
ecino, line 8: `<field name="planned_sponsorships" position="after">`

ecino (approved):

PR #722, opened 2018-07-18
CO-1886 save campain on hold and copy it on sponsorships

nbornand:


ecino (commented):
Are you sure your editor has the right file format enabled?
It's hard to review since all file is changed...
image

child_compassion/models/compassion_hold.py
```diff
@@ -127,6 +127,7 @@ def get_hold_values(self):
         self.ensure_one()
         vals = self.read(self.get_fields())[0]
         vals['primary_owner'] = vals['primary_owner'][0]
+        vals['campaign_id'] = self.campaign_id.id
```
ecino: This fix is not working when only child_compassion module is installed, as `campaign_id` is defined in module `crm_compassion`. I looked at it while doing the hold request for SMS sponsorships, you can look the fix here : https://github.com/CompassionCH/compassion-modules/pull/724/commits/403b9ef34c619a8b46a848386cecd064f055b694

So I propose that you just remove this line and we can keep the name signature of the method you changed.

ecino (changes requested):

ecino (approved):

PR #738, opened 2018-07-26
Co 1885 omr left and config per language

nbornand:
Moving the OMR on the left side and making it configurable on a per language basis


ecino:
The model partner.communication.omr.config has no access rules, consider adding one (this is causing Travis test to fail)

ecino (approved):

PR #743, opened 2018-07-27
AP-11 Get message implemented

quentingigon:
I am not sure about values for links fields, do we do calls on compassion UK URLs ?

### DIFF
--- a/mobile_app_connector/models/compassion_child.py
+++ b/mobile_app_connector/models/compassion_child.py
@@ -47,12 +47,6 @@ class CompassionChild(models.Model):
             result.append(mapping.get_connect_data(child))
         return result
 
-    @api.model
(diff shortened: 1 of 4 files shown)


mobile_app_connector/models/compassion_child.py
```diff
@@ -47,6 +47,17 @@ def mobile_sponsor_children(self, **other_params):
             result.append(mapping.get_connect_data(child))
         return result
 
+    @api.model
+    def mobile_get_child_bio(self, **other_params):
+        values = dict(other_params)
+        child = self.env['compassion.child'].search([
```
ecino: this is never used.

ecino (changes requested):
lint is failing

ecino (approved):

PR #747, opened 2018-07-31
CO-1932 Error Message

nfluckiger:


sms_sponsorship/webapp/src/App.js
```diff
@@ -41,13 +41,16 @@ class Main extends React.Component {
     };
 
     getChild = () => {
-        let lang = i18n.languages[1];
-        let requestId = getRequestId();
-        if (!requestId) {
+        if (!getRequestId()) {
```
ecino: Here you are reverting the code to an old version...

sms_sponsorship/webapp/src/App.js
```diff
@@ -41,13 +41,16 @@ class Main extends React.Component {
...
             return;
         }
-        let url = "/sms_sponsorship/step1/" + requestId + "/get_child_data";
-        jsonRPC(url, {lang: lang}, (res) => {
+        let url = "/sms_sponsorship_api";
```
ecino: This is an old route version, please revert this

sms_sponsorship/webapp/src/App.js
```diff
@@ -58,16 +61,10 @@ class Main extends React.Component {
     };
 
     changeChild = () => {
-        let requestId = getRequestId();
-        let lang = i18n.languages[1];
-        let url = "/sms_sponsorship/step1/" + requestId + "/change_child";
-        let form = document.forms.other_child_form;
+        let url = "/sms_change_child";
```
ecino: This is an old version

sms_sponsorship/webapp/src/App.js
```diff
@@ -99,81 +95,80 @@ class Main extends React.Component {
                 <div>
                     {topAppBar}
                     <div>
-                        <Message text={t("error_noRequestID")}/>
+                        <Error appContext={this} text="Error : no SMS request ID. Please send an other SMS."/>
```
ecino: Text should be placed in the translations and use here the translate function. See how it's done elsewhere. This text is already existing in the translations so you just need to lookup its id in the file i18n.js and fetch it: `t("error_noRequestID")`

sms_sponsorship/webapp/src/App.js
```diff
@@ -99,81 +95,80 @@ class Main extends React.Component {
...
                     </div>
                 </div>
             )
         }
 
-        if (!child.has_a_child && !this.state.success) {
+        if (!child.has_a_child) {
```
ecino: This is a bad revert

sms_sponsorship/webapp/src/App.js
```diff
@@ -99,81 +95,80 @@ class Main extends React.Component {
...
             clearTimeout(window.getChildTimeout);
-            window.getChildTimeout = setTimeout(() => {
+            if(this.count_try > 20){
+                return (
+                            <div>
+                                <Error appContext={this} text="Error : No child found. Try again later"/>
```
ecino: Text should be placed in the translations file and loaded with the translate function.

sms_sponsorship/webapp/src/App.js
```diff
@@ -99,81 +95,80 @@ class Main extends React.Component {
...
+                window.getChildTimeout = setTimeout(() => {
                 this.getChild();
-            }, 1000);
+                }, 1000);
+                ++this.count_try;
+            }
+        }
+        if (this.state.success) {
+            return (
+                <div>success !</div>
```
ecino: This is again a very bad revert that cancels a lot of changes we made. We have a better success message than this.

sms_sponsorship/webapp/src/App.js
```diff
@@ -99,81 +95,80 @@ class Main extends React.Component {
...
+                )}
+            </div>
         )
     }
 }
 
-// extended main view with translate hoc
-export default translate('translations')(Main);
\ No newline at end of file
+export default () => {
```
ecino: bad revert

ecino (changes requested):
We see here that it's a bad practice to copy/paste a file to avoid conflicts, because you have the risk to push back old code. The best way to update one's work upon the latest repository changes is to use `git rebase` and fix the conflicts.

sms_sponsorship/webapp/src/App.js
ecino: Much better now, but this if condition still needs to be reverted

ecino (approved):

PR #749, opened 2018-08-02
CO-1928 Tests for sms_sponsorship

quentingigon:


quentingigon:
Travis test was failing because US servers were down

sms_sponsorship/tests/test_sms_compassion.py
```diff
@@ -19,12 +19,112 @@ class TestSmsCompassion(BaseSponsorshipTest):
 
     def setUp(self):
         super(TestSmsCompassion, self).setUp()
-        self.child = self.create_child('UG1119182')
-        self.child.hold_id.channel = 'sms'
+        self.child = self.env['compassion.child'].search([
+            ('id', '=', 5)
```
ecino: Never browse by id in test, you never know what the id will be on another test database... Why did you change how it was done previously?

sms_sponsorship/tests/test_sms_compassion.py
```diff
@@ -19,12 +19,112 @@ class TestSmsCompassion(BaseSponsorshipTest):
...
+        ])
+        # self.child.hold_id.channel = 'sms'
+        self.partner = self.env['res.partner'].search([
+            ('id', '=', 15)
```
ecino: Not a good practice, you should use self.env.ref() instead and find a xml_id in the demo data of base module.

sms_sponsorship/tests/test_sms_compassion.py
```diff
@@ -19,12 +19,112 @@ class TestSmsCompassion(BaseSponsorshipTest):
...
+        ])
+        self.event = self.env['crm.event.compassion'].search([
+            ('id', '=', 1030)
```
ecino: This will clearly fail on empty databases.

ecino (changes requested):

sms_sponsorship/tests/test_sms_compassion.py
quentingigon: It didn't work with create_child method 
(Error: IntegrityError: duplicate key value violates unique constraint "compassion_project_icp_id_uniq")

ecino:
Oops, again confilcts :( tests are failing as well.

quentingigon:
My tests are working fine locally and fail on the server. I have this error : 
UserError: ('Please give connect_url and connect_api_key values in your Odoo configuration file.', ''). But there are values for those fields in odoo.conf..

ecino:
Odoo.conf is not committed and not available to travis (this would reveal all our connection endpoints with username password, so please don't commit it). You should therefore have tests that never use them or mock them to avoid using real connections.

ecino (approved):

PR #754, opened 2018-08-03
CO-1925 simplify code and perform an extra combinations (all payments)

nbornand:


sponsorship_compassion/models/account_invoice.py
```diff
@@ -124,65 +124,53 @@ def _group_or_split_reconcile(self):
...
-                            matching_lines += open_payments[prev_index:][index]
-                            prev_index += index + 1
-                        break
-                if matching_lines:
-                    break
+            all_combinations = \
+                (combination for n in range(2, len(open_payments) + 1)
+                 for combination in itertools.combinations(open_payments, n))
+            for combination in all_combinations:
+                total = sum(payment.credit for payment in combination)
```
ecino: Even better : `total = sum(combination.mapped('credit'))`. This becomes more clear if `combination` is renamed to `payments_combination`

sponsorship_compassion/models/account_invoice.py
```diff
@@ -124,65 +124,53 @@ def _group_or_split_reconcile(self):
...
+                if total == reconcile_amount:
+                    for payment in combination:
```
ecino: Why not directly do `(combination | move_lines).reconcile()`

ecino (changes requested):
I think it's getting better, just small propositions.

sponsorship_compassion/models/account_invoice.py
nbornand: I guessed it was there for a reason. Don't both operands have to be RecordSets?

sponsorship_compassion/models/account_invoice.py
nbornand: not sure the mapped works as itertools.combinations yields tupples

sponsorship_compassion/models/account_invoice.py
```diff
@@ -150,13 +150,13 @@ def _group_or_split_reconcile(self):
...
                 (combination for n in range(2, len(open_payments) + 1)
                  for combination in itertools.combinations(open_payments, n))
-            for combination in all_combinations:
-                total = sum(payment.credit for payment in combination)
-                if total == reconcile_amount:
-                    for payment in combination:
+            for payment_combination in all_payment_combinations:
+                combination_amount = sum(p.credit for p in payment_combination)
+                if combination_amount == reconcile_amount:
+                    for payment in payment_combination:
```
ecino: You can avoid this loop by doing directly `(payment_combination | move_lines).reconcile()`

ecino (changes requested):

sponsorship_compassion/models/account_invoice.py
ecino: Ah you're right, I didn't see these are tuples.

sponsorship_compassion/models/account_invoice.py
ecino: It was my same mistaking, I didn't see combination is a tuple and not a recordset. So I think it's good now.

ecino (approved):

ecino:
@nbornand Did you try a similar case we looked at to see if behaviour still looks good?

PR #757, opened 2018-08-06
Devel to production - SMS sponsorship + various fixes

ecino:


partner_communication/models/report_with_omr.py
```diff
@@ -0,0 +1,23 @@
...
+
+class OmrAwareReport(models.Model):
+
+    _inherit = 'report'
+
+    @api.multi
+    def get_pdf(self, docids, report_name, html=None, data=None):
+        pdf_data = super(OmrAwareReport, self) \
+            .get_pdf(docids, report_name, html=html, data=data)
+        job = self.env['partner.communication.job'].browse(docids).exists()
```
ecino: @nbornand 
I didn't see this before, but reports can be for other models than partner.communication.job. With this you would risk to add OMR marks to another report if you find a partner.communication.job that has same ID as the record of another model on which we want to print a report.

We would need to add the check that the report is for a partner.communication.job object.

PR #758, opened 2018-08-06
Small fixes for OMR

ecino:


partner_communication/migrations/10.0.1.0.1/post-migration.py
```diff
@@ -16,21 +16,18 @@ def migrate(env, version):
...
             omr_add_attachment_tray_2
         ) SELECT
             id,
-            {},
             omr_enable_marks,
             omr_should_close_envelope,
             omr_add_attachment_tray_1,
             omr_add_attachment_tray_2
         FROM partner_communication_config;
-    """.format(lang.id))
```
nbornand, line 36: why don't you force the language? it there a value by default?

nbornand (approved):

PR #765, opened 2018-08-10
CO-1968 renamed function and removed ambassador tag

quentingigon:


crm_compassion/models/res_partner.py
```diff
@@ -31,7 +31,7 @@ def open_events(self):
         }
 
     @api.multi
-    def make_ambassador(self):
+    def create_odoo_user(self):
```
ecino, line 34: It's not the method in python that must be renamed but the action string which is somwhere in a xml file. But you can as well rename the code method, but then please update as well in the xml of the action, where this method is called, otherwise it will fail!

crm_compassion/models/res_partner.py
```diff
@@ -41,7 +41,6 @@ def make_ambassador(self):
             partner = user.partner_id
             user.email = partner.firstname[0].lower() + \
                 partner.lastname.lower() + '@cs.local'
-        portal.action_apply()
```
ecino: this is not the good fix. You should leave this and search somwhere in the compassion-switzerland repository where we add a 'category_id' to the partner. This is what we want to remove.

ecino (changes requested):

quentingigon:
I have modified a file in compassion-switzerland so I will do a PR on both repos

ecino (approved):

PR #766, opened 2018-08-13
CO-1996 better assignment of event to sms_child_request

quentingigon:


ecino (changes requested):
Not a good fix. The first version was better because now you always look at first for future events (that has not started yet)! However we are at first looking for an event that has already started.

=> We want to find an event that has started (start_date in the past) but start_date is not longer than 1 week. This should be done in one single search query.

sms_sponsorship/models/sms_child_request.py
```diff
@@ -83,10 +83,14 @@ def _compute_full_url(self):
     @api.depends('date')
     def _compute_event(self):
         for request in self.filtered('date'):
-            request.event_id = self.env['crm.event.compassion'].search([
+            event_id = self.env['crm.event.compassion'].search([
                 ('accepts_sms_booking', '=', True),
-                ('start_date', '<=', request.date)
-            ], limit=1)
+                ('start_date', '<=', request.date),
+                ('start_date', '>=', datetime.today() -
```
ecino: You should compare to a string and not to a date object, so here you should use the function `fields.Datetime.to_string(date)` to convert your datetime object.

ecino (changes requested):
You have some weird flake8 misalignments that you added in your commit. Please restore them.

ecino:
Not yet good:
```
sms_sponsorship/models/sms_child_request.py:106:23: E127 continuation line over-indented for visual indent
sms_sponsorship/models/sms_child_request.py:116:24: E131 continuation line unaligned for hanging indent
sms_sponsorship/models/sms_child_request.py:167:13: E125 continuation line with same indent as next logical line
sms_sponsorship/models/sms_child_request.py:180:31: E131 continuation line unaligned for hanging indent
sms_sponsorship/models/sms_child_request.py:213:24: E131 continuation line unaligned for hanging indent
```

sms_sponsorship/models/sms_child_request.py
```diff
@@ -83,10 +83,14 @@ def _compute_full_url(self):
     @api.depends('date')
     def _compute_event(self):
         for request in self.filtered('date'):
-            request.event_id = self.env['crm.event.compassion'].search([
+            event_id = self.env['crm.event.compassion'].search([
                 ('accepts_sms_booking', '=', True),
-                ('start_date', '<=', request.date)
-            ], limit=1)
+                ('start_date', '<=', fields.Datetime.to_string(request.date)),
```
ecino: Oops, this shouldn't be put inside to_string, because request.date is already a string!

sms_sponsorship/models/sms_child_request.py
```diff
@@ -83,10 +83,14 @@ def _compute_full_url(self):
...
+                ('start_date', '>=', fields.Datetime.to_string(
```
ecino: Oops, to_string must englobe the whole date, not only today().

ecino (changes requested):

sms_sponsorship/models/sms_child_request.py
quentingigon: In the fields of sms_child_request, date is a fields.Datetime object.. request is not a sms_child_request ?

PR #779, opened 2018-08-17
Co 1947 event duration

nbornand:


crm_compassion/tests/test_crm_compassion.py
```diff
@@ -107,7 +119,7 @@ def _create_event(self, lead, type):
                 'name': event_dico['context']['default_name'],
                 'type': type,
                 'start_date': now,
-                'end_date': now,
+                'end_date': now + ' 8:43:00',
```
ecino, line 128: For what is that for? I think we should replace both start_date and end_date by `fields.Datetime.now()`

crm_compassion/models/event_compassion.py
```diff
@@ -482,17 +483,16 @@ def _get_calendar_vals(self):
...
+                      fields.Datetime.from_string(self.start_date))
+        duration_in_hours = math.ceil(time_delta.days * 24 +
+                                      time_delta.seconds / 3600.0)
         calendar_vals = {
             'name': self.name,
             'compassion_event_id': self.id,
             'categ_ids': [
                 (6, 0, [self.env.ref('crm_compassion.calendar_event').id])],
-            'duration': number_of_days * 8,
+            'duration': duration_in_hours,
```
ecino: I think now we still have the problem that if start_date == end_date, the duration will be 0? Maybe we should put here `duration_in_hours or 3` to put a minimum of 3 hours duration for instance?

ecino (changes requested):

nbornand:
To end up being 0, not only the dates must be the same but the hours as well. Will this really happen?

ecino:
This makes it possible : https://github.com/CompassionCH/compassion-modules/blob/10.0/crm_compassion/models/event_compassion.py#L413

Also I checked on production, I found 112 events with start_date=end_date.

nbornand:
now the duration is at least 3h

ecino (approved):

PR #784, opened 2018-08-27
Co 2015 sms child reservation failure

nbornand:


ecino:
Can you tell me if/when this is ready to be reviewed?

nbornand:
Should be fine, I'm not sure why you added the "need fixes" yesterday

ecino:
Because of conflicting files. You could resolve them by rebasing your branch.

sbc_compassion/models/correspondence.py
```diff
@@ -276,17 +276,17 @@ def _set_partner_review(self):
     @api.depends('page_ids')
     def _compute_original_text(self):
         for letter in self:
-            letter.original_text = letter._get_text('original_text')
+            letter.original_text = letter._get_sms_message('original_text')
```
ecino, line 279: This is a bad search and replace in my opinion

sbc_compassion/models/correspondence.py
```diff
@@ -276,17 +276,17 @@ def _set_partner_review(self):
...
 
     @api.depends('page_ids')
     def _compute_translated_text(self):
         for letter in self:
-            letter.translated_text = letter._get_text('translated_text')
+            letter.translated_text = letter._get_sms_message('translated_text')
```
ecino, line 284: This is a bad search and replace in my opinion

sbc_compassion/models/correspondence.py
```diff
@@ -276,17 +276,17 @@ def _set_partner_review(self):
...
 
     @api.depends('page_ids')
     def _compute_english_translated_text(self):
         for letter in self:
-            letter.english_text = letter._get_text('english_translated_text')
+            letter.english_text = letter._get_sms_message('english_translated_text')
```
ecino, line 289: This is a bad search and replace in my opinion

ecino:
I think we will put this on hold as the process is likely to change based on what we discussed with David today, having first a set of children already available and only on high demand fetching new children by batches.

PR #794, opened 2018-09-03
[react] Fix error message on heavy load

ecino:


ecino:
Will merge from devel directly

PR #803, opened 2018-09-12
CO-2040 gifts of correspondent/partners generates same ref number

maxime-beck:


contract_compassion/models/contracts.py
```diff
@@ -222,10 +222,16 @@ def _compute_contract_duration(self):
     def create(self, vals):
         if 'commitment_number' not in vals:
             partner_id = vals.get('partner_id')
+            correspondent_id = vals.get('correspondent_id')
```
ecino: This should be done in module `sponsorship_compassion` where the field `correspondent_id` is defined.

ecino (changes requested):

sponsorship_compassion/models/contracts.py
```diff
@@ -249,7 +249,38 @@ def create(self, vals):
...
+            partner_id = vals.get('partner_id')
+            correspondent_id = vals.get('correspondent_id')
+            if partner_id:
+                other_nums = self.search([
+                    ('partner_id', '=', partner_id)
+                ]).mapped('commitment_number')
+
+                other_nums.extend(self.search([
+                    ('partner_id', '=', correspondent_id)
+                ]).mapped('commitment_number'))
```
ecino: You could have only one search if you put the following criterias:
```
'|', ('partner_id', '=', partner_id),
('partner_id', '=', correspondent_id)
```

ecino: Maybe you should as well count the contracts of the correspondent ? It would become : 
```
'|', '|', '|', ('partner_id', '=', partner_id),
('partner_id', '=', correspondent_id),
('correspondent_id', '=', partner_id),
('correspondent_id', '=', correspondent_id),
```

ecino (changes requested):

sponsorship_compassion/models/contracts.py
```diff
@@ -249,7 +249,37 @@ def create(self, vals):
                 'sponsor_id': vals.get('correspondent_id', vals['partner_id'])
             })
 
-        return super(SponsorshipContract, self).create(vals)
+        """ Generates commitment number for contracts BVRs
+        """
+        if 'commitment_number' not in vals:
+            partner_id = vals.get('partner_id')
+            correspondent_id = vals.get('correspondent_id')
```
ecino: correspondent_id = vals.get('correspondent_id', partner_id)
So that you take the partner in case the correspondent is not defined (could happen in non-sponsorship contracts)

contract_compassion/models/contracts.py
```diff
@@ -218,32 +218,6 @@ def _compute_contract_duration(self):
     ##########################################################################
     #                              ORM METHODS                               #
     ##########################################################################
-    @api.model
```
ecino: Actually you don't have to remove this code, it makes the tests fail. It won't do anything since you already set the commitment_number in sponsorship_compassion, and when you call super it will come here and do nothing because of the line 223 that tests if commitment_number is already set.

sponsorship_compassion/models/contracts.py
```diff
@@ -249,7 +249,37 @@ def create(self, vals):
...
+                    ('partner_id', '=', correspondent_id),
+                    ('correspondent_id', '=', partner_id),
+                    ('correspondent_id', '=', correspondent_id)
+                ]).mapped('commitment_number')
+
+                vals['commitment_number'] = max(other_nums or [-1]) + 1
+            else:
+                vals['commitment_number'] = 1
+
+        new_sponsorship = super(SponsorshipContract, self).create(vals)
```
ecino: You can directly return super at this point and remove the last part which will stay in contract_compassion module.

ecino (changes requested):
One flake8 error to fix:
sponsorship_compassion/models/contracts.py:284:5: E303 too many blank lines (2)

ecino (approved):

PR #806, opened 2018-09-19
CO-2058 better change child step1 

quentingigon:
WIP


quentingigon:
I couldn't reproduce the bug of the slow loading after many child changes. The rest should be OK

quentingigon:
Rest is not OK

quentingigon:
Working but there is an issue (opened in JIRA) when asking for a child in some countries (e.g. Indonesia) where there are more than one field office

PR #815, opened 2018-10-02
CO-2065 updated view and code to add child's preferred_name in partner view

quentingigon:


sponsorship_compassion/views/res_partner_view.xml
```diff
@@ -58,6 +58,7 @@
                             <tree colors="blue:state == 'draft';black:state == 'active';green: state == 'waiting';gray:state in ('terminated','cancelled')">
                                 <field name="child_id" />
                                 <field name="child_name" />
+                                <field name="preferred_name" />
```
ecino, line 62: You can add the field as well in line 42.

ecino (changes requested):
Just one place more to add the field in the view.

ecino (approved):

PR #817, opened 2018-10-02
CO-2060 BVR ref is filled at creation

quentingigon:


ecino:
This is not the good place to add this change. Everything should be in compassion-switzerland. If needed, this method `associate_group` can be inherited in module `sms_939`

PR #828, opened 2018-10-10
CO-2099 Fix id data xml

eicher31:


hr_attendance_management/data/hr_attendance_settings.xml
```diff
@@ -2,13 +2,13 @@
 <odoo>
     <data noupdate="1">
         <!-- Free break-->
-        <record id="hr_attendance.free_break" model="ir.config_parameter">
+        <record id="hr_attendance_management.free_break" model="ir.config_parameter">
```
ecino, line 5: Here you can only remove the module prefix since you are in the same module.

ecino (changes requested):

hr_attendance_management/data/hr_attendance_settings.xml
eicher31: Not here... we are in hr_switzerland

PR #836, opened 2018-10-17
CO-2114 Interventions corrections

quentingigon:


intervention_compassion/mappings/intervention_mapping.py
```diff
@@ -119,7 +120,6 @@ class InterventionMapping(OnrampMapping):
         "ActualCostPerBeneficiaryUSD": None,
         "EstimatedCostPerBeneficiaryUSD": None,
         "EstimatedLocalContribution": None,
-        "TotalActualCosts": None,
         "TotalEstimatedCost": None,
```
ecino: This should be mapped as well (it's as well a local currency cost)

intervention_compassion/models/compassion_intervention.py
```diff
@@ -87,6 +87,11 @@ class CompassionIntervention(models.Model):
     total_income = fields.Char(
         'Total income', compute='_compute_move_line', readonly=True)
     total_amendment = fields.Float()
+    total_actual_cost_local = fields.Float(help='Total costs in local '
+                                                'currency')
+    local_currency = fields.Many2one('res.currency',
```
ecino: name it  `local_currency_id`

intervention_compassion/models/compassion_intervention.py
```diff
@@ -592,7 +597,7 @@ def create_commitment(self):
             'res_model': 'compassion.intervention.commitment.wizard',
             'context': self.with_context({
                 'default_intervention_id': self.id,
-                'default_commitment_amount': self.hold_amount,
+                'default_commitment_amount': self.commitment_amount,
```
ecino: Here, the fix is this one: we want to make sure we don't commit more money than what the intervention costs that day in USD dollars. So we want to refresh the intervention first and commit maximum of intervention amount.
```suggestion
                'default_commitment_amount': max(self.hold_amount,
                                                                          self.estimated_costs),
```

intervention_compassion/models/global_intervention.py
```diff
@@ -53,10 +53,10 @@ class GenericIntervention(models.AbstractModel):
     # Budget Information (all monetary fields are in US dollars)
     ####################
     currency_usd = fields.Many2one('res.currency', compute='_compute_usd')
-    estimated_costs = fields.Float(readonly=True)
+    estimated_costs = fields.Float()
```
ecino: This should be readonly

intervention_compassion/models/global_intervention.py
```diff
@@ -53,10 +53,10 @@ class GenericIntervention(models.AbstractModel):
...
     remaining_amount_to_raise = fields.Float(readonly=True)
     pdc_costs = fields.Float(help='Program development costs', readonly=True)
-    total_cost = fields.Float(readonly=True)
```
ecino: This should be readonly

intervention_compassion/views/compassion_intervention_view.xml
```diff
@@ -90,6 +90,8 @@
                                     <field name="estimated_costs" widget="monetary" options="{'currency_field': 'currency_usd'}"/>
                                     <field name="estimated_impacted_beneficiaries"/>
                                     <field name="estimated_local_contribution" widget="monetary" options="{'currency_field': 'currency_usd'}"/>
+                                    <field name="local_currency"/>
```
ecino: You can make it `invisible="1"` since we already see the currency in the cost field.

intervention_compassion/views/compassion_intervention_view.xml
```diff
@@ -90,6 +90,8 @@
                                     <field name="estimated_costs" widget="monetary" options="{'currency_field': 'currency_usd'}"/>
```
ecino, line 90: Add the estimated costs in local currency as well

intervention_compassion/mappings/global_intervention_mapping.py
```diff
@@ -55,7 +55,7 @@ class GlobalInterventionMapping(OnrampMapping):
                              'compassion.global.partner'),
         "RemainingAmountToRaise": "remaining_amount_to_raise",
         "FundingStatus": "funding_status",
-        "PlannedSponsorshipBeneficiariesNumber":
+        "EstimatedImpactedBeneficiaryQuantity":
```
ecino: Remove this change, this is not good. You are in global_intervention_mapping which is not the same as intervention_mapping. In this message the fields are different. In fact I verified and the field is correctly mapped, it's just that GMC sends null values as soon as the intervention is commited.

intervention_compassion/models/compassion_intervention.py
```diff
@@ -592,7 +597,7 @@ def create_commitment(self):
             'res_model': 'compassion.intervention.commitment.wizard',
             'context': self.with_context({
                 'default_intervention_id': self.id,
-                'default_commitment_amount': self.hold_amount,
```
ecino, line 595: Add `self.get_infos()` before returning.

ecino (changes requested):

intervention_compassion/models/compassion_intervention.py
```diff
@@ -592,7 +600,8 @@ def create_commitment(self):
...
+                'default_commitment_amount': max(
```
ecino: No, it should be min instead of max.

ecino (changes requested):

ecino (approved):

PR #837, opened 2018-10-18
CO-2121 No more NaN in login widget

quentingigon:


hr_attendance_management/static/src/js/attendance.js
```diff
@@ -36,27 +35,27 @@ odoo.define('hr_attendance_management.attendance', function (require) {
                     });
 
                     // auto-counter
-                    $('#moment_pl').html(Date.now());
                     setInterval(function () {
+                        $('#moment_pl').html(Date.now());
```
ecino, line 39: I'm not sure, but is the timer correctly increased (meaning the worked hours are additioning if we stay on the page?)

quentingigon:
It's not working at all, I will close the PR for now

PR #838, opened 2018-10-19
CO-2118 Modify the use of Hr_location in Hr_Employee

nfluckiger:


hr_attendance_management/migrations/10.0.1.1.1/pre-migration.py
```diff
@@ -0,0 +1,23 @@
...
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    if not version:
+        return
+
+    openupgrade.rename_fields(env, [
+        ('hr_employee', 'hr_employee',
```
ecino: First parameter should be `hr.employee` which is the odoo model name.

hr_attendance_management/models/hr_employee.py
```diff
@@ -42,12 +42,24 @@ class HrEmployee(models.Model):
...
 
     ##########################################################################
     #                             FIELDS METHODS                             #
     ##########################################################################
 
+    @api.multi
+    def _compute_work_location(self):
+        for employee in self:
+            actual_location = self.env['hr.attendance'].search([
+                ('employee_id', '=', employee.id)])
```
ecino: Here you will loop over all attendances of the employee which is very inefficient. You could instead only take his last attendance by doing 
```
actual_location = self.env['hr.attendance'].search([
    ('employee_id', '=', employee.id),
    ('check_out', '=', False)], limit=1)
```
It works because the attendances are by default sorted from the latest to the oldest.

hr_attendance_management/models/hr_employee.py
```diff
@@ -42,12 +42,24 @@ class HrEmployee(models.Model):
...
+            for attendance in actual_location:
```
ecino: Here, simply `employee.work_location = attendance.location_id.name`

hr_attendance_management/views/hr_employee.xml
```diff
@@ -41,6 +41,9 @@
                     <field name="extra_hours" widget="float_time"/>
                 </div>
             </field>
+            <xpath expr="//field[@name='work_location']" position="replace">
```
ecino, line 44: Perfect in the form view of hr.employee, but you should as well replace it in the tree view and form view of hr.attendance

ecino (changes requested):

hr_attendance_management/__manifest__.py
```diff
@@ -38,7 +38,7 @@
         'hr_public_holidays',
         'hr_contract'
     ],
-    'version': '10.0.1.1.0',
+    'version': '10.0.1.1.1',
```
ecino: Stéphane just merged another change and pushed version to 1.1, so you should put 1.2

ecino (changes requested):

hr_attendance_management/models/hr_employee.py
```diff
@@ -42,12 +42,24 @@ class HrEmployee(models.Model):
...
     #                             FIELDS METHODS                             #
     ##########################################################################
 
+    @api.multi
+    def _compute_work_location(self):
+        for employee in self:
+            actual_location = self.env['hr.attendance'].search([
+                ('employee_id', '=', employee.id),
+                ('check_out', '=', False)], limit=1)
+            for attendance in actual_location:
```
ecino: You can remove this loop, because actual_location is a single record.

ecino (approved):

PR #844, opened 2018-10-23
CO-2111 Add field medical check in sponsorship impact

eicher31:


sponsorship_compassion/models/contracts_report.py
```diff
@@ -150,6 +152,19 @@ def get_nb_meal(sponsorship):
...
 
+    @api.multi
+    def _compute_medic_check(self):
+        def get_nb_check(sponsorship):
+            nb_year = sponsorship.contract_duration / 365
+            country = sponsorship.child_id.field_office_id
+            return nb_year * country.icp_medical_check
+
+        for partner in self:
+            total_check = sum(
```
ecino, line 163: You forgot to include the church members in the count.

ecino (changes requested):

sponsorship_compassion/models/contracts_report.py
eicher31: No, they are include in related_sponsorships

ecino (approved):

PR #845, opened 2018-10-24
CO-2112 migration script is working as intended

quentingigon:


ecino (changes requested):
I looked at the production data to see if your migration fits what we have and I suggest some modifications.
1. for the search criterias, you can change to this :
```
('state', '=', 'terminated'),
('global_id', '=', False),
('activation_date', '!=', False)
```
2. Then inside the loop we want to filter only those who have a contract duration greater than 5 days (less than 10 days could mean we never had time to transmit the sponsorship to US). For that you can simply use the computed field `contract_duration` and check `contract_duration > 5`

sponsorship_compassion/migrations/10.0.1.0.6/post-migration.py
```diff
@@ -0,0 +1,36 @@
...
+        ('state', 'ilike', 'terminated'),
+        ('global_id', '=', False)
+    ])
+
+    random_global_ids = range(0, 1000000)
+    random.shuffle(random_global_ids)
+
+    for sponsorship in old_sponsorships:
+        sponsorship.write({
+            'global_id': 'hold-' + str(random_global_ids.pop())
```
ecino: change `hold-` prefix to `missing-`

sponsorship_compassion/migrations/10.0.1.0.6/post-migration.py
```diff
@@ -0,0 +1,40 @@
...
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    if not version:
+        return
+
+    end_date_limit = dateutil.parser.parse("2016-12-13").date()
+
+    old_sponsorships = env['recurring.contract'].search([
+        ('end_date', '<=', fields.Date.to_string(end_date_limit)),
```
ecino: Replace this condition by `('type', 'like', 'S')` to filter only sponsorships. The end_date makes no sense to filter after reflection.

sponsorship_compassion/migrations/10.0.1.0.6/post-migration.py
```diff
@@ -0,0 +1,40 @@
...
+#    @author: Quentin Gigon <gigon.quentin@gmail.com>
+#
+#    The licence is in the file __manifest__.py
+#
+##############################################################################
+from odoo import fields
+from openupgradelib import openupgrade
+
+import random
+import dateutil.parser
```
ecino: This won't be needed anymore

sponsorship_compassion/migrations/10.0.1.0.6/post-migration.py
```diff
@@ -0,0 +1,40 @@
...
+        ('end_date', '<=', fields.Date.to_string(end_date_limit)),
+        ('state', '=', 'terminated'),
+        ('global_id', '=', False),
+        ('activation_date', '!=', False)
+    ])
+
+    random_global_ids = range(0, 1000000)
+    random.shuffle(random_global_ids)
+
+    min_transmition_delay = 5
```
ecino: 10 days is better. Rename var to `min_days_duration`

ecino (changes requested):

ecino (approved):

PR #855, opened 2018-10-30
CO-2047 skip children when making a search with criteria

quentingigon:


child_compassion/wizards/global_child_search.py
```diff
@@ -94,6 +94,8 @@ class GlobalChildSearch(models.TransientModel):
     skip = fields.Integer(size=4, default=50000)
     take = fields.Integer(default=80, size=4)
 
+    skip_value = fields.Integer()
```
ecino: This field is not needed. The context value is something else and doesn't require any field to exist, so you can remove it.

ecino (changes requested):

ecino (approved):

PR #857, opened 2018-11-01
CO-2090 Search on linked partner work now with archived partner

eicher31:


cms_form_compassion/forms/match_partner_form.py
```diff
@@ -122,6 +125,7 @@ def form_before_create_or_update(self, values, extra_values):
                 partner.write(source_vals)
             self.partner_id = partner.id
             values['partner_id'] = partner.id
+            partner.active = True
```
ecino: Instead of doing another write, you could use the write at line 125 to put it active here.

ecino (commented):
Small improvement to avoid two writes.

ecino:
You can as well add this logic in sms_sponsorship/models/sms_child_request.py line 119 in the create method

ecino (approved):

PR #870, opened 2018-11-21
CO-2131 Resupply event

quentingigon:


crm_compassion/models/weekly_demand.py
```diff
@@ -109,8 +109,14 @@ def _compute_demand_events(self):
                 resupply += event.number_allocate_children - \
                     event.planned_sponsorships
 
+            if resupply < 0:
+                if week.average_unsponsored_web >= abs(resupply):
+                    # should we add abs(value) ?
+                    week.average_unsponsored_web += abs(resupply)
```
ecino: No it's the contrary that we are looking for, we want to substract the number (we will have less resupply web).
You should as well define resupply_events in that case. `week.resupply_events = 0`

ecino: BTW I'm not sure we are allowed to define another field value in the compute method of a specific field. Did you test the result?

ecino (changes requested):

ecino:
I was not saying you must separate compute demand and compute resupply. I was wondering if it's working to set a non-computed field (in that case `week.average_unsponsored_web`) inside a compute method of another field (in that case `week.resupply_event`.

Please revert the join compute method but test that your computation is working.

crm_compassion/models/weekly_demand.py
quentingigon: > BTW I'm not sure we are allowed to define another field value in the compute method of a specific field. Did you test the result?

I did and it is not possible I think. The value is correct during the execution but the field value is never updated

ecino: This was what I was afraid of. So you should move your logic in the write method after event resupply is modified.

ecino:
Not a good fix. We need to discuss it.

PR #883, opened 2018-12-07
FIX extra hours computation

ecino:
I think I found a bug in extra hours computation.
@eicher31 Please review and make sure adding dependency of leave_ids is good.

### DIFF
--- a/hr_attendance_management/models/hr_attendance_day.py
+++ b/hr_attendance_management/models/hr_attendance_day.py
@@ -297,8 +297,9 @@ class HrAttendanceDay(models.Model):
     @api.multi
     @api.depends('paid_hours', 'due_hours', 'extra_hours_lost')
     def _compute_extra_hours(self):
+        sick_leave = self.env.ref('hr_holidays.holiday_status_sl')
         for att_day in self:
-            if att_day.leave_ids.holiday_status_id.name == 'Sick Leaves':
+            if sick_leave in att_day.leave_ids.mapped('holiday_status_id'):
                 att_day.extra_hours = 0
             else:
                 extra_hours = att_day.paid_hours - att_day.due_hours


hr_attendance_management/models/hr_attendance_day.py
```diff
@@ -295,10 +295,11 @@ def _compute_break_total(self):
                 att_day.break_ids.mapped('total_duration') or [0])
 
     @api.multi
-    @api.depends('paid_hours', 'due_hours', 'extra_hours_lost')
+    @api.depends('paid_hours', 'due_hours', 'extra_hours_lost', 'leave_ids')
     def _compute_extra_hours(self):
```
eicher31, line 299: due_hours already depends of leave_ids

hr_attendance_management/models/hr_attendance_day.py
```diff
@@ -295,10 +295,11 @@ def _compute_break_total(self):
...
+        sick_leave = self.env.ref('hr_holidays.holiday_status_sl')
         for att_day in self:
-            if att_day.leave_ids.holiday_status_id.name == 'Sick Leaves':
+            if sick_leave in att_day.leave_ids.mapped('holiday_status_id'):
```
eicher31, line 302: Or this:
```suggestion
            if 'Sick Leaves' in att_day.leave_ids.mapped(
                    'holiday_status_id.name'):
```

ecino: No this is less good because the name can change depending of the language of the user! It's always better to check equality at  the record value instead of names that can be translated!

hr_attendance_management/models/hr_attendance_day.py
ecino: Ok I remove it.

eicher31 (approved):

PR #885, opened 2018-12-12
CO-2230 payment mode is updated for partner when contract is validated

quentingigon:


ecino:
The above line does exactly the same as you are doing. So this problem was maybe older and is now working. This PR is not needed but thanks for having check.

PR #892, opened 2018-12-19
Co 2256 Update Mapping and selection field to match US message

nfluckiger:


ecino (approved):
Thanks! It looks like US have some typos in their field names :)

PR #900, opened 2019-01-15
CO-2280 New sponsorship campaign saved

quentingigon:


ecino (commented):
Have a look at this one, to see if it is causing the issue : 
https://github.com/compassionch/compassion-modules/blob/770ae448805b38326f82d5f5898fc02389f81b20/sponsorship_compassion/models/contracts.py#L895

I would prefer that you find the issue instead of providing a workaround.

quentingigon:
Can not reproduce

PR #904, opened 2019-01-24
CO-2129 : Rajouter un bouton dans impact parrainage qui ouvre les invoice line

c-meier:


sponsorship_compassion/models/contracts_report.py
```diff
@@ -226,3 +226,25 @@ def open_sponsorship_report(self):
...
+            'type': 'ir.actions.act_window',
+            'name': 'Donations details',
+            'res_model': 'account.invoice.line',
+            'views': [[False, 'list']],
+            'context': self.with_context(
+                search_default_group_product=1,
+                form_view_ref='sponsorship_compassion'
+                              '.view_invoice_line_partner_tree '
+            ).env.context,
+            "domain": [("partner_id", "=", self.id),
```
ecino: In this view, we group as well the statistics of all church members, to have the contribution of one church. Therefore, you can display the donations of all church member by adding the following filter domain:
```
'|', ("partner_id", "=", self.id), ("partner_id", "=", self.church_id.id),
```

ecino (changes requested):
Thank you for the changes. Only one small thing to add to allow displaying all donations of church members.

ecino (approved):

PR #917, opened 2019-02-08
CO-2269 Dissocier la communication Welcome Muskathlon d'un nouveau Portal User

jo-bloiz-ornot:
Wizard that will let you choose if you want to create a communication and let you choose which template (for now there is only Muskathlon template)


ecino (changes requested):
There is something wrong here. In this module, you cannot rely on what is done in `partner_communication_switzerland`, because this is the generic one that can be installed alone. 
As this module doesn't depend on `partner_communication`, you have no rights to use the communications objects. This will probably make travis tests fail.
So I think all the logic that you introduce here is not supposed to be in this module but should be moved in `partner_communication_switzerland`.

ecino (approved):

PR #928, opened 2019-02-19
CO-2174 Ajouter l'option d'envoi "e-mail uniquement"

jo-bloiz-ornot:
add forth option in communication preference

need double check on migration

### DIFF
--- a/partner_communication/migrations/10.0.1.0.2/post-migration.py
+++ b/partner_communication/migrations/10.0.1.0.2/post-migration.py
@@ -1,17 +1,15 @@
 # -*- coding: utf-8 -*-
 ##############################################################################
 #
-#    Copyright (C) 2018 Compassion CH (http://www.compassion.ch)
+#    Copyright (C) 2019 Compassion CH (http://www.compassion.ch)
 #    Releasing children from poverty in Jesus' name
-#    @author: Nicolas Bornand
+#    @author: Joel Vaucher
 #
 #    The licence is in the file __manifest__.py
 #
 ##############################################################################
-from openupgradelib import openupgrade
 
 
-@openupgrade.migrate(use_env=True)
 def migrate(env, version):
     if not version:
         return
@@ -34,7 +32,8 @@ def migrate(env, version):
     env.cr.execute("""
         UPDATE res_partner
         SET global_communication_delivery_preference='auto_digital_only'
-        WHERE email_only=true AND global_communication_delivery_preference!='none'
+        WHERE email_only=true AND
+        global_communication_delivery_preference!='none'
         """)
     env.cr.execute("""
         UPDATE res_partner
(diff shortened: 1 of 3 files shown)


partner_communication/migrations/10.0.1.0.2/post-migration.py
```diff
@@ -0,0 +1,54 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (C) 2018 Compassion CH (http://www.compassion.ch)
```
ecino: 2019

partner_communication/migrations/10.0.1.0.2/post-migration.py
```diff
@@ -0,0 +1,54 @@
...
+#    Releasing children from poverty in Jesus' name
+#    @author: Nicolas Bornand
```
ecino: Put your name

partner_communication/migrations/10.0.1.0.2/post-migration.py
```diff
@@ -0,0 +1,54 @@
...
+#
+#    The licence is in the file __manifest__.py
+#
+##############################################################################
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate(use_env=True)
```
ecino: Since you are never using the environment, you can remove this decorator, remove the openupgrade import, and the method signature will take cr as first parameter instead of env. Then you can simply use it in the code to do cr.execute(...)

partner_communication/models/communication_config.py
```diff
@@ -200,19 +210,30 @@ def get_inform_mode(self, partner):
...
-                'digital': 'physical' if not partner.email_only else 'none',
+                'digital': 'physical',
+                'digital_only': 'digital',
                 'both': 'physical',
             },
             'digital': {
                 'none': 'none',
                 'physical': 'physical' if self.print_if_not_email else 'none',
                 'digital': 'digital',
-                'both': 'both' if self.print_if_not_email else 'digital',
```
ecino: I think this rule was correct and doesn't need to be changed.

partner_communication/models/communication_config.py
```diff
@@ -200,19 +210,30 @@ def get_inform_mode(self, partner):
...
+                'digital': 'digital',
+                'digital_only': 'digital',
+                'both': 'digital',
+
             },
             'both': {
                 'none': 'none',
                 'physical': 'physical',
-                'digital': 'both' if not partner.email_only else 'digital',
+                'digital': 'digital',
```
ecino:
```suggestion
                'digital': 'both',
```

ecino (changes requested):

partner_communication/migrations/10.0.1.0.2/post-migration.py
jo-bloiz-ornot: replace the name or add ?

ecino: Replace, it's a new file.

ecino (approved):

ecino:
small flake8 issue : `partner_communication/migrations/10.0.1.0.2/post-migration.py:35:34: W291 trailing whitespace`

PR #929, opened 2019-02-20
CO-2347 messages of hold removals are now correctly handled 

quentingigon:
The changes in contract_compassion/models/contracts.py came during a rebase

### DIFF
--- a/message_center_compassion/models/gmc_message.py
+++ b/message_center_compassion/models/gmc_message.py
@@ -256,14 +256,21 @@ class GmcMessagePool(models.Model):
         and call the process_commkit method on the related object. """
         object_ids = list()
         action = self.mapped('action_id')
+        success = False
         for message in self:
             model_obj = self.env[action.model]
             commkit_data = [json.loads(message.content)]
-            object_ids.extend(
-                map(str,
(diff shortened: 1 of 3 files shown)


message_center_compassion/models/gmc_message.py
```diff
@@ -256,14 +256,21 @@ def _perform_incoming_action(self):
         and call the process_commkit method on the related object. """
         object_ids = list()
         action = self.mapped('action_id')
+        success = False
```
ecino: Not sure that all these changes are needed in this file. Doesn't it work without it ? `object_ids` won't evaluate to false if it is `[-1]`

ecino (changes requested):
Please remove the code in contract_compassion.

message_center_compassion/models/gmc_message.py
quentingigon: No it doesn't work without the code, there is an error when computing object_ids on line 268 because it tries to find a hold with id == -1

quentingigon: But maybe success is a bad choice of name for the variable

ecino (approved):

PR #931, opened 2019-02-20
 CO-2267 : Controller Wordpress qui crée des partenaires

c-meier:
Doit être fusionné en même temps que CompassionCH/compassion-switzerland#788

### DIFF
--- a/cms_form_compassion/models/match_partner.py
+++ b/cms_form_compassion/models/match_partner.py
@@ -39,19 +39,12 @@ class MatchPartner(models.AbstractModel):
         if partner_id:
             partner = partner_obj.browse(partner_id)
 
-        if not partner or len(partner) > 1:
-            partner = partner_obj.search([
-                ('email', '=ilike', infos['email']),
-                '|', ('active', '=', True), ('active', '=', False),
-            ])
-
-        if not partner or len(partner) > 1:
-            partner = partner_obj.search([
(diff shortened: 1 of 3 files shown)


cms_form_compassion/models/match_partner.py
```diff
@@ -0,0 +1,144 @@
...
+        for the creation of the partner.
+        """
+        valid = self._match_get_valid_create_fields()
+        create_infos = {}
+        for key, value in infos.iteritems():
+            if key in valid:
+                create_infos[key] = value
+
+        # Mark the partner to be validated
+        create_infos['state'] = 'pending'
```
ecino: One problem here, the `state` field is not present when using this module only. The field is declared in `partner_compassion` module and thus cannot be used here. It should be moved in a inheritance inside switzerland repository, for instance in `website_compassion` module, or in `partner_compassion`.

ecino (changes requested):
One small logic to be moved out from the module, otherwise looks good to me.

One suggestion about code disposition: I usually follow the principle to put public methods first and private methods at the end of the file. I try as well to declare methods that are used by other methods below, so that when you read the file you can read from top level and go into the details when you scroll down.

ecino:
You have one import that is no more used: `cms_form_compassion/models/match_partner.py:12:1: F401 'odoo._' imported but unused`

ecino (approved):

PR #932, opened 2019-02-20
CO-2340 Cadeau de parrainage waiting mandate

jo-bloiz-ornot:
add check on the recurent.contract is_active field

### DIFF
--- a/gift_compassion/models/contracts.py
+++ b/gift_compassion/models/contracts.py
@@ -47,9 +47,21 @@ class SponsorshipContract(models.Model):
             if invl.product_id.categ_name == GIFT_CATEGORY and \
                     invl.contract_id.child_id:
                 # Create the Sponsorship Gift
-                self.env['sponsorship.gift'].create_from_invoice_line(invl)
+                gift = self.env['sponsorship.gift']\
+                    .create_from_invoice_line(invl)
+
+                if not invl.invoice_id.is_active:
+                    gift.state = 'verify'
 
         super(SponsorshipContract, self).invoice_paid(invoice)
+
+    @api.onchange('is_active')
+    def onchange_is_active(self):
+        if self.is_active:
+            for invl in self.invoice_line_ids.filtered('gift_id'):
+                gift = invl.gift_id
+                if gift.state == 'verify':
+                    gift.state = 'draft'
 
     @api.multi
     def invoice_unpaid(self, invoice):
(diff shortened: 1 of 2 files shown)


gift_compassion/models/contracts.py
```diff
@@ -47,10 +47,22 @@ def invoice_paid(self, invoice):
             if invl.product_id.categ_name == GIFT_CATEGORY and \
                     invl.contract_id.child_id:
                 # Create the Sponsorship Gift
-                self.env['sponsorship.gift'].create_from_invoice_line(invl)
+                gift = self.env['sponsorship.gift']\
+                    .create_from_invoice_line(invl)
+
+                if not invl.invoice_id.is_active:
```
ecino:
```suggestion
                if not invl.contract_id.is_active:
```

gift_compassion/models/contracts.py
```diff
@@ -47,10 +47,22 @@ def invoice_paid(self, invoice):
...
+                    gift.state = 'verify'
 
         super(SponsorshipContract, self).invoice_paid(invoice)
 
+    @api.onchange('is_active')
```
ecino: You cannot use api.onchange method to listen to a change in another object. Onchange decorator is meant only to listen to changes in a form view ([see documentation here](https://www.odoo.com/documentation/10.0/reference/orm.html#module-odoo.api))

The correct way is to go in the contract object and in the method `contract_active`, searching for any pending gifts related to the contract and write them in state draft.

ecino (changes requested):

jo-bloiz-ornot:
TODO not merge before asking david about the is_active or has_mandate problem

gift_compassion/models/contracts.py
```diff
@@ -47,10 +47,26 @@ def invoice_paid(self, invoice):
...
+                    .create_from_invoice_line(invl)
+
+                if not invl.contract_id.is_active:
+                    gift.state = 'verify'
 
         super(SponsorshipContract, self).invoice_paid(invoice)
 
+    @api.multi
+    @api.depends('activation_date', 'state')
+    def _compute_active(self):
```
ecino: No, it's still not good, because _compute_active is a compute method which purpose is to set a value of a field. You cannot change other values from here, it's a dangerous behaviour that can do unpredictable results. You should instead inherit the `contract_active` [method found here](https://github.com/compassionch/compassion-modules/blob/770ae448805b38326f82d5f5898fc02389f81b20/sponsorship_compassion/models/contracts.py#L673).

jo-bloiz-ornot:
https://github.com/CompassionCH/compassion-switzerland/pull/789
Linked pull request

ecino (approved):

PR #941, opened 2019-02-25
CO-2221 : Faire un CRON pour notifier le staff 1x par semaine s'il y'a des demandes SMS

c-meier:


sms_sponsorship/models/sms_child_request.py
```diff
@@ -337,3 +345,30 @@ def sms_reminder_cron(self):
...
+
+    @api.model
+    def sms_notification_unfinished_cron(self):
+        """
+        CRON job that sends mails weekly, which notify the staff that some SMS
+        Sponsorship are ongoing.
+        :return: True
+        """
+        nb_sms_requests = self.search_count([
+            ('date', '<', fields.Date.today()),
```
ecino: I would as well include the request of the current day

sms_sponsorship/models/sms_child_request.py
```diff
@@ -337,3 +345,30 @@ def sms_reminder_cron(self):
...
+            ('state', 'in', ['new', 'child_reserved', 'step1']),
+        ])
+
+        # send staff notification
+        notify_ids = self.env['staff.notification.settings'].get_param(
+            'new_partner_notify_ids')
+        if notify_ids:
```
ecino:
```suggestion
        if nb_sms_requests and notify_ids:
```
To avoid sending a notification when there are no requests.

ecino (changes requested):

ecino (approved):

PR #945, opened 2019-02-27
CO-2397 : Use match partner for sms sponsorship

c-meier:
As the field new_partner was removed from sms.child.request, all new sms sponsorship must now be manually confirmed.


ecino (approved):
Looks great, thanks!

PR #948, opened 2019-03-01
CO-2398 Select template automatically when reply

eicher31:


crm_request/models/request.py
```diff
@@ -13,26 +13,76 @@ class CrmClaim(models.Model):
     name = fields.Char(string='Subject')
     code = fields.Char(string='Number')
     claim_type = fields.Many2one(string='Type')
+    user_id = fields.Many2one(string='Assign to')
+    description = fields.Html()
```
ecino: Would be better to change the widget in the view, by adding the attribute `widget="html"`, instead of changing the field type which could break functionality with other modules.

crm_request/models/request.py
```diff
@@ -13,26 +13,76 @@ class CrmClaim(models.Model):
     name = fields.Char(string='Subject')
     code = fields.Char(string='Number')
     claim_type = fields.Many2one(string='Type')
+    user_id = fields.Many2one(string='Assign to')
```
ecino, line 16: Would be better to change the string attribute in the view. Not sure if Assign to is better than 'Responsible'. It's just a matter of habit.

crm_request/models/request.py
```diff
@@ -13,26 +13,76 @@ class CrmClaim(models.Model):
...
+
+    @api.multi
+    def action_reply(self):
+        """
+        This function opens a window to compose an email, with the default
+        template message loaded by default"""
+        self.ensure_one()
+        ir_model_data = self.env['ir.model.data']
+        try:
+            template_id = self.claim_type.template_id.id
```
ecino: This will never raise a Value Error. You can remove the try block.

crm_request/models/request.py
```diff
@@ -13,26 +13,76 @@ class CrmClaim(models.Model):
...
+        except ValueError:
+            template_id = False
+        try:
+            compose_form_id = ir_model_data.get_object_reference(
```
ecino: You should better use the `ref` function. Don't assume errors if you know the object exists:
`compose_form_id = self.env.ref('mail.email_compose_message_wizard_form').id`

crm_request/models/request.py
```diff
@@ -13,26 +13,76 @@ class CrmClaim(models.Model):
...
+            'default_composition_mode': 'comment',
+            'mark_so_as_sent': True,
+        })
+        return {
+            'type': 'ir.actions.act_window',
+            'view_type': 'form',
+            'view_mode': 'form',
+            'res_model': 'mail.compose.message',
+            'views': [(compose_form_id, 'form')],
+            'view_id': compose_form_id,
```
ecino: In fact you don't need to specify the view id if you just need the default view. So I would advise to just remove this in order to clean a bit the code.

crm_request/models/request.py
```diff
@@ -13,26 +13,76 @@ class CrmClaim(models.Model):
...
+            template_id = self.claim_type.template_id.id
+        except ValueError:
+            template_id = False
+        try:
+            compose_form_id = ir_model_data.get_object_reference(
+                'mail', 'email_compose_message_wizard_form')[
+                1]
+        except ValueError:
+            compose_form_id = False
+        ctx = dict()
```
ecino: Why not directly initialize the dict like this ?
```
ctx = {
    'default_model': ...,
    ....
}
```

crm_request/models/request.py
```diff
@@ -13,26 +13,76 @@ class CrmClaim(models.Model):
...
+        ctx.update({
+            'default_model': 'crm.claim',
+            'default_res_id': self.ids[0],
```
ecino: `self.id` instead of `self.ids[0]`

crm_request/models/request.py
```diff
@@ -13,26 +13,76 @@ class CrmClaim(models.Model):
...
+            'default_template_id': template_id,
+            'default_composition_mode': 'comment',
+            'mark_so_as_sent': True,
+        })
+        return {
+            'type': 'ir.actions.act_window',
+            'view_type': 'form',
+            'view_mode': 'form',
+            'res_model': 'mail.compose.message',
+            'views': [(compose_form_id, 'form')],
```
ecino: No need to add this. The default view will be taken and it will be enough.

ecino (changes requested):

ecino (approved):

PR #949, opened 2019-03-03
[MOV] move module hr_attendance_management in the repo CompassionCH/hr

jo-bloiz-ornot:
Must be apply only after OCA/hr validate the pull request of
CompassionCH about this new module

https://github.com/CompassionCH/hr/pull/2


jo-bloiz-ornot:
Tache CO-2287

ecino:
We need to know how to do the migration at production without losing data. @eicher31 can you handle that?

eicher31:
Yes, I will check that with @joblo-compassion.

PR #952, opened 2019-03-05
CO-2414 Add mail source to model request

eicher31:


crm_request/models/request.py
```diff
@@ -49,7 +52,12 @@ def message_new(self, msg, custom_values=None):
         """
         if custom_values is None:
             custom_values = {}
-        custom_values.update({'description': msg.get('body')})
+        txt = msg.get('to')
+        alias_char = txt[txt.find('"') + 1:txt.find("@")]
```
ecino: That could possibly fail if the `to` field contains another format or unexpected destination. In that case all operation will be aborted which could lead to losing information.
```suggestion
        alias_char = txt.split("@")[0].replace('"', '')
```

crm_request/views/request.xml
```diff
@@ -18,6 +18,10 @@
                 <attribute name="string">Description</attribute>
             </xpath>
 
+            <field name="team_id" position="after">
+                <field name="alias_id" groups="base.group_no_one"/>
```
ecino, line 94: Why adding group in a specific field? I would just let the field security be the same as the whole view no?

ecino (changes requested):

ecino:
Conflicts with changes from Christopher

crm_request/views/request.xml
eicher31: I find that is a technical feature that pollute the view for the user.

crm_request/models/request.py
```diff
@@ -55,9 +59,15 @@ def message_new(self, msg, custom_values=None):
         """
         if custom_values is None:
             custom_values = {}
+
+        alias_char = parseaddr(msg.get('to'))[1].split('@')[0]
```
eicher31, line 76: @ecino Even better... ;-)

ecino:
Conflicts to resolve.

ecino:
Please resolve this...

ecino (approved):

PR #953, opened 2019-03-05
WP-20 : Use the date of the sending of the message instead of the date of the fetching.

c-meier:


ecino (approved):

ecino:
Wrong commit name, shouldn't be associated to WP-20. Can you just edit that?

c-meier:
Which issue should it be associated with ?

ecino:
I think it's me that got confused, sorry.

PR #954, opened 2019-03-06
CO-2415 : Unlink partner from support request email

c-meier:


crm_request/models/request.py
```diff
@@ -49,8 +55,20 @@ def message_new(self, msg, custom_values=None):
...
             custom_values = {}
-        custom_values.update({'description': msg.get('body')})
-        return super(CrmClaim, self).message_new(msg, custom_values)
+
+        defaults = {
+            'description': msg.get('body')
+        }
+
+        if 'partner_id' not in custom_values:
+            match_obj = self.env['res.partner.match.wp']
```
ecino: Why do you use res.partner.wp instead of res.partner.match? wordpress module is not dependent of cmr_compassion.

ecino (changes requested):

crm_request/models/request.py
c-meier: The remains from a copy-paste. Corrected

ecino:
Thanks for the correction. There are code conflicts with WP-20 that I just merged. Can you fix that?

ecino (approved):

PR #961, opened 2019-03-12
CO-2425 Remove previous followers for change responsable

jonathantarabbia:


crm_request/models/request.py
```diff
@@ -101,3 +101,15 @@ def message_post(self, **kwargs):
                     'crm_request', 'stage_wait_customer')[1]
 
         return result
+
+    @api.multi
+    def write(self, values):
+        # Get all followers before change
```
eicher31, line 107: You should first check if it is in the "assign to" action.

for record in self:
    if values.get('user_id')  and record.user_id is None:
        ...

crm_request/models/request.py
```diff
@@ -101,3 +101,15 @@ def message_post(self, **kwargs):
...
+        super(CrmClaim, self)._get_followers
+
+        #  unsubscribe all followers before change
+        super(CrmClaim, self).message_unsubscribe(
+            partner_ids=[self.message_partner_ids.id])
```
eicher31, line 112: Unsubscribe only the current user
record.message_unsubscribe([self.env.uid])

crm_request/models/request.py
```diff
@@ -101,3 +101,15 @@ def message_post(self, **kwargs):
...
+
+        # launching the standard process "write"
+        return super(CrmClaim, self).write(values)
```
eicher31, line 115: You should start by the "super", thus you avoid to maybe add again the current user as follower.

eicher31 (changes requested):

PR #966, opened 2019-03-14
CO-2424 : improve interaction resume

c-meier:


ecino:
Travis error: 
```
ParseError: "External ID not found in the system: mail_tracking.view_partner_form" while parsing /home/travis/build/CompassionCH/compassion-modules/crm_compassion/views/res_partner_view.xml:49, near
<record id="view_res_partner_crm_remove_tracking" model="ir.ui.view">
        <field name="name">res.partner.crm.compassion.remove.tracking</field>
        <field name="model">res.partner</field>
        <field name="inherit_id" ref="mail_tracking.view_partner_form"/>
        <field eval="18" name="priority"/>
            <button name="%(mail_tracking.action_view_mail_tracking_email)d" position="replace"/>
        </field>
    </record>
```
I think you should just add `mail_tracking` to the dependencies of the manifest since you inherit one view of the module.

ecino (changes requested):
For the deletion of the buttons, I think it's better to add the attribute invisible, for instance:
```
<button name="%(crm_claim.crm_claim_category_claim0)d" position="attributes">
    <attribute name="invisible">1</attribute>
</button>
```
This way we would avoid possible crashes if some other modules would inherit and place elements according to this button, or simply reference it. Just test if it works the same as replacing it.

ecino:
Travis still failing. It is not properly configured. You can add the dependency to repository `social` inside the file `oca_dependencies.txt` and it should work.

ecino (approved):

PR #972, opened 2019-03-20
WP-29 : Tool to authenticate REST requests to Wordpress.

c-meier:
For Wordpress:

- The plugin [wp-api-jwt-auth](https://wordpress.org/plugins/jwt-authentication-for-wp-rest-api/) must be installed
- The following config options must be set in `wp-config.php` (before the DEBUG define) :
```
define('JWT_AUTH_SECRET_KEY', 'your-top-secret-key-that-must-be-changed');
define('JWT_AUTH_CORS_ENABLE', true);
```
- The following instruction must be set in `.htaccess` :
```
SetEnvIf Authorization "(.*)" HTTP_AUTHORIZATION=$1
```

### DIFF
--- a/mobile_app_connector/models/wordpress_post.py
+++ b/mobile_app_connector/models/wordpress_post.py
@@ -8,7 +8,8 @@
 #
 ##############################################################################
 import logging
-import requests
+from ..tools import wp_requests
+
 
 from odoo import api, models, fields, _
 from odoo.exceptions import UserError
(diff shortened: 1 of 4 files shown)


ecino:
`JWT_AUTH_SECRET_KEY` is never needed in the Odoo code? For what is it used then?

c-meier:
It is used by wordpress to sign the token it sends, so that when wp receive a request with a token it can be certain that he was the one to create it.

ecino (approved):

PR #976, opened 2019-03-21
CO-2417 Auto detect type of demand

eicher31:


crm_request/models/request_type.py
```diff
@@ -13,3 +13,27 @@ class RequestType(models.Model):
...
                                   domain="[('model_id', '=', 'crm.claim')]")
+    keywords = fields.Char(
+        string='Keywords',
+        help='List of keywords (separated by a comma ",") who could be '
+             'contained in the demand subject')
+
+    @api.constrains('keywords')
+    def _check_existing_key(self):
+        # Avoid having two identical keys
+        records = self.search([('keywords', '!=', False)])
```
c-meier, line 24: You could go from O(n^2) to average O(n) by getting all the keyword in one go, creating a set from the keywords and checking if the length is the same.

crm_request/models/request.py
```diff
@@ -103,10 +103,21 @@ def message_new(self, msg, custom_values=None):
...
 
+        # Find the corresponding type
+        subject = msg.get('subject')
+        type_ids = self.env['crm.claim.type'].search(
+            [('keywords', '!=', False)])
+        type_id = False
+        for record in type_ids:
+            if any(word in subject for word in record.get_keys()):
+                type_id = record.id
+                break
```
c-meier, line 114: Would it be worth it to make a statistical analysis of the keywords ? For exemple: There was 3 word in the subject that matched type1 and 1 word that matched type 2 so the type that is set is type1 ...

c-meier (approved):

ecino (approved):

PR #981, opened 2019-03-22
CO-2451 Add action to assign

eicher31:


crm_request/models/request.py
```diff
@@ -172,3 +173,21 @@ def write(self, values):
...
+
+    @api.multi
+    def assign_to(self):
+        self.ensure_one()
+        model = self.env.context.get('active_model')
+        model_id = self.env.context.get('active_id')
+        request = self.env[model].browse(model_id)
+        request.user_id = self.user_id
+        if self.intern_note:
+            request.message_post(subject=u'Message pour ' + self.user_id.name,
```
ecino: Never put french text in code. It should be English wrapped in the translate function _()

crm_request/models/request.py
```diff
@@ -17,8 +17,9 @@ class CrmClaim(models.Model):
         help="The destination email address that the contacts used.")
     code = fields.Char(string='Number')
     claim_type = fields.Many2one(string='Type')
-    user_id = fields.Many2one(string='Assign to')
-    stage_id = fields.Many2one(group_expand='_read_group_stage_ids')
+    user_id = fields.Many2one(string='Assign to', track_visibility=False)
```
ecino: Why did you remove the track_visibility attributes? It's a bit strange to remove this, as it won't notify anyone if we change the assignee directly from the field. Maybe then you should put the field readonly in the view to make sure nobody changes it without going through your wizard.

crm_request/models/request.py
```diff
@@ -17,8 +17,9 @@ class CrmClaim(models.Model):
...
+    stage_id = fields.Many2one(group_expand='_read_group_stage_ids',
+                               track_visibility=False)
```
ecino: I don't understand why your remove this. It could be useful to be notified when the stage of request changed.

ecino (changes requested):

ecino:
LINT check failed.

ecino (approved):

PR #989, opened 2019-03-28
CO-2470 Change workflow for request

eicher31:


ecino (commented):
I think we miss the case if the supporter replies, that it should bring back the claim to the 'Waiting on support' stage (and notify the assigned user).

ecino:
Just waiting to fix travis issues (in my pull request) and then we can relaunch the build here.

ecino (approved):

PR #990, opened 2019-03-28
add message to tree view interaction_resume

jonathantarabbia:
- update text_collapse_html widget for formatted html on tree view

### DIFF
--- a/crm_compassion/views/interaction_resume_view.xml
+++ b/crm_compassion/views/interaction_resume_view.xml
@@ -13,6 +13,8 @@
                 <field name="email"/>
                 <field name="subject"/>
                 <field name="tracking_status"/>
+                <field name="subject"/>
+                <field name="body"/>
             </tree>
         </field>
     </record>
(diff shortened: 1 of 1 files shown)


crm_compassion/views/interaction_resume_view.xml
```diff
@@ -13,6 +13,8 @@
                 <field name="email"/>
                 <field name="subject"/>
                 <field name="tracking_status"/>
+                <field name="subject"/>
```
ecino, line 15: subject is already present two lines above. Please keep only one of them.

crm_compassion/views/interaction_resume_view.xml
```diff
@@ -57,6 +59,7 @@
                 <group expand="0" string="Group By...">
                     <filter name="Type" domain="[]" context="{'group_by':'communication_type'}"/>
                     <filter name="Direction" domain="[]" context="{'group_by':'direction'}"/>
+                    <filter name="communication_date" domain="[]" context="{'group_by':'communication_date'}"/>
```
ecino: I just noticed that filter `name` attribute is a technical name and you should add a `string` attribute with human readable name (for the user interface). Can you as well add the `string` attribute in the other filters?

ecino (changes requested):

ecino (approved):

PR #997, opened 2019-04-03
CO-2503 Remove the auto assign

eicher31:
Remove the auto assign when the request move in "Waiting on support".


ecino:
Is that a request from the employees?

eicher31:
It's a fix, to avoid to replace the current assigned user.

ecino (approved):

PR #1003, opened 2019-04-08
Fix webservices app connector

jonathantarabbia:
- Fix get tiles when Offline
 - Fix get tiles when logged
 - Fix get Children when tile.mode is many
 - Fix Mapping of child object
 - Add Child pictures Object and Mapping
 - Fix Tile CH2 webservices


mobile_app_connector/controllers/mobile_app_controller.py
```diff
@@ -92,7 +92,7 @@ def get_hero(self, hero_type=None, view=None, **parameters):
                  note: only one item is used by the app.
         """
         hero = request.env['mobile.app.banner'].search([
-            ('type', 'ilike', hero_type),
+            # ('type', 'ilike', hero_type),
```
ecino: Is the app not providing the hero_type ? Why did you remove this ? It means we won't be able to display different banners depending if the user is logged in or not.

mobile_app_connector/models/app_hub.py
```diff
@@ -124,7 +130,8 @@ def _fetch_wordpress_tiles(self, **pagination):
         return messages
 
     def _construct_hub_message(self, partner_id, messages, children=None,
-                               start=0, limit=100, **kwargs):
+                               children_pictures=None, start=0, limit=100,
```
ecino: You added this parameter, but it's never used in the method?

ecino (changes requested):
You should rebase onto compassion/devel and remove your commit AP-47, because it's already merged.

mobile_app_connector/models/app_hub.py
jonathantarabbia: indeed, it is not necessary

mobile_app_connector/controllers/mobile_app_controller.py
ecino: I checked and iOS neither gives this parameter. Therefore I propose that you remove the parameter `hero_type` from the function and that you remove as well the field `type` in the object `mobile.app.banner`, as it won't be useful if it's not supported by the app. We can therefore only define one banner that is active for both public and private.

mobile_app_connector/models/compassion_child.py
```diff
@@ -29,17 +29,21 @@ def get_app_json(self, multi=False):
...
             data = mapping.get_connect_data(self)
+            # data['Images'] = children_pictures.get_app_json(multi=True)
         else:
             data = []
             for child in self:
                 data.append(mapping.get_connect_data(child))
-        return {wrapper: data}
+        return {wrapper: data,
+                'Images': [children_pictures.filtered(lambda r: r.image_url)
+                               .get_app_json(multi=True)]}
```
ecino, line 46:
```suggestion
                           .get_app_json(multi=True)]}
```

mobile_app_connector/mappings/compassion_child_pictures_mapping.py
```diff
@@ -0,0 +1,25 @@
...
+    ODOO_MODEL = 'compassion.child.pictures'
+    MAPPING_NAME = 'mobile_app_child_pictures'
+
+    CONNECT_MAPPING = {
+        'Date': 'date',
+        'Url': 'image_url'
+    }
+
+    FIELDS_TO_SUBMIT = {k: None for k, v in CONNECT_MAPPING.iteritems() if v}
+
```
ecino, line 25:
```suggestion

```
(make sure you have just a line return without any space or content on the last line)
https://github.com/PyCQA/pycodestyle/issues/365

ecino (changes requested):

PR #1004, opened 2019-04-08
Fix give gift tile

jonathantarabbia:


ecino:
I will check this one after we merge #1003

ecino:
Will merge in #1011

PR #1005, opened 2019-04-09
CO-2504 Removing gender info for some titles.

vwvw:


sponsorship_compassion/mappings/res_partner_mapping.py
```diff
@@ -32,7 +32,9 @@ class ResPartnerMapping(OnrampMapping):
     }
 
     FIELDS_TO_SUBMIT = {
-        'Gender': lambda gender: 'Female' if gender == 'F' else 'Male',
+        'Gender': lambda title:
+            title.gender if title in
```
ecino:
```suggestion
            title.gender if title.with_context(lang='en_US').name in
```
I think title is a record so title.name should be used to check your inclusion. I also included the language force to make sure we are comparing the english names.

sponsorship_compassion/migrations/10.0.1.0.7/pre-migration.py
```diff
@@ -0,0 +1,23 @@
...
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    if not version:
+        return
+
+    for partner in env['res.partner'].search(
+            ['partner.title.gender', '!=', False]):
```
ecino:
```suggestion
            [('partner.title.gender', '!=', False), ('global_id', '!=', False)]):
```
To make sure we loop through registered partners on GMC side (global id is given by GMC).

sponsorship_compassion/migrations/10.0.1.0.7/pre-migration.py
```diff
@@ -0,0 +1,23 @@
...
+        if partner.title.name not in ['Madam', 'Mister', 'Misters', 'Ladies']:
```
ecino: This condition could be added in the search criterias to avoid making a big loop and testing each time. Then you can as well simply call `upsert_constituent` on the recordset, avoiding to have to make a loop.

ecino (changes requested):

ecino (approved):

ecino:
I will fix the flake8 issues afterwards (with other that are in the repository)

PR #1008, opened 2019-04-09
Reply append body

jonathantarabbia:


ecino:
In order to avoid including all unmerged commits when you start a new task, you can create a branch starting from compassionch/devel. It means you should checkout compassionch/devel as a new branch instead of starting from your previous working branch.

Now in order to fix it, you can rebase your branch reply-append-body onto compassionch/devel, pick only the commit CO-2418 and force push.

ecino:
Improved in #1015

PR #1009, opened 2019-04-10
CO-2508 added the "has attachment" field

Alaggaerion:


crm_compassion/models/interaction_resume.py
```diff
@@ -162,3 +163,14 @@ def open_object(self):
...
         self.ensure_one()
         return True
+
+    def _compute_has_attachment(self):
+        """
+        Check in each row if e-mail (outgoing) or messages (in) contain at
+        least 1 attachment.
+        """
+        for row in self:
+            if row.email_id.attachment_ids:
```
ecino: Here you can simply put it like this:
```
            row.has_attachment = row.email_id.attachment_ids or \
                row.message_id.attachment_ids
```

ecino (changes requested):
Nice code overall. I like that you put a comment in the function, it's a good practice. One suggestion to make the code shorter and more comprehensible.

crm_compassion/models/interaction_resume.py
```diff
@@ -162,3 +163,12 @@ def open_object(self):
...
         return True
+
+    def _compute_has_attachment(self):
+        """
+        Check in each row if e-mail (outgoing) or message (incoming) contains
+        at least 1 attachment.
+        """
+        for row in self:
+            row.has_attachment = row.email_id.attachment_ids or \
+                                 row.message_id.attachment_ids
```
ecino:
```suggestion
                row.message_id.attachment_ids
```
Only one more tab indent is needed after \

ecino (approved):

PR #1010, opened 2019-04-10
CO-2526 Added a field to see  attachments in mail_message's form view

Alaggaerion:


crm_compassion/__manifest__.py
```diff
@@ -62,6 +62,7 @@
         'views/demand_weekly_revision.xml',
         'views/hold_view.xml',
         'views/calendar_view.xml',
+        'views/email_view.xml',
```
ecino: Naming the file `email_view` is confusing because the related object is not email but mail_message. So I would suggest that you name the file `mail_message_view.xml` instead. Other than that everything looks good!

ecino (changes requested):

ecino (approved):

PR #1017, opened 2019-04-12
CO-2511 Field getting days since last letter

Alaggaerion:
Added an invisible field to compute how many days passed since the last letter from the sponsor to the child
Prod templates still have to be updated in case the sponsor recently wrote to the child

### DIFF
--- a/contract_compassion/models/contracts.py
+++ b/contract_compassion/models/contracts.py
@@ -232,20 +232,19 @@ class RecurringContract(models.Model):
             try:
                 # Try to get days difference between today and last letter
                 contract.last_letter = (
-                        date.today() - fields.Datetime.from_string(
-                            contract.env['correspondence'].search([
-                                ('direction', '=', 'Supporter To Beneficiary'),
-                                ('sponsorship_id', '=', contract.id)],
-                                limit=1).scanned_date).date()).days
+                    date.today() - fields.Datetime.from_string(
+                        contract.env['correspondence'].search([
+                            ('direction', '=', 'Supporter To Beneficiary'),
+                            ('sponsorship_id', '=', contract.id)],
+                            limit=1).scanned_date).date()).days
             except AttributeError:
                 try:
                     # Try to get days difference between today and create_date
                     contract.last_letter = (
-                            date.today() - fields.Datetime.from_string(
-                                contract.create_date).date()).days
+                        date.today() - fields.Datetime.from_string(
+                            contract.create_date).date()).days
                 except AttributeError:
                     contract.last_letter = 0
-
 
     ##########################################################################
     #                              ORM METHODS                               #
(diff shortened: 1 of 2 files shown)


contract_compassion/views/contract_view.xml
```diff
@@ -83,6 +83,7 @@
                         <field name="group_freq"/>
                         <field name="months_paid"/>
                         <field name="last_paid_invoice_date"/>
+                        <field name="last_letter" invisible="1"/>
```
ecino: It's not needed to add it in the view if you keep it invisible. Either remove it or keep it visible.

contract_compassion/models/contracts.py
```diff
@@ -89,6 +89,9 @@ class RecurringContract(models.Model):
     contract_duration = fields.Integer(compute='_compute_contract_duration',
                                        help='Contract duration in days')
 
+    # Field used to compute days number since last sponsor's letter
+    last_letter = fields.Integer(compute='_compute_last_letter')
```
ecino: add a `help` parameter describing what the field is instead of using a comment above.

contract_compassion/models/contracts.py
```diff
@@ -215,6 +218,18 @@ def _compute_contract_duration(self):
                 contract.contract_duration = (
                     end_date - contract_start_date).days
 
+    def _compute_last_letter(self):
+        """
+        Get the date of the last letter sent to the child by the sponsor and
+        compute since how many days he didn't write him any letter
+        :return: int value
+        """
+        self.last_letter = (date.today() - datetime.strptime(
```
ecino: The compute method should always be able to be invoked on a recordset. This means that you should not compute only for one record but put a for loop and setting the values on each record:
```
for contract in self:
    contract.last_letter = ....
```

ecino: Instead of using `datetime.strptime`, use `fields.Datetime.from_string()` method

contract_compassion/models/contracts.py
```diff
@@ -215,6 +218,18 @@ def _compute_contract_duration(self):
...
+            self.env['correspondence'].search([(
+                'direction', '=', 'Supporter To Beneficiary'), (
+                'sponsorship_id', '=', self.id)], limit=1).scanned_date,
```
ecino: Did you think of the case if there are no letters? It should not fail but return the number of days since the creation of the sponsorship (using the field create_date)

ecino (changes requested):

ecino:
flake8 issues:
```
contract_compassion/models/contracts.py:235:25: E126 continuation line over-indented for hanging indent
contract_compassion/models/contracts.py:244:29: E126 continuation line over-indented for hanging indent
contract_compassion/models/contracts.py:250:5: E303 too many blank lines (2)
```

ecino (approved):

PR #1018, opened 2019-04-15
Fix webservices

jonathantarabbia:

PR #1019, opened 2019-04-16
Mobileapp post agenda download

jonathantarabbia:


ecino:
I will review this after #1018 is fixed and merged (because you have the same commits here). You will be able to rebase after and keep only relevant commit (AP-43). Alternatively, you could include AP-43 in the other branch (pull request) and close this one.

PR #1020, opened 2019-04-16
AP-49 Ensure user is requesting his hub

Alaggaerion:
Added if statement to ensure user is requesting his corresponding hub

### DIFF
--- a/mobile_app_connector/controllers/mobile_app_controller.py
+++ b/mobile_app_connector/controllers/mobile_app_controller.py
@@ -15,7 +15,7 @@ from ..mappings.compassion_login_mapping import MobileLoginMapping
 from ..mappings.app_banner_mapping import AppBannerMapping
 from odoo import http
 from odoo.http import request
-from werkzeug.exceptions import NotFound, MethodNotAllowed
+from werkzeug.exceptions import NotFound, MethodNotAllowed, Unauthorized
 
 
 class RestController(http.Controller):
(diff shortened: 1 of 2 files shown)


mobile_app_connector/controllers/mobile_app_controller.py
```diff
@@ -73,9 +73,13 @@ def get_hub_messages(self, partner_id, **parameters):
         """
         hub_obj = request.env['mobile.app.hub'].sudo()
         if partner_id:
-            # This will ensure the user is logged in
-            request.session.check_security()
-            hub_obj = hub_obj.sudo(request.session.uid)
+            # Check if requested url correspond to the current user
+            if partner_id == request.env.user.partner_id.id:
```
ecino, line 77: Did you test it? Are you sure `request.env.user` is set at that point? I'm not sure that the variable exists here but I may be wrong.

Alaggaerion: Yes it works, I already tested it and I just checked again

ecino: Ok, then it looks good. However it's breaking the automated tests in `test_mobile_app_http`. Can you have a look at it? You can see the errors in the Travis build.

mobile_app_connector/tests/test_mobile_app_http.py
```diff
@@ -55,11 +55,11 @@ def test_hub_authentication(self, oauth_patch):
         json_data = simplejson.loads(response.read())
         self.assertIn('Messages', json_data)
         # Private messages without login should fail
-        response = self.url_open(url + '1')
+        response = self.url_open(url + '3')
```
ecino: `self.env.ref('base.partner_root').id`

ecino (approved):

PR #1022, opened 2019-04-18
AP-43 include agenda and download WP posts

ecino:
Replaces #1019 

- Implementing the agenda and download post services
- Fix image on post service when the medium thumbnail is incorrect
- Add two Cron for refresh data from Wordpress to Odoo


ecino:
@jonathantarabbia I refactored the post fetching to reuse the same code for all post types. I also added a field to enable/disable the posts on the hub (this prevents to re-download the empty posts and check again that they are empty).

PR #1027, opened 2019-04-24
AP-60 Fix feedback service

jonathantarabbia:
- Add new Model Feedback to manage the feedback Partner
 - Add new route feedback to the mobile_app_controller

### DIFF
--- a/mobile_app_connector/controllers/mobile_app_controller.py
+++ b/mobile_app_connector/controllers/mobile_app_controller.py
@@ -10,7 +10,7 @@
 #    The licence is in the file __manifest__.py
 #
 ##############################################################################
-
+from odoo.exceptions import UserError
 from ..mappings.compassion_login_mapping import MobileLoginMapping
 from ..mappings.app_banner_mapping import AppBannerMapping
 from odoo import http
@@ -106,6 +106,41 @@ class RestController(http.Controller):
         host = request.env['ir.config_parameter'].get_param('web.base.url')
         letter_id = parameters['correspondenceid']
         letter = request.env['correspondence'].browse([int(letter_id)])
-        if letter.exists() and letter.letter_image:
-            return host + "b2s_image?id=" + letter.uuid
-        raise NotFound()
+
+        if letter and letter.letter_image:
+            pdf = host + "b2s_image?id=" + letter.uuid
+        return pdf
+
+    @http.route('/mobile-app-api/get-project',
+                type='json', auth='user', methods=['GET'])
+    def get_project(self, **parameters):
+        icp_id = parameters.get('IcpId', '0')
+        if not icp_id:
+            raise UserError(_("Please add IcpId in your configuration"))
+        data = request.env['compassion.project'].sudo().mobile_get_project(
+            icp_id)
+        return data
+
+    @http.route('/mobile-app-api/feedback',
+                type='json', auth='user', methods=['GET', 'POST'])
+    def feedback(self, **parameters):
+        user_id = parameters.get('userid', 0)
+        if not user_id:
+            raise UserError(_("Please add userid in your configuration"))
+        star = parameters.get('star', '3.0')
+        # convert to float for IOS
+        star = "%.1f" % int(star)
+        if not star:
+            raise UserError(_("Please add star in your configuration"))
+        what_did_u_like = parameters.get('Whatdidulike', '-')
+        if not what_did_u_like:
+            raise UserError(_("Please add Whatdidulike in your configuration"))
+        # source parameter is not defined in ios application
+        source = parameters.get('source', 'IOS')
+        improve_app = parameters.get('Improveapp', '-')
+        if not improve_app:
+            raise UserError(_("Please add Improveapp in your configuration"))
+        record = request.env['mobile.app.feedback'].sudo().create(
+            {'name': what_did_u_like, 'partner_id': user_id, 'source': source,
+             'improve_app': improve_app, 'star': star})
+        return record.id
--- a/mobile_app_connector/controllers/mobile_app_controller.py
+++ b/mobile_app_connector/controllers/mobile_app_controller.py
@@ -13,7 +13,7 @@
 from odoo.exceptions import UserError
 from ..mappings.compassion_login_mapping import MobileLoginMapping
 from ..mappings.app_banner_mapping import AppBannerMapping
-from odoo import http
+from odoo import http, _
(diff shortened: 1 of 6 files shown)


mobile_app_connector/models/mobile_feedback.py
```diff
@@ -0,0 +1,42 @@
...
+#    Releasing children from poverty in Jesus' name
+#    @author: Jonathan Tarabbia
+#
+#    The licence is in the file __manifest__.py
+#
+##############################################################################
+import logging
+
+from datetime import datetime
+from odoo import models, fields, _
```
ecino:
```suggestion
from odoo import models, fields
```

mobile_app_connector/controllers/mobile_app_controller.py
```diff
@@ -10,7 +10,7 @@
 #    The licence is in the file __manifest__.py
 #
 ##############################################################################
-
+from odoo.exceptions import UserError
 from ..mappings.compassion_login_mapping import MobileLoginMapping
 from ..mappings.app_banner_mapping import AppBannerMapping
 from odoo import http
```
ecino:
```suggestion
from odoo import http, _
```

mobile_app_connector/controllers/mobile_app_controller.py
```diff
@@ -106,6 +106,41 @@ def download_pdf(self, **parameters):
         host = request.env['ir.config_parameter'].get_param('web.base.url')
         letter_id = parameters['correspondenceid']
         letter = request.env['correspondence'].browse([int(letter_id)])
-        if letter.exists() and letter.letter_image:
```
ecino, line 109: Please revert those changes, because I specifically added this check to avoid errors when there is no pdf image.

mobile_app_connector/controllers/mobile_app_controller.py
```diff
@@ -106,6 +106,41 @@ def download_pdf(self, **parameters):
...
-            return host + "b2s_image?id=" + letter.uuid
-        raise NotFound()
+
+        if letter and letter.letter_image:
+            pdf = host + "b2s_image?id=" + letter.uuid
+        return pdf
+
+    @http.route('/mobile-app-api/get-project',
+                type='json', auth='user', methods=['GET'])
+    def get_project(self, **parameters):
```
ecino: Remove this route which is no more used.

mobile_app_connector/models/mobile_feedback.py
```diff
@@ -0,0 +1,42 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class AppBanner(models.Model):
```
ecino: Correct the class name

mobile_app_connector/models/mobile_feedback.py
```diff
@@ -0,0 +1,42 @@
...
+    partner_id = fields.Many2one('res.partner', 'Partner', readonly=True)
+    improve_app = fields.Char(translate=True, readonly=True)
+    source = fields.Char(translate=True, readonly=True)
+    star = fields.Selection([
+        ('1.0', 'disappointing'),
+        ('2.0', 'Well'),
+        ('3.0', 'very well'),
+        ('4.0', 'Great'),
+        ('5.0', 'Extraordinary')], select=True, readonly=True)
+    create_date = fields.Datetime(
```
ecino, line 40: You don't need to add this field, it is automatically present in all Odoo models.

mobile_app_connector/models/mobile_feedback.py
```diff
@@ -0,0 +1,42 @@
...
+    _description = 'Mobile App Feedback'
+    _order = 'id desc'
+
+    ##########################################################################
+    #                                 FIELDS                                 #
+    ##########################################################################
+    name = fields.Char('What did you like', translate=True, required=True,
+                       readonly=True)
+    partner_id = fields.Many2one('res.partner', 'Partner', readonly=True)
+    improve_app = fields.Char(translate=True, readonly=True)
```
ecino: No need to translate the field, as this will be sent by the mobile app user (we won't translate their feedback)

mobile_app_connector/models/mobile_feedback.py
```diff
@@ -0,0 +1,42 @@
...
+
+class AppBanner(models.Model):
+    _name = 'mobile.app.feedback'
+    _description = 'Mobile App Feedback'
+    _order = 'id desc'
+
+    ##########################################################################
+    #                                 FIELDS                                 #
+    ##########################################################################
+    name = fields.Char('What did you like', translate=True, required=True,
```
ecino: No need to translate the field, as this will be sent by the mobile app user (we won't translate their feedback)

mobile_app_connector/models/mobile_feedback.py
```diff
@@ -0,0 +1,42 @@
...
+                       readonly=True)
+    partner_id = fields.Many2one('res.partner', 'Partner', readonly=True)
+    improve_app = fields.Char(translate=True, readonly=True)
+    source = fields.Char(translate=True, readonly=True)
```
ecino: Should this be a Selection field (we control the possible source values), or is it something sent by the user that can be free text?

mobile_app_connector/models/mobile_feedback.py
```diff
@@ -0,0 +1,42 @@
...
+    star = fields.Selection([
+        ('1.0', 'disappointing'),
```
ecino:
```suggestion
        ('1.0', 'Disappointing'),
```

mobile_app_connector/models/mobile_feedback.py
```diff
@@ -0,0 +1,42 @@
...
+        ('2.0', 'Well'),
+        ('3.0', 'very well'),
```
ecino:
```suggestion
        ('3.0', 'Very well'),
```

ecino (changes requested):
You should add security rules for the new model feedback in file `security/ir.model.access.csv`. Since all mobile app users will be able to send feedback, the group portal users should have creation rights. I suggest these :
```
user_access_mobile_app_feedback,User access app feedback,model_mobile_app_feedback,base.group_portal,1,1,1,0
admin_access_mobile_app_feedback,Admin access app feedback,model_mobile_app_feedback,base.group_system,1,1,1,1
```
You should as well add record rules in `access_rules.xml` to restrict access to portal users so that they can only access their own feedback.

PR #1030, opened 2019-04-25
AP-55 Add feature to configure a FontAwesome on the product

jonathantarabbia:


mobile_app_connector/mappings/compassion_donation_mapping.py
```diff
@@ -19,7 +19,15 @@ class MobileDonationMapping(OnrampMapping):
...
         "DisplayOrder": "id",
         "FundName": "name",
-        "Id": "id"
+        "Id": "id",
+        "ImageIcon": "image_icon",
     }
 
     FIELDS_TO_SUBMIT = {k: None for k, v in CONNECT_MAPPING.iteritems() if v}
+
+    def _process_connect_data(self, connect_data):
```
ecino: Is this function really doing something? I feel it is useless and only putting back the same values as the original ones...

ecino (changes requested):

mobile_app_connector/mappings/compassion_donation_mapping.py
jonathantarabbia: it was a test. sorry. I remove it.

PR #1034, opened 2019-05-01
AP-70 Adding dependency on timezonefinder. Add a timezone field to all project

vwvw:


ecino:
Please edit your commit to include the JIRA task number in the commit message.

ecino:
How long does the migration take ?

vwvw:
It's fast :-) Not api call

ecino (approved):

PR #1040, opened 2019-05-07
Co 2552 Debug the mail to linked partner in crm request

nfluckiger:


crm_request/models/mail_mail.py
```diff
@@ -0,0 +1,23 @@
+from odoo import models, api
```
ecino, line 2:
```suggestion
# coding: utf-8
from odoo import models, api
```

crm_request/models/mail_mail.py
```diff
@@ -0,0 +1,23 @@
...
+
+    @api.multi
+    def _postprocess_sent_message(self, mail_sent=True):
+
+        s = super(MailMail, self)._postprocess_sent_message(mail_sent)
+
+        for mail in self:
+            partner_ids = mail.partner_ids
+            for partner in mail.partner_ids:
+                if partner.contact_type == "attached" and partner.active == True:
```
ecino: Is the value `unarchived_partners` passed in the context? If so, it would be safier to check that and only call the toggle if the partner_id is set in this value in the context.

It would a bit restrict this toggle specifically to this scenario and avoid archiving systemmaticaly when we send e-mails to attached contacts.

ecino (changes requested):

crm_request/models/mail_mail.py
```diff
@@ -0,0 +1,24 @@
...
+
+    @api.multi
+    def _postprocess_sent_message(self, mail_sent=True):
+
+        s = super(MailMail, self)._postprocess_sent_message(mail_sent)
+
+        for mail in self:
+            partner_ids = mail.partner_ids
+            for partner in mail.partner_ids:
+                if partner.contact_type == "attached" and partner.active is True:
```
ecino: Line too long
```suggestion
                if partner.contact_type == "attached" and partner.active:
```

crm_request/models/mail_mail.py
```diff
@@ -0,0 +1,24 @@
...
+                    partner.toggle_active()
+
+            mail.write({'partner_ids': partner_ids})
```
ecino, line 21:
```suggestion
```
This is useless.

crm_request/models/request.py
```diff
@@ -90,6 +90,7 @@ def action_reply(self):
             # flag to re-archive them once the email is sent.
             if partner.contact_type == 'attached' and not partner.active:
                 ctx['unarchived_partners'] = [partner.id]
```
ecino: I think the context value is no more used so you can remove it completely.
```suggestion
```

ecino (changes requested):

crm_request/models/mail_mail.py
```diff
@@ -0,0 +1,25 @@
...
+    _inherit = "mail.mail"
+
+    '''
+        Inherit mail.mail to deactivate a linked partner after a send.
+    '''
+
+    @api.multi
+    def _postprocess_sent_message(self, mail_sent=True):
+
+        s = super(MailMail, self)._postprocess_sent_message(mail_sent)
```
ecino: The method `_postprocess_sent_message` can delete some e-mails. Therefore, the code below will fail (see the error in Travis). I think you should do your method before and call the super at the end.

ecino (changes requested):

ecino (approved):

PR #1044, opened 2019-05-09
AP-68 Sorting algorithm

vwvw:


mobile_app_connector/models/app_tile.py
```diff
@@ -32,6 +32,8 @@ class AppTile(models.Model):
     display_name = fields.Char(
         'Name', compute='_compute_display_name', store=True, readonly=True)
     view_order = fields.Integer('View order', required=True)
+    order_date = fields.Date('Date of publication for ordering')
```
ecino: This field is never used, I guess it can be removed

vwvw: Good catch!

ecino (approved):

PR #1046, opened 2019-05-16
AP-73 Firebase Connector

vwvw:


ecino:
The README of the module is wrong. Have a look at [the doc](https://confluence.compassion.ch/pages/viewpage.action?pageId=10485880) to see how to generate a fresh readme properly.

firebase_connector/models/firebase_registration.py
```diff
@@ -0,0 +1,91 @@
...
+    _name = 'firebase.registration'
+    _description = 'Device registered with Firebase Cloud Messaging'
+
+    registration_id = fields.Char(required=True,
+                                  string="Firebase Registration ID")
+    partner_id = fields.Many2one('res.partner', string="Partner")
+    partner_name = fields.Char(related='partner_id.name', readonly=True)
+    firebase_credentials =\
+        credentials.Certificate(config.get('firebase_api_key'))
+    firebase_app = firebase_admin.initialize_app(
```
ecino: Are you sure you want to do this at the class level? Plus it looks like this variable is never used?

mobile_app_connector/controllers/json_request.py
```diff
@@ -46,7 +54,7 @@ def __init__(self, *args):
                 self.jsonrequest = {}
                 # PUT The GET parameters as the parameters for the controller
                 self.params = {
-                    key: val for key, val in self.httprequest.args.iteritems()
+                    key: val for key, val in self.httprequest.values.iteritems()
```
ecino: Are you sure we don't lose any parameters by changing this? Shouldn't we merge the two fields?

ecino:
Please check the lint status result and correct the errors.

mobile_app_connector/controllers/json_request.py
vwvw: https://werkzeug.palletsprojects.com/en/0.15.x/wrappers/#werkzeug.wrappers.BaseRequest.values

firebase_connector/models/firebase_registration.py
vwvw: It's used internally by firebase.

ecino:
Two travis errors:
``` 
google_application_credentials is not correctly configured in odoo.conf
The model firebase.registration has no access rules, consider adding one. E.g. access_firebase_registration,access_firebase_registration,model_firebase_registration,base.group_user,1,0,0,0
```

- For the first, you could put a fake api_key in the conf file of travis by modifing the command in the travis yml file [here](https://github.com/compassionch/compassion-modules/blob/277a7f534bec195e946c77623736780671cceb8a/.travis.yml#L51)
- For the second one, it's good to define access rules to your new models.

firebase_connector/models/firebase_registration.py
```diff
@@ -0,0 +1,107 @@
...
+#
+##############################################################################
+import logging
+
+try:
+    import firebase_admin
+    from firebase_admin import credentials
+    from firebase_admin import messaging
+except ImportError as e:
+    raise e
```
ecino: What's the point to use the try/except block in this case? I think it would be better to log the error only but not raise it.

firebase_connector/models/firebase_registration.py
```diff
@@ -0,0 +1,107 @@
...
+                                  string="Firebase Registration ID")
+    partner_id = fields.Many2one('res.partner', string="Partner")
+    partner_name = fields.Char(related='partner_id.name', readonly=True)
+
+    _sql_constraints = [
+        ('firebase_id_unique',
+         'UNIQUE(registration_id)',
+         'Firebase registration ID should be unique')]
+
+    def name_get(self):
```
ecino: Alternatively, you can just define the value `_rec_name = 'registration_id'`

firebase_connector/models/firebase_registration.py
```diff
@@ -0,0 +1,107 @@
...
+        res = []
+        for record in self:
+            res.append((record.id, record.registration_id))
+        return res
+
+    def send_message_from_interface(self, context):
+        """
+        Allows to send a notification from Odoo web interface.
+        Button is in the form view but generally hidden.
+        :param context: Context from the view
```
ecino: Normally the context is automatically passed to the method and accessible via `self.env.context`

mobile_app_connector/__manifest__.py
```diff
@@ -39,13 +39,13 @@
...
         'partner_contact_birthdate',
         'base_geolocalize',
+        'firebase_connector',
     ],
     'external_dependencies': {
         'python': ['simplejson'],
     },
     'data': [
         'security/ir.model.access.csv',
-        'security/access_rules.xml',
```
ecino: I don't think you should remove this file from the manifest.

ecino (changes requested):

vwvw:
Should be good now

ecino (approved):

PR #1049, opened 2019-05-21
AP-86 add field for record_filter

jo-bloiz-ornot:


sbc_compassion/models/contracts.py
```diff
@@ -39,6 +39,10 @@ class Contracts(models.Model):
         'Days since sponsor wrote',
         compute='_compute_last_letter'
     )
+    incite_to_write = fields.Boolean(
```
ecino: Here is my suggestion for the field name.
```suggestion
    write_for_birthday_alert = fields.Boolean(
```

sbc_compassion/models/contracts.py
```diff
@@ -39,6 +39,10 @@ class Contracts(models.Model):
...
+        'Increase visibility of writing content',
```
ecino: Add a hint on the purpose on the field (beware of line too long because of PEP8)
```suggestion
        help='True when child will have birthday in less than 3 months and sponsor has not written in this period.',
```

sbc_compassion/models/contracts.py
```diff
@@ -71,6 +75,17 @@ def _compute_last_letter(self):
...
+    def _compute_incite_to_write(self):
+
+        if self.last_letter == -1 or self.last_letter < 30:
+            next_birthday = fields.Date.from_string(
+                self.child_id.birthdate).replace(year=date.today.year)
+
+            if next_birthday - date.today() < timedelta(days=0):
+                next_birthday.replace(year=next_birthday.year+1)
+
+            return next_birthday - date.today() > timedelta(days=30)
```
ecino: the compute method should assign the value to the field instead of returning it. The compute method can also be called on a recordset, so in general a compute method has this signature:
```
def _compute_write_for_birthday_alert_(self):
    for contract in self:
        contract.write_for_birthday_alert = result
```

mobile_app_connector/data/tile_type_data.xml
```diff
@@ -256,8 +256,7 @@ You never wrote to your sponsored child.
         <field name="code">LE_T3</field>
         <field name="view_order">10</field>
         <field name="default_model_id" ref="model_recurring_contract"/>
-<!--        TODO AP-86 call a filter for only selecting those that have birthday in three months and not yet written-->
-        <field name="default_records_filter">lambda c: c.correspondent_id == c.env.user.partner_id</field>
+        <field name="default_records_filter">lambda c: c.incite_to_write</field>
```
ecino:
```suggestion
        <field name="default_records_filter">write_for_birthday_alert</field>
```

ecino (changes requested):
Your compute method is not correct. It should put the Boolean field to true only if the child have his birthday in the following 3 months and that the sponsor didn't write in the 3 months preceding the birthday. This is explained in the task description in Jira.

jo-bloiz-ornot:
======== Testing test_flake8 ========
sbc_compassion/models/contracts.py:90:17: F841 local variable 'write_for_birthday_alert' is assigned to but never used

The problem is that is technically true, how should I fix it

sbc_compassion/models/contracts.py
```diff
@@ -71,6 +76,20 @@ def _compute_last_letter(self):
...
+            if contract.last_letter == -1 or contract.last_letter > 30:
+                next_birthday = fields.Date.from_string(
+                    contract.child_id.birthdate).replace(year=date.today.year)
+
+                # take next year birthday
+                # if birthday already pass for this year
+                if next_birthday - date.today() < timedelta(days=0):
+                    next_birthday.replace(year=next_birthday.year+1)
+
+                write_for_birthday_alert = \
```
ecino:
```suggestion
                contract.write_for_birthday_alert = \
```

sbc_compassion/models/contracts.py
```diff
@@ -71,6 +76,20 @@ def _compute_last_letter(self):
             except TypeError:
                 contract.last_letter = -1
 
+    def _compute_write_for_birthday_alert(self):
+        for contract in self:
+            if contract.last_letter == -1 or contract.last_letter > 30:
```
ecino:
```suggestion
            if contract.last_letter == -1 or contract.last_letter > 90:
```

sbc_compassion/models/contracts.py
```diff
@@ -71,6 +76,20 @@ def _compute_last_letter(self):
...
+                next_birthday = fields.Date.from_string(
+                    contract.child_id.birthdate).replace(year=date.today.year)
+
+                # take next year birthday
+                # if birthday already pass for this year
+                if next_birthday - date.today() < timedelta(days=0):
```
ecino:
```suggestion
                if next_birthday < date.today():
```

sbc_compassion/models/contracts.py
```diff
@@ -71,6 +76,20 @@ def _compute_last_letter(self):
...
+                    next_birthday.replace(year=next_birthday.year+1)
```
ecino:
```suggestion
                    next_birthday = next_birthday.replace(year=next_birthday.year+1)
```

sbc_compassion/models/contracts.py
```diff
@@ -71,6 +76,20 @@ def _compute_last_letter(self):
...
+
+                write_for_birthday_alert = \
+                    next_birthday - date.today() > timedelta(days=30)
```
ecino:
```suggestion
                    next_birthday - date.today() <= timedelta(days=90)
```
I think the test should be the opposite, we want to check that the birthday is in less than 3 months.

ecino (changes requested):

sbc_compassion/models/contracts.py
```diff
@@ -71,6 +76,20 @@ def _compute_last_letter(self):
...
+    def _compute_write_for_birthday_alert(self):
+        for contract in self:
+            if contract.last_letter == -1 or contract.last_letter > 90:
+                next_birthday = fields.Date.from_string(
+                    contract.child_id.birthdate).replace(year=date.today.year)
+
+                # take next year birthday
+                # if birthday already pass for this year
+                if next_birthday < date.today():
+                    next_birthday = next_birthday.replace(year=next_birthday.year+1)
```
ecino: line too long
```suggestion
                    next_birthday = next_birthday.replace(
                        year=next_birthday.year+1)
```

ecino (approved):

PR #1050, opened 2019-05-21
salutation full name

jo-bloiz-ornot:


jo-bloiz-ornot:
https://github.com/CompassionCH/compassion-switzerland/pull/883
linked pull request

thankyou_letters/models/res_partner.py
```diff
@@ -70,12 +71,14 @@ def _compute_salutation(self):
...
                     partner.firstname
+                partner.full_salutation = title_salutation + ' ' + \
+                    partner.firstname + ' ' + partner.lastname
             else:
                 partner.salutation = _("Dear friends of ") + \
                     self.env.user.company_id.name
                 partner.short_salutation = partner.salutation
-                partner.informal_salutation = ("Dear friend of ") + \
-                    self.env.user.company_id.name
+                partner.informal_salutation = partner.salutation
```
ecino: Why did you change the `informal_salutation`? It's not good, it looks like the same, but the informal is in singular `friend` vs `friends`. Please revert this change. The `full_salutation` should be equal to the normal salutation (this is right).

ecino (changes requested):

joblo-compassion:
======== Testing test_flake8 ========
sbc_compassion/models/contracts.py:90:17: F841 local variable 'write_for_birthday_alert' is assigned to but never used

The problem is that is localy true

ecino (approved):

PR #1056, opened 2019-05-28
Fix request stage transition

eicher31:


crm_request/models/request.py
```diff
@@ -188,11 +188,14 @@ def message_new(self, msg, custom_values=None):
...
-                "crm_request.business_closed_email_template").id
-            request.with_context(keep_stage=False).message_post_with_template(
-                template_id)
+        try:
+            if request.holiday_closure_id:
+                template_id = self.env.ref(
+                    "crm_request.business_closed_email_template").id
+                request.with_context(
+                    keep_stage=True).message_post_with_template(template_id)
+        except exceptions.UserError as e:
```
ecino: Why would this code throw a UserError, did you see this case happen?

eicher31: Yes if the template is malformed!

ecino: In that case, it won't throw a `UserError`, but a `QWebException`
https://github.com/odoo/odoo/blob/0f0c5db49dc433520dbb11b46219a2f676f150b3/odoo/addons/base/ir/ir_qweb/qweb.py#L286

ecino (approved):

PR #1063, opened 2019-06-12
CO-2624: Base default language is added to spoken_languages in all cases

Theyiot:
Previously, when a user selected a list of spoken languages which did not propose its base language, a profile was created without the base language to be added to the spoken languages in that case.
This pull request also provide a post-migration.py file, which contains a request to update the spoken language field of all users to also contain the base language. This request does not modify users for which the field was already correct.


ecino:
Great changes, thank you!

PR #1065, opened 2019-06-13
CO-2464

Alaggaerion:
Added no index for controller routes in sms_sponsorship


ecino (changes requested):
Please change the file `oca_dependencies.txt` to change the repository `website` so that it takes the CompassionCH branch instead of the OCA one where the module `no_index` is not available.

ecino (approved):

PR #1066, opened 2019-06-13
AP-94 Provide links for old children's images

Theyiot:
Older pictures from children can now have their own link, to be added in some application's tiles. All that is required is a migration of the module with a provided migration file. Links do not have a random part, but access control has been set up, in order to prevent anybody to browse through the entire list of children's images.

### DIFF
--- a/child_compassion/migrations/10.0.1.3.0/post-migration.py
+++ b/child_compassion/migrations/10.0.1.3.0/post-migration.py
@@ -18,9 +18,7 @@ def migrate(cr, version):
         UPDATE compassion_child_pictures AS pics
         SET image_url = CONCAT(
             'https://erp.compassion.ch/web/image/compassion.child.pictures/',
-            id,
-            '/fullshot/child.jpg'
+            id, '/fullshot/', pics.date, '_', pics.child_id, '.jpg'
         )
         WHERE (pics.image_url <> '') IS NOT TRUE
     """)
-
--- a/child_compassion/migrations/10.0.1.3.0/post-migration.py
+++ b/child_compassion/migrations/10.0.1.3.0/post-migration.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (C) 2018 Compassion CH (http://www.compassion.ch)
+#    @author: Théo Nikles <theo.nikles@gmail.com>
+#
+#    The licence is in the file __manifest__.py
+#
+##############################################################################
+
+
+def migrate(cr, version):
+    if not version:
+        return
+
+    # Update pictures older pictures that don't have a link
+    cr.execute("""
+        UPDATE compassion_child_pictures AS pics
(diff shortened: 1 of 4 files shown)


child_compassion/migrations/10.0.1.3.0/post-migration.py
```diff
@@ -0,0 +1,26 @@
...
+    if not version:
+        return
+
+    # Update pictures older pictures that don't have a link
+    cr.execute("""
+        UPDATE compassion_child_pictures AS pics
+        SET image_url = CONCAT(
+            'https://erp.compassion.ch/web/image/compassion.child.pictures/',
+            id,
+            '/fullshot/child.jpg'
```
ecino: I think I would prefer if we can name the file based on the date of the picture and the local id of the child. It would be like:
```
'/fullshot/', pics.date, '_', child.local_id, '.jpg'
```
It requires a join on the compassion_child table although...

child_compassion/migrations/10.0.1.3.0/post-migration.py
```diff
@@ -0,0 +1,26 @@
...
+        )
+        WHERE (pics.image_url <> '') IS NOT TRUE
+    """)
+
```
ecino: I think you have one line return that should be removed to make travis green.
```suggestion
```

child_compassion/security/child_pictures_security.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="child_model_access" model="ir.model.access">
+        <field name="name">Child model access</field>
```
ecino:
```suggestion
        <field name="name">Child access to portal users</field>
```

child_compassion/security/child_pictures_security.xml
```diff
@@ -0,0 +1,24 @@
...
+        <field name="model_id" ref="model_compassion_child"/>
+        <field name="groups" eval="[(4, ref('base.group_portal'))]"/>
+        <field name="domain_force">[('sponsor_id', '=', user.partner_id.id)]</field>
+        <field name="perm_read" eval="1"/>
+        <field name="perm_write" eval="0"/>
+        <field name="perm_unlink" eval="0"/>
+        <field name="perm_create" eval="0"/>
+    </record>
+    <record id="child_pictures_model_access" model="ir.model.access">
+        <field name="name">Child pictures model access</field>
```
ecino:
```suggestion
        <field name="name">Child pictures access to portal users</field>
```

ecino (changes requested):

child_compassion/migrations/10.0.1.3.0/post-migration.py
Theyiot: Fortunately, the table directly provides the child_id, so no join is required in that case !

Thank you for all the feedbacks

ecino (approved):

PR #1073, opened 2019-06-18
Minor fix: adapt email subject to the number it refers to (is/are)

Theyiot:
I received a mail this morning displaying "1 SMS sponsorship are ongoing". This PR simply change this to write "1 SMS sponsorship is ongoing" in that case.

### DIFF
--- a/sms_sponsorship/models/sms_child_request.py
+++ b/sms_sponsorship/models/sms_child_request.py
@@ -362,7 +362,7 @@ class SmsChildRequest(models.Model):
                 body=_("{} partner(s) have ongoing SMS Sponsorship").format(
                     nb_sms_requests),
                 subject=_("{} SMS Sponsorship {} ongoing").format(
-                    nb_sms_requests, 'is' if nb_sms_requests <= 1 else 'are'),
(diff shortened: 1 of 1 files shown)


sms_sponsorship/models/sms_child_request.py
```diff
@@ -363,8 +363,8 @@ def sms_notification_unfinished_cron(self):
             self.message_post(
                 body=_("{} partner(s) have ongoing SMS Sponsorship").format(
                     nb_sms_requests),
-                subject=_("{} SMS Sponsorship are ongoing").format(
-                    nb_sms_requests),
+                subject=_("{} SMS Sponsorship {} ongoing").format(
+                    nb_sms_requests, 'is' if nb_sms_requests <= 1 else 'are'),
```
ecino: is and are should be wrapped in the translation function.
```suggestion
                    nb_sms_requests, _('is') if nb_sms_requests <= 1 else _('are')),
```

ecino (changes requested):

sms_sponsorship/models/sms_child_request.py
Theyiot: Done.

ecino (approved):

PR #1075, opened 2019-06-18
AP-109_ForgotPwd

Alaggaerion:
Created forgot password service


mobile_app_connector/controllers/mobile_app_controller.py
```diff
@@ -134,3 +134,20 @@ def mobile_app_sponsorship_request(self, lang_code=None, source=None,
...
+        """
+        Called by app when user forgot his password, try to match his email
+        address and then send him reset instructions
+        :return:
+        """
+        if 'email' not in parameters:
+            return {
+                'status': 1,
+                'message': "No email entered"
+                    }
```
ecino: line over-indented, causing travis to fail

mobile_app_connector/models/match_user.py
```diff
@@ -0,0 +1,75 @@
...
+#
+#    The licence is in the file __manifest__.py
+#
+##############################################################################
+
+from odoo import api, models
+from odoo.addons.base.ir.ir_mail_server import MailDeliveryException
+
+
+class MatchUser(models.AbstractModel):
```
ecino: I think that this class is not useful. Why not integrating the logic directly in the controller? Furthermore, the `skip_password_reset` option is there but never used, I think you can simplify a bit the code by removing unecessary options.

ecino (changes requested):

ecino (approved):

PR #1081, opened 2019-06-25
AP-91 Add support for waiting payments in Odoo

Theyiot:
Support is provided for Odoo by this PR. The tile has been created, the functioning of the back-end has been modified so that tiles for not yet supported children can be displayed (but only the ones for the awaiting payments), the communication between front and back-and has been modified accordingly. A few  bugs or small errors are fixed by this PR as well.


mobile_app_connector/models/app_tile.py
```diff
@@ -146,6 +146,40 @@ def render_tile(self, tile_data=None):
...
+        res = []
+        if unpaid_tile is None:
+            unpaid_tile = {}
+        tile_json = unpaid_mapping.get_connect_data(self)
+        records = self._get_records(unpaid_tile)
+        if records:
+            # Convert text templates
+            if self.mode == 'one':
+                tile_json.update(self._render_single_tile(records))
+                tile_json['Child']['SupporterGroupId'] = self.env.user.id
```
ecino: I think this is the single difference between your method and the regular `render_tile` method. I would suggest to remove this method and use the regular one, fix what is needed in there if it needs changes. But it seems that you did the same fix as well in `app_hub.py ` line 178.

ecino (changes requested):
Your changes seem to break the unit tests. Please have a look at the travis result execution. You can test on your server as well using the parameter `--test-enable -i mobile_app_connector` when you run Odoo.

ecino:
Tests are still failing. Something must have broken the hub construction in certain cases (or the tests are no more adapted). Please have a look at that.

Theyiot:
> Tests are still failing. Something must have broken the hub construction in certain cases (or the tests are no more adapted). Please have a look at that.

I am actually working on it !

mobile_app_connector/models/app_hub.py
```diff
@@ -71,12 +77,36 @@ def mobile_get_message(self, partner_id, **pagination):
...
         res = self._construct_hub_message(
             partner_id, messages, children, **pagination)
+
+        # Handle children with awaiting payment
+        if unpaid_children is not None:
+            unpaid_dict = unpaid_children.get_app_json(multi=True)
+            if unpaid_dict.get('Children') is not None:
+                unpaid_dict['UnpaidChildren'] = unpaid_dict['Children']
+                del unpaid_dict['Children']
+                for i in range(len(unpaid_dict['UnpaidChildren'])):
```
ecino, line 101: Isn't it handled already at line `89`?

mobile_app_connector/models/app_hub.py
```diff
@@ -71,12 +77,36 @@ def mobile_get_message(self, partner_id, **pagination):
             'product.product': products,
             'correspondence': letters,
         }
+        unpaid_data = {
+            'recurring.contract': unpaid
+        }
         # TODO handle pagination properly
         limit = int(pagination.get('limit', 1000))
         messages = available_tiles[:limit].render_tile(tile_data)
+        msg_tmp = self.env['mobile.app.tile'].search([
```
ecino: Better to directly grab the tile instead of searching by name which may be changed in the interface:
```
msg_tmp = self.env.ref('mobile_app_connector.tile_gi7').render_tile(unpaid_data)
```

ecino (changes requested):

mobile_app_connector/models/app_hub.py
Theyiot: No, it is two different things. At line `89`, messages containing tile info are updated. At line `100` and afterward, it updates objects that will be sent as **HomeResponse**, which also contains information for "awaiting payment" tiles but not the same.
But I think their should be a solution to avoid both, but I cannot find it...

mobile_app_connector/models/app_hub.py
Theyiot: Perfect, this is changed.

mobile_app_connector/data/tile_type_data.xml
```diff
@@ -119,6 +119,20 @@ Make a donation]]></field>
...
         <field name="tile_preview" type="base64" file="mobile_app_connector/static/img/GI5.png"/>
     </record>
+    <record id="tile_subtype_gi7" model="mobile.app.tile.subtype">
+        <field name="type_id" ref="tile_type_donation"/>
+        <field name="name">Awaiting sponsorship</field>
+        <field name="code">GI7</field>
+        <field name="view_order">5</field>
+        <field name="default_model_id" ref="model_recurring_contract"/>
+        <field name="default_title">${ctx['objects'].sudo().child_id.preferred_name}</field>
+        <field name="default_records_filter">lambda c: (c.state = 'draft' or c.state == 'waiting')
```
ecino:
```suggestion
        <field name="default_records_filter">lambda c: (c.state in ['draft', 'waiting'])
```

mobile_app_connector/models/app_hub.py
```diff
@@ -50,7 +50,13 @@ def mobile_get_message(self, partner_id, **pagination):
         #  is correspondent (to avoid viewing letters when he doesn't write)
         sponsorships = (partner.contracts_correspondant +
                         partner.contracts_fully_managed).filtered('is_active')
+        unpaid = (partner.contracts_correspondant +
```
ecino: I would only take `contracts_fully_managed`, because we don't want to invite him to pay the sponsorship if he isn't the person who pays.

mobile_app_connector/models/app_hub.py
```diff
@@ -50,7 +50,13 @@ def mobile_get_message(self, partner_id, **pagination):
...
+                  partner.contracts_fully_managed).filtered(
+            lambda c: not c.is_active and not c.parent_id and
+                      (c.state == 'waiting' or c.state == 'draft'))
```
ecino:
```suggestion
                      (c.state in ['waiting', 'draft']))
```

ecino (changes requested):

mobile_app_connector/models/app_hub.py
Theyiot: Done. Thank you !

ecino (approved):

PR #1082, opened 2019-06-26
CO-2509 Add button who simplify adding an attachment

eicher31:


partner_communication/report/a4_no_margin.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <report id="report_a4_no_margin"
+            model="partner.communication.job"
+            string="A4 BVR no margin    "
+            report_type="qweb-pdf"
+            name="report_compassion.a4_bvr_no_margin"
+    />
+    <record id="report_a4_no_margin" model="ir.actions.report.xml">
+        <field name="paperformat_id" ref="l10n_ch_payment_slip.paperformat_paymentslip"/>
```
ecino: Use `report.paperformat_euro` instead and remove `l10n_ch` dependency. We shouldn't add any CH dependencies in this repository. Wrap the report in `data noupdate` tag so that we can then change the paperformat without the module to reset it.

partner_communication/report/a4_no_margin.xml
```diff
@@ -0,0 +1,16 @@
...
+    </record>
+
+    <template id="a4_no_margin">
+        <t t-call="report_compassion.partner_communication"/>
```
ecino: We should not use `ch` modules here! I think we can just remove this template because it's never used.

partner_communication/report/a4_no_margin.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <report id="report_a4_no_margin"
+            model="partner.communication.job"
+            string="A4 BVR no margin    "
+            report_type="qweb-pdf"
+            name="report_compassion.a4_bvr_no_margin"
```
ecino: Please put a unique name here starting with partner_communication.

partner_communication/report/a4_no_margin.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <report id="report_a4_no_margin"
+            model="partner.communication.job"
+            string="A4 BVR no margin    "
```
ecino: Please update the string

ecino (changes requested):

partner_communication/report/a4_no_margin.xml
eicher31: Yes, I had understood after the test crash..

partner_communication/report/a4_no_margin.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data noupdate="1">
+        <report id="report_a4_no_margin"
+                model="partner.communication.job"
+                string="A4 BVR no margin"
```
ecino: Please update the string (nothing to do with BVR)

ecino (changes requested):

eicher31:
@ecino other comments?

ecino (approved):

PR #1085, opened 2019-06-27
AP-78 Service for letter writing

vwvw:
Test failing as in devel

### DIFF
--- a/mobile_app_connector/models/compassion_correspondence.py
+++ b/mobile_app_connector/models/compassion_correspondence.py
@@ -8,12 +8,14 @@
 #
 ##############################################################################
 
-from odoo import models, api  # , fields
+from odoo import models, api, fields
+from odoo.tools import plaintext2html
 from ..mappings.compassion_correspondence_mapping import \
     MobileCorrespondenceMapping, FromLetterMapping
 from werkzeug.exceptions import NotFound
+from base64 import b64encode
 
-# from datetime import date, timedelta
+from datetime import date, timedelta
 
 
 class CompassionCorrespondence(models.Model):
@@ -37,11 +39,10 @@ class CompassionCorrespondence(models.Model):
             data = []
             for letter in self:
                 data.append(mapping.get_connect_data(letter))
-        # TODO change to
-        # unread_recently = not(self.email_read and self.read_date <
-        #                       fields.Date.toString(
-        #                           date.today() + timedelta(days=3)))
-        unread_recently = not self.email_read
+
+        unread_recently = not(self.email_read <
+                              fields.Date.to_string(
+                                  date.today() + timedelta(days=3)))
         return {
             'Child': child.get_app_json_no_wrap(),
             wrapper: data,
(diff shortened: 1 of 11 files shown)


mobile_app_connector/models/compassion_correspondence.py
```diff
@@ -114,3 +115,70 @@ def _get_required_param(self, key, params):
         if key not in params:
             raise ValueError('Required parameter {}'.format(key))
         return params[key]
+
+    def mobile_get_preview(self, *_, **other_params):
+        body = self._get_required_param('letter-copy', other_params)
```
ecino, line 138: A method comment explaining the function would be welcome. Please rename as well `_` parameter which interfers with the translation function of Odoo. `args` is more pythonic.

mobile_app_connector/models/compassion_correspondence.py
```diff
@@ -114,3 +115,70 @@ def _get_required_param(self, key, params):
...
+    @api.multi
+    def mobile_send_letter(self, *params, **other_params):
+        params = params[0]
+        template_id = self._get_required_param('TemplateID', params)
+        if 'DbId' in params:
+            # The letter was submitted on the last api call
+            if template_id == "0":
+                return "Letter Submitted "
+            return "Card Submitted "
+        # body = self._get_required_param('Message', params)
```
ecino, line 196: Cleanup your code

mobile_app_connector/models/compassion_correspondence.py
```diff
@@ -114,3 +115,70 @@ def _get_required_param(self, key, params):
...
+        gen.preview()
+        web_base_url = \
+            self.env['ir.config_parameter'].get_param('web.external.url')
+        url = web_base_url + "/web/image/" + gen._name + "/" + \
+            str(gen.id) + "/preview_pdf"
+        return url
+
+    @api.multi
+    def mobile_send_letter(self, *params, **other_params):
+        params = params[0]
```
ecino, line 189: A method comment would be welcome.

mobile_app_connector/models/compassion_correspondence.py
```diff
@@ -114,3 +115,70 @@ def _get_required_param(self, key, params):
...
+        # partner_id = self._get_required_param('SupporterId', params)
+        # iOS and Android do not return the same format
+        if template_id == '0' or template_id == 0:
+            # write a card -> default template
+            template_id = '2'
+        child_id = self._get_required_param('Need', params)
+        if isinstance(child_id, list):
+            child_id = child_id[0]
+        child = \
+            self.env['compassion.child'].search([('id', '=', str(child_id))])
```
ecino: Is id an integer? why converting it to string? I think this is more appropriate :
```suggestion
            self.env['compassion.child'].browse(child_id)
```

mobile_app_connector/models/mobile_writing.py
```diff
@@ -0,0 +1,118 @@
...
+    external_url = fields.Char(translate=True)
+    date_start = fields.Date(
+        readonly=True,
+        states={'new': [('readonly', False)]}
+    )
+    date_stop = fields.Date(
+        readonly=True,
+        states={'new': [('readonly', False)]}
+    )
+    is_active = fields.Boolean()
```
ecino:
```suggestion
    active = fields.Boolean()
```
We did it badly in other models, but `active` is a standard Odoo field which is there for that.

mobile_app_connector/models/mobile_writing.py
```diff
@@ -0,0 +1,118 @@
...
+        return [
+            x.get_json() for x in actives
+        ]
+
+    @api.multi
+    def get_json(self):
+        self.ensure_one()
+        web_base_url = \
+            self.env['ir.config_parameter'].get_param('web.external.url')
+        base_url = web_base_url + "/web/image/" + self.template_ids[0]._name \
```
ecino, line 109: Why having a many2many relation if we use only the first relation? Wouldn't be better to change it to a many2one field?

ecino (changes requested):

vwvw:
One minor fix coming

ecino (approved):

ecino:
Ok, waiting for it.

PR #1086, opened 2019-06-27
AP-122 Add support for ''Not sponsor' tiles

Theyiot:
Users that are connected but do not yet sponsor a child can use this new type of tile: _Not Sponsor_. It is more restrictive than _Public_ because users need to be logged into the app, and less than _Private_ because it allows users that are not the sponsor of a child yet to see these tiles.


ecino:
Not sure if that is needed.

ecino:
Not for now.

PR #1093, opened 2019-07-02
AP-84 Follow change in compassion Switzerland with email_read going from

vwvw:
boolean to datetime

### DIFF
--- a/mobile_app_connector/models/app_hub.py
+++ b/mobile_app_connector/models/app_hub.py
@@ -262,7 +262,7 @@ class AppHub(models.AbstractModel):
         for tile in to_order:
             recent_group = recent_content[tile['SubType']]
             if recent_group['max_number_tile'] > len(recent_group['tiles']) \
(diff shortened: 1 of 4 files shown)


mobile_app_connector/models/app_hub.py
```diff
@@ -262,7 +262,7 @@ def _assign_order(self, messages):
         for tile in to_order:
             recent_group = recent_content[tile['SubType']]
             if recent_group['max_number_tile'] > len(recent_group['tiles']) \
-                    and (tile['SubType'] != 'LE_T1' or tile['UnReadRecently']):
+                    and (tile['SubType'] != 'LE_T1' or tile.get('UnReadRecently', False)):
```
ecino: line too long. Note that returning False is the default behaviour of the get method.
```suggestion
                    and (tile['SubType'] != 'LE_T1' or tile.get('UnReadRecently')):
```

ecino (changes requested):

ecino (approved):

PR #1097, opened 2019-07-04
AP-126 Odoo is now responsible to retrieve the FAQ for the mobile app

Theyiot:
it was previously the application that took care of requesting and parsing the FAQ from the server. Since the IOS version also requires it, instead of having two additional libraries (one for each platform) and doing twice the work, the responsibility has been transmitted to Odoo.
The FAQ is supported in three languages: German (by default), French and Italian. Odoo uses the context to determine which language is the most appropriated.


mobile_app_connector/models/frequently_asked_questions.py
```diff
@@ -0,0 +1,59 @@
...
+import logging
+from odoo import models
+
+_logger = logging.getLogger(__name__)
+
+try:
+    import requests
+    from bs4 import BeautifulSoup
+except ImportError:
+    _logger.log('Please install bs4 and requests')
```
ecino:
```suggestion
    _logger.warning('Please install bs4 and requests')
```

mobile_app_connector/models/frequently_asked_questions.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class PrivacyStatementAgreement(models.Model):
```
ecino: Since this class has only one method and no fields, it can be Abstract and therefore won't need any security access rules.
```suggestion
class FAQ(models.AbstractModel):
```

mobile_app_connector/models/frequently_asked_questions.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    This class contains everything that is related to the Frequently Asked
+    Questions for the application.
+    """
+    _name = 'frequently.asked.questions'
+
```
ecino, line 30:
```suggestion
    _description = 'Mobile App FAQ'

```

ecino (changes requested):

ecino:
Still some pending correction + conflicts to resolve.

mobile_app_connector/security/ir.model.access.csv
```diff
@@ -38,3 +38,4 @@ admin_access_mobile_app_tile_subtype,admin access_mobile_app_tile_subtype,model_
 admin_access_mobile_app_writing,full access_mobile_app_writing,model_mobile_app_writing,sales_team.group_sale_manager,1,1,1,1
 user_access_mobile_app_feedback,User access app feedback,model_mobile_app_feedback,base.group_portal,1,1,1,0
 admin_access_mobile_app_feedback,Admin access app feedback,model_mobile_app_feedback,base.group_system,1,1,1,1
+access_frequently_asked_questions_default,access_frequently_asked_questions_default,model_frequently_asked_questions,base.group_user,1,0,0,0
```
ecino: Access rules are not necessary for abstract models.

mobile_app_connector/models/__init__.py
```diff
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import logging
```
ecino: This is not needed

mobile_app_connector/models/__init__.py
```diff
@@ -1,4 +1,5 @@
...
 ##############################################################################
```
ecino: Can you remove this comment? It's not needed in this init file.

ecino (changes requested):

mobile_app_connector/security/ir.model.access.csv
Theyiot: Done.

mobile_app_connector/models/__init__.py
Theyiot: Done.

mobile_app_connector/models/__init__.py
Theyiot: I'll let it in the code since every init file have it. Thanks for the feedbacks !

ecino (approved):

PR #1106, opened 2019-07-12
AP-137 Fixed weather issue

Theyiot:
Some possible weather values from OpenWeatherApi used to make the application crash because there was no match. Since the application only proposes four possible values, namely:
- Clear
- Clouds
- Storm
- Rain

Every possible value from OpenWeatherApi was mapped on one of these types.

### DIFF
--- a/child_compassion/models/project_compassion.py
+++ b/child_compassion/models/project_compassion.py
@@ -260,18 +260,18 @@ class CompassionProject(models.Model):
         [
             ('Clear', 'Clear'),
             ('Clouds', 'Clouds'),
-            ('Rain', 'Rainy'),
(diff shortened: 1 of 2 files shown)


ecino:
That won't be enough because the second string is only used for user interface, however what will be returned to the app is the first string. You must add in the mapping a conversion so that the correct value is sent to the app. Or add as well the missing values in the app (but that must be more difficult I think).

Theyiot:
> That won't be enough because the second string is only used for user interface, however what will be returned to the app is the first string. You must add in the mapping a conversion so that the correct value is sent to the app. Or add as well the missing values in the app (but that must be more difficult I think).

ecino (approved):

PR #1115, opened 2019-07-25
AP-155 Wordpress posts encoding bug

Theyiot:
The posts fetched from Wordpress used to have an encoding issue when they were entered on the DB. Indeed, they were HTML entities with decimal encoding of some characters. The incriminated strings are now unescaped to prevent this unwanted behaviour.
No script for migration has been written but the posts will be updated in the database at the next Cron. It can even be triggered manually to update the data.

### DIFF
--- a/mobile_app_connector/models/wordpress_post.py
+++ b/mobile_app_connector/models/wordpress_post.py
@@ -8,15 +8,18 @@
 #
 ##############################################################################
 import logging
-from HTMLParser import HTMLParser
 from ..tools import wp_requests
-
 
 from odoo import api, models, fields, _
 from odoo.exceptions import UserError
 from odoo.tools import config
 
 _logger = logging.getLogger(__name__)
+
+try:
+    from HTMLParser import HTMLParser
+except ImportError:
+    _logger.warning("Please install HTMLParser")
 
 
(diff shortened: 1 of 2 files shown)


mobile_app_connector/models/wordpress_post.py
```diff
@@ -92,10 +93,17 @@ def fetch_posts(self, post_type):
         category_obj = self.env['wp.post.category']
         found_ids = []
         try:
+            h = HTMLParser()
             with wp_requests.Session() as requests:
                 for lang in self._supported_langs():
                     params['lang'] = lang.code[:2]
                     wp_posts = requests.get(wp_api_url, params=params).json()
+                    for post in wp_posts:
```
ecino: Why adding this loop and not just changing at line `162`?
```
 'name': h.unescape(post_data['title']['rendered']),
```

ecino (changes requested):

mobile_app_connector/models/wordpress_post.py
Theyiot: Since the field "excerpt" is not used, we can definitely do as you propose.

ecino (approved):

PR #1116, opened 2019-07-26
AP-184 Added date for B2S letter tiles

simhayoz:


ecino:
Merged with some code additions. Thanks.

PR #1126, opened 2019-08-13
AP-223 Method send/deleteDeviceId used to throw an exception, now fixed

Theyiot:
We obtained the following error:
`java.lang.IllegalStateException: Expected a string but was BEGIN_OBJECT at line 1 column 2 path`

### DIFF
--- a/mobile_app_connector/models/firebase_registration.py
+++ b/mobile_app_connector/models/firebase_registration.py
@@ -104,4 +104,4 @@ class GetPartnerMessage(models.Model):
             raise NotImplemented('Operation %s is not supported by Odoo' %
                                  operation)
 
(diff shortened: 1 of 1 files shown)


ecino (commented):
Did you check that this change works as well for iOS? It might be implemented differently on the two apps as we already discovered for some other messages.

Theyiot:
> Did you check that this change works as well for iOS? It might be implemented differently on the two apps as we already discovered for some other messages.

The iOS version gracefully ignores anything that is returned by this function. Hence it works for both Android and iOS as it is.

ecino:
I'm not sure that `gracefully` can qualify anything on the iOS app, but I will still merge it then. Thanks.

PR #1130, opened 2019-08-15
CO-2687 default templates for specific invoice lines

jo-bloiz-ornot:


jo-bloiz-ornot:
upgrade module
-u thankyou_letters

thankyou_letters/views/product_view.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <openerp>
     <data>
-        <record id="product_view" model="ir.ui.view">
+        <record id="    " model="ir.ui.view">
```
ecino, line 4: This is probably a mistake

ecino:
One more thing we should change: when generating the thankyou letter, it was choosing between 3 templates depending on the amount. Now we should remove this and only chose one config, but override some communication values at creation: user_id and send_mode. It should as well do this when it's a custom template for a product.

Please reach to me if it's not clear enough.

PR #1131, opened 2019-08-19
CO-2667 Migrate advanced_translation to V11

ecino:
The only change we had to do is to change a max int to a float("inf")
because int has no max in python 3.
Replaces #1109


advanced_translation/models/res_partner_test.py
```diff
@@ -0,0 +1,10 @@
+from odoo import models
+from odoo.tools import config
+
+if config.get('test_enable'):
+    # Only register the class in testing mode
+    class PartnerTest(models.Model):
+        """ Test class to test the translations by making res.partner
+        a translatable class."""
+        _inherit = ['res.partner', 'translatable.model']
+        _name = 'res.partner'
```
eicher31, line 10: Doesn't work if moved in the folder test?

eicher31 (approved):

advanced_translation/models/res_partner_test.py
ecino: No I tried again but the database is not updated correctly when the class is defined inside the tests folder.

PR #1142, opened 2019-08-27
AP-234 Add info about number of sponsorship beneficiaries

simhayoz:
to get_project request


ecino (changes requested):
Instead of this, you should simply add the field mapping in the file `mappings/compassion_project_mapping.py`

ecino:
You can add as well a new field in mapping like requested here https://github.com/CompassionCH/compassion_uk_android/pull/95#discussion_r318484135

ecino (approved):

ecino:
Thanks for the changes!

PR #1144, opened 2019-08-27
AP-253 Not logged in users can access FAQ

simhayoz:


mobile_app_connector/controllers/mobile_app_controller.py
```diff
@@ -69,6 +69,21 @@ def mobile_app_contact_us(self, **parameters):
         """
         return request.env['crm.claim'].mobile_contact_us(request.jsonrequest)
 
+    @http.route(['/mobile-app-api/frequently.asked.questions/get_faq',
+                 '/mobile-app-api/frequently.asked.questions/get_faq/'],
```
ecino: Why having twice the same path?

mobile_app_connector/controllers/mobile_app_controller.py
```diff
@@ -69,6 +69,21 @@ def mobile_app_contact_us(self, **parameters):
...
+                type='json', auth='public', methods=['GET'])
+    def mobile_app_get_faq(self, **parameters):
+        """
+        FAQ entry point for not logged in users
+        :param parameters: all other optional parameters sent by the request
+        :return: json data for mobile app
+        """
+        odoo_obj = request.env.get('frequently.asked.questions').with_context(
```
ecino: No need to convert language
```suggestion
        faq_handler = request.env.get('frequently.asked.questions')
```

mobile_app_connector/controllers/mobile_app_controller.py
```diff
@@ -69,6 +69,21 @@ def mobile_app_contact_us(self, **parameters):
...
+            lang=_get_lang(request, parameters))
+        model_method = 'mobile_get_faq'
```
ecino: Here you can directly call the method, no need to search for it, as it's not meant to be generic...
```suggestion
        return faq_handler.mobile_get_faq(**parameters)
```

ecino (changes requested):

ecino:
Please check privacy statement route as well, I think the same problem happens for it.

simhayoz:
Corrected route for privacy statement as well (yes there was indeed the same bug for this route)

ecino (approved):
Thank you!

PR #1148, opened 2019-08-29
AP-251 When partner email is not found in app registration add link

simhayoz:
to a template email for setting account manually

### DIFF
--- a/mobile_app_connector/forms/registration_form.py
+++ b/mobile_app_connector/forms/registration_form.py
@@ -349,17 +349,18 @@ if not testing:
                 # partner is not sponsoring a child (but answered yes (form))
                 if not partner or len(partner) > 1:
                     # TODO AP-102 :Ask child ref to try to get a match
-                    link_text = 'Click here to send the template ' \
-                                'email request.'
-                    to = 'info@compassion.ch'
-                    subject = 'Account Setup: User Not Found'
-                    body = 'Please set my mobile app account with the ' \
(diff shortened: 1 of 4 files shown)


mobile_app_connector/forms/registration_form.py
```diff
@@ -349,9 +349,17 @@ def form_before_create_or_update(self, values, extra_values):
...
+                                'email request.'
+                    to = 'info@compassion.ch'
+                    subject = 'Account Setup: User Not Found'
+                    body = 'Please set my mobile app account with the ' \
+                           'following email address: %22' + str(
+                            extra_values['partner_email']) + '%22'
                     raise ValidationError(_(
                         "We couldn't find your sponsorships. Please contact "
-                        "us for setting up your account."))
+                        "us for setting up your account. " +
```
ecino: You should close the translation function `_()` and not include your function inside, because the content is dynamic here.

ecino (changes requested):
Thank you for the changes. I have one problem with your solution, it's that the template is hardcoded and not translatable to the language of the sponsor, as well that  info@compassion.ch can not be changed if another office would use our module for instance.

To circumvent this, I propose that you create an `xml record` of the model `mail.template` and that you put your template inside it. Here in the code you can load the record with `self.env.ref()` and use `body`, `email_to` fields to construct your e-mail.

This way, the marketing team will be able to change the texts as they want. Thanks.

ecino (approved):

PR #1149, opened 2019-09-04
Co 2714: new field for logcall and change for pending call

samybucher:
CO-2714: New field: compassion_id -> we can can now got to the record from the the log. The pending call create now a log instead of a note in the bottom of the page.


ecino:
You should rebase your branch and remove commit CO-2708 which is already merged. You can as well rename your commit CO-2697 to CO-2714.

PR #1156, opened 2019-09-11
AP-238 Implement generate PDF and correspondence

simhayoz:


simhayoz:
Everything listed on JIRA is done but the final generation (it can be done by taking the function generate_pdf on correspondence_template.py and adapting it to the current system)
Most of the explanation are on the README on _/sbc_compassion/FPDF/_
As said generate_pdf is nearly generic and should not need to much refactor to work with the current system
The record of the used example is defined in _/data/correspondence_template_data.xml_

sbc_compassion/models/correspondence_template.py
```diff
@@ -97,10 +103,30 @@ class CorrespondenceTemplate(models.Model):
...
-    text_box_top_position = fields.Float(help='In millimeters')
-    text_box_width = fields.Float(help='In millimeters')
-    text_box_height = fields.Float(help='In millimeters')
+    header_box = fields.Many2one('correspondence.text.box', string="Header")
+    header_box2 = fields.Many2one('correspondence.text.box', string="Header 2")
+    header_box3 = fields.Many2one('correspondence.text.box', string="Header 3")
+    header_box4 = fields.Many2one('correspondence.text.box', string="Header 4")
+    text_boxes = fields.Many2many('correspondence.text.box', 'text_boxes_rel',
+                                  string='Text boxes')
+    text_boxes2 = fields.Many2many('correspondence.text.box',
```
simhayoz, line 112: I'm not sure if this is the right way to do this, but as we have a fixed number of template pages (at most 4 defined by the number of template_image that are non-null), I have fixed the number of header boxes, text boxes and image boxes
If you want something more generic we could just create one many2many for each boxes and template images and create a class with many2many of the current type so that we could have as many template page as needed
Ex:
`text_boxes = fields.Many2many('correspondence.text.template')` and a class `'correspondence.text.template' ` with a field: `fields.Many2many('correspondence.text.box')`

PR #1173, opened 2019-10-15
CO-2758: Fix Stripe and postfinance payement

Nooka10:


cms_form_compassion/controllers/payment_controller.py
```diff
@@ -13,24 +13,97 @@
...
 from odoo.addons.cms_form.controllers.main import FormControllerMixin
+from odoo.exceptions import MissingError
+from werkzeug.exceptions import NotFound
 
 
 class PaymentFormController(website_account, FormControllerMixin):
 
+    @route(['/compassion/payment/invoice/<int:invoice_id>'], type='http',
+           website=True, methods=['GET', 'POST'],
+           auth='public', noindex=['header', 'meta', 'robots'], csrf=False)
```
ecino: I'm not sure that putting csrf to False is a good idea, it could be a security breach. Could you remove it and test if it works without it? Otherwise we have to figure out why it fails and how we can fix that without removing the security check. Especially for payment controllers, I think it's better to be cautious.

cms_form_compassion/controllers/payment_controller.py
```diff
@@ -13,24 +13,97 @@
...
+        }
+
+        template = 'cms_form_compassion.modal_payment_submit' if \
+            kwargs.get('display_type') == 'modal' else \
+            'cms_form_compassion.payment_submit_full'
+        return request.render(template, values)
+
     @route(['/compassion/payment/<int:transaction_id>'],
-           auth="public", website=True)
+           auth="public", website=True, csrf=False)
```
ecino: Remove csrf

cms_form_compassion/controllers/payment_controller.py
```diff
@@ -39,13 +112,81 @@ def payment(self, transaction_id, **kwargs):
...
+            'acquirer_id': transaction.acquirer_id.id,
         }
         template = 'cms_form_compassion.modal_payment_submit' if \
             kwargs['display_type'] == 'modal' else \
             'cms_form_compassion.payment_submit_full'
         return request.render(template, values)
 
+    @route('/compassion/payment/validate',
+           type='http', auth='public', website=True,
+           noindex=['header', 'meta', 'robots'], csrf=False)
```
ecino: Remove csrf

cms_form_compassion/controllers/payment_controller.py
ecino: Add as well the `noindex` parameter here.

cms_form_compassion/controllers/payment_controller.py
```diff
@@ -39,13 +112,81 @@ def payment(self, transaction_id, **kwargs):
...
+        # If no redirection is set, use the default template
+        return request.render(
+            'invoice_postfinance_payment_controller.payment_redirect',
+            {'payment_ok': payment_ok,
+             'tx': tx,
+             }
+        )
+
+    @route(['/compassion/payment/stripe/create_charge'], type='json',
+           auth='public', csrf=False)
```
ecino: Remove csrf

ecino (changes requested):
Add module `website_no_index` in the dependency section of the __manifest__, increase the module version in the manifest and update the readme to explain better what the module does.

ecino (approved):

ecino:
Rebased and Merged in #1176.

PR #1186, opened 2019-10-24
CO-2770 Intervention compassion migration v11

quentingigon:


quentingigon:
I just can't seem to be able to reduce my commit history... :( plz help

quentingigon:
When making a hold on interventions of one of thoses types : _Learning Materials_ and _Wash,_
an error occurs with no stacktrace: **'int' object is not iterable** and the hold is not made

PR #1193, opened 2019-10-31
Ap 134

josueBulle:


mobile_app_connector/models/partner_communication.py
```diff
@@ -0,0 +1,83 @@
...
+            "mobile_notification_auto_send",
+            "mobile_notification_title",
+            "mobile_notification_body",
+            "mobile_notification_destination",
+            "mobile_notification_topic",
+        ])
+
+        return super(CommunicationJob, self)._get_default_vals(vals, default_vals)
+
+    @api.model
```
ecino:
```suggestion
    @api.multi
```

mobile_app_connector/models/partner_communication.py
```diff
@@ -0,0 +1,83 @@
...
+    def send(self):
+        """ Create a mobile notification when requested """
+
```
ecino: You should add a loop because multiple communications can be sent in a single call
```suggestion
        for communication in self:
```

mobile_app_connector/models/partner_communication.py
```diff
@@ -0,0 +1,83 @@
...
+            self.mobile_notification = mobile_notif
+
+            if self.mobile_notification_auto_send:
+                self.mobile_notification.send()
+
+        return super(CommunicationJob, self).send()
+
+    @api.multi
+    def unlink(self):
+        self.mobile_notification.unlink()
```
ecino: Maybe only remove it if it's not sent?
```suggestion
        self.mapped('mobile_notification').filtered(lambda n: not n.sent).unlink()
```

mobile_app_connector/models/partner_communication.py
```diff
@@ -0,0 +1,83 @@
...
+        ('Prayer', 'Prayer'),
+    ], "Destination", default="MyHub")
+
+    mobile_notification_topic = fields.Selection([
+        ('general_notification', 'General channel'),
+        ('child_notification', 'Child channel'),
+        ('spam', "None (bypass user's preferences)")
+    ], "Topic", default='general_notification')
+
+    mobile_notification = fields.Many2one("firebase.notification")
```
ecino: Convention for relational fields naming is to suffix them with _id or _ids:
```suggestion
    mobile_notification_id = fields.Many2one("firebase.notification")
```

mobile_app_connector/models/partner_communication.py
```diff
@@ -0,0 +1,83 @@
...
+        """ Create a mobile notification when requested """
+
+        if self.mobile_notification_send:
+            mobile_notif = self.env["firebase.notification"].create({
+                'title': self.mobile_notification_title,
+                'body': self.mobile_notification_body,
+                'destination': self.mobile_notification_destination,
+                'topic': self.mobile_notification_topic,
+            })
+            mobile_notif.partner_ids = self.partner_id
```
ecino: Directly in the create dict:
```suggestion
                'partner_ids': [(4, self.partner_id.id)]
```

mobile_app_connector/models/partner_communication.py
```diff
@@ -0,0 +1,83 @@
...
+            self.mobile_notification = mobile_notif
+
+            if self.mobile_notification_auto_send:
+                self.mobile_notification.send()
+
+        return super(CommunicationJob, self).send()
```
ecino: Maybe I would start with super() and then send the notification. it seems more logic to me to first send the communication (like e-mail) and then the notification.

mobile_app_connector/models/partner_communication.py
```diff
@@ -0,0 +1,83 @@
...
+        """
+            Check whether the partner has at least one registration_id
+            (i.e. we can send him a notification, he downloaded and logged in the app)
+        """
+        for job in self:
+            job.firebase_registration_exists = \
+                job.partner_id and len(job.partner_id.firebase_registration_ids) > 0
+
+            if not job.firebase_registration_exists:
+                job.mobile_notification_send = False
```
ecino: You cannot assign other fields in a compute method of one field, this is a design that will likely cause issues. Maybe you can just add the check in the `send` method : `if job.firebase_registration_exists`

ecino (changes requested):
Thanks! Here are some fixes:

- You can revert changes on module firebase_connector, they are not relevant

ecino (approved):

ecino:
Thanks!

PR #1207, opened 2019-11-13
CO-2771 Label migration v11

quentingigon:


label/models/label_print.py
```diff
@@ -25,8 +24,8 @@ class LabelPrint(models.Model):
         'ir.actions.act_window', 'Sidebar action', readonly=True,
         help="Sidebar action to make this template available on records "
         "of the related document model")
-    ref_ir_value = fields.Many2one(
-        'ir.values', 'Sidebar button', readonly=True,
+    ref_ir_default = fields.Many2one(
+        'ir.default', 'Sidebar button', readonly=True,
         help="Sidebar button to open the sidebar action")
     model_list = fields.Char('Model List', size=256)
```
ecino: I think this is never used, could you try to delete every reference and see if it still works?

label/models/label_print.py
```diff
@@ -66,6 +65,9 @@ def onchange_model(self):
                     if model_ids:
                         model_list.append(key)
         self.model_list = model_list
+        # TODO: this avoid crashing the app but may not be correct
```
ecino: I think this whole method `onchange_model` is not useful and can be removed (the field in the view is invisible). Try to remove the method and see if it still works.

label/models/label_print.py
```diff
@@ -146,6 +127,6 @@ def name_search(self, name='', args=None, operator='ilike', limit=None):
         data = self.env.context.get('model_list')
         if data:
             args.append(('model', 'in', safe_eval(data)))
-        ret_vat = super(IrModelFields, self).name_search(
+        ret_vat = super().name_search(
```
ecino: Remove the whole class inheritance, I think it's useless. See if it still works.

ecino (commented):
Try some simplification by removing the field model_list.

ecino:
`label/models/label_print.py:12:1: F401 'odoo.tools.safe_eval' imported but unused`

ecino:
`label/models/label_print.py:12:1: F401 'odoo.tools.safe_eval' imported but unused`

label/models/label_print.py
```diff
@@ -51,101 +45,49 @@ def onchange_image(self):
...
                 'name': button_name,
                 'type': 'ir.actions.act_window',
                 'res_model': 'label.print.wizard',
                 'src_model': src_obj,
                 'view_type': 'form',
-                'context': "{'label_print' : %d}" % (data.id),
+                'context': f"{{'label_print' : {data.id}}}",
                 'view_mode': 'form,tree',
                 'target': 'new',
+                'binding_model_id': self.env.ref("base.model_res_partner").id,
```
ecino:
```suggestion
                'binding_model_id': src_obj,
```

ecino (changes requested):
On the label print view, please add labels to the first two fields (by putting them inside a <group>, it should work) :
image

When editting a label.print.field, please make the fields more visible and add labels for the first two fields as well:
image

ecino:
You have mixed all your branches into this one which is not what I want. So please rebase onto Compassion/11.0 and keep only the commits concerning `label`. You can also squash them into one single commit.

PR #1209, opened 2019-11-13
CO-2816: Migrate firebase_connector to v11

Nooka10:


ecino (changes requested):
Fix lints and add `firebase_admin` in the `requirements.txt`.

ecino (approved):

PR #1210, opened 2019-11-15
Move hr_attendance_management from CompassionCH/hr

eicher31:


ecino:
Please make the travis green ;) you should put this on branch 11.0 I think.

ecino:
I propose that we keep the 10.0 on OCA because we don't need to change in production and do the change only for version 11.0

ecino:
Please fix this.

PR #1213, opened 2019-11-19
CO-2826 added colors back in thankyou_letters tree view

quentingigon:


thankyou_letters/views/communication_job_view.xml
```diff
@@ -31,7 +31,7 @@
         <field name="name">communication.job.tree.thankyou.donation.amount</field>
         <field name="model">partner.communication.job</field>
         <field name="arch" type="xml">
-            <tree default_order="amount, date desc">
+            <tree colors="red:state == 'cancel';green:state == 'done';grey:state == 'pending';blue:state == 'call'" default_order="amount, date desc">
```
ecino: Those are not the same colors as before. It should be the same as https://github.com/CompassionCH/compassion-modules/blob/10.0/partner_communication/views/communication_job_view.xml#L73

ecino (changes requested):

ecino (approved):

PR #1221, opened 2019-11-28
CO-2847 smart_tagger migration

josueBulle:


ecino (approved):
@josueBulle Did you check what you wanted to check as you mentioned in our team meeting? Otherwise code looks fine for me.

PR #1222, opened 2019-11-29
CO-2849 Sponsorship tracking migration v11

quentingigon:
There are still a few bugs, the principal being that the kanban columns have no title in SUB Sponsorship view


sponsorship_tracking/__manifest__.py
```diff
@@ -30,12 +29,14 @@
...
     'name': 'Compassion Sponsorships Tracking',
-    'version': '10.0.2.0.1',
+    'version': '11.0.0.0.0',
     'category': 'Other',
     'author': 'Compassion CH',
     'license': 'AGPL-3',
     'website': 'http://www.compassion.ch',
-    'depends': ['sponsorship_compassion'],
+    'depends': ['sponsorship_compassion',
+                'crm_compassion',
```
ecino: Why adding `crm_compassion` as dependency?

quentingigon: It must have been an error, I removed it

PR #1223, opened 2019-12-05
CO-2841 fix bad resolution of composed PDF of letters b2s

quentingigon:


ecino (commented):
Is 300 not an extreme resolution producing output pdf bigger than original? Would 150 be better?

ecino:
@davidwul Any opinion?

ecino (approved):

PR #1231, opened 2019-12-11
CO-2855 responsable of communication determined by lang

quentingigon:
At the creation of the communication, the user responsible for the communication is either the current user or the user who was assigned to the omr_config with the same lang as the partner

### DIFF
--- a/partner_communication/models/communication_config.py
+++ b/partner_communication/models/communication_config.py
@@ -89,6 +89,8 @@ class CommunicationOmrConfig(models.Model):
     config_id = fields.Many2one(
         'partner.communication.config', 'Communication type')
     lang_id = fields.Many2one('res.lang', 'Language')
+    user_id = fields.Many2one(
+        'res.users', 'From', domain=[('share', '=', False)])
 
 
 class CommunicationConfig(models.Model):
@@ -179,6 +181,19 @@ class CommunicationConfig(models.Model):
     ##########################################################################
     #                             PUBLIC METHODS                             #
     ##########################################################################
+    @api.model
+    def get_config_for_lang(self, lang):
+        specific_config_found = False
+        for config in self.omr_config_ids:
+            if config.lang_id:
+                if config.lang_id == lang:
+                    specific_config_found = True
+                    return config
+
+        if not specific_config_found:
+            return self.omr_config_ids.filtered(lambda c: not c.lang_id)
+        return None
+
     @api.model
     def get_send_mode(self):
         send_modes = self.get_delivery_preferences()
--- a/partner_communication/models/communication_config.py
+++ b/partner_communication/models/communication_config.py
@@ -181,18 +181,14 @@ class CommunicationConfig(models.Model):
     ##########################################################################
     #                             PUBLIC METHODS                             #
     ##########################################################################
-    @api.model
+    @api.multi
     def get_config_for_lang(self, lang):
         specific_config_found = False
         for config in self.omr_config_ids:
-            if config.lang_id:
-                if config.lang_id == lang:
-                    specific_config_found = True
-                    return config
+            if config.lang_id == lang:
+                return config
 
-        if not specific_config_found:
-            return self.omr_config_ids.filtered(lambda c: not c.lang_id)
-        return None
+        return self.omr_config_ids.filtered(lambda c: not c.lang_id)
 
     @api.model
     def get_send_mode(self):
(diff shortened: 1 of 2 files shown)


quentingigon:
This PR also corrects the issue CO-2854

quentingigon:
The user assigned to the communication is now chosen in that order:
1. from the omr_config with same lang as the partner
2. from the config itself if no omr_config was found
3. from the current user if none were found before

partner_communication/models/communication_config.py
```diff
@@ -179,6 +181,19 @@ def _validate_attachment_function(self):
     ##########################################################################
     #                             PUBLIC METHODS                             #
     ##########################################################################
+    @api.model
```
ecino:
```suggestion
    @api.multi
```

partner_communication/models/communication_config.py
```diff
@@ -179,6 +181,19 @@ def _validate_attachment_function(self):
...
+    def get_config_for_lang(self, lang):
+        specific_config_found = False
+        for config in self.omr_config_ids:
+            if config.lang_id:
+                if config.lang_id == lang:
```
ecino: This test is sufficient. If config.lang_id is not set, it won't equal to lang...
```
if config.lang_id == lang:
```

partner_communication/models/communication_config.py
```diff
@@ -179,6 +181,19 @@ def _validate_attachment_function(self):
...
+                    specific_config_found = True
```
ecino: This is not useful since you return just afterwards.
```suggestion
```

partner_communication/models/communication_config.py
```diff
@@ -179,6 +181,19 @@ def _validate_attachment_function(self):
...
+                    return config
+
+        if not specific_config_found:
```
ecino: This test will always be true since you return from the function if a specific config is found. You can therefore remove the variable and the test and just return here the filtered function. If not config is found from the filtered function it will return an empty recordset which is fine.
```suggestion
        return self.omr_config_ids.filtered(lambda c: not c.lang_id)
```

partner_communication/models/communication_job.py
```diff
@@ -269,19 +269,29 @@ def _get_default_vals(self, vals, default_vals=None):
...
+            partner = self.env['res.partner'].browse(vals.get('partner_id'))
+            lang_of_partner = self.env['res.lang'].search([
+                ('code', 'like', partner.lang)
+            ])
+            omr_config = config.get_config_for_lang(lang_of_partner)
+            # responsible for the communication is user specified in the omr_config
+            # or user specified in the config itself
+            # or the current user
+            user_id = self.env.uid
+            if omr_config:
```
ecino: This test is enough.
```suggestion
            if omr_config.user_id:
```

partner_communication/models/communication_job.py
```diff
@@ -269,19 +269,29 @@ def _get_default_vals(self, vals, default_vals=None):
...
+                if omr_config.user_id:
+                    user_id = omr_config.user_id.id
+            elif config.user_id:
+                user_id = config.user_id
```
ecino:
```suggestion
                user_id = config.user_id.id
```

ecino (changes requested):

partner_communication/models/communication_job.py
quentingigon: déformation professionnelle par Java :P

ecino:
partner_communication/models/communication_config.py:186:9: F841 local variable 'specific_config_found' is assigned to but never used

partner_communication/models/communication_job.py
```diff
@@ -269,19 +269,29 @@ def _get_default_vals(self, vals, default_vals=None):
...
-        orm_config_of_right_lang = config.omr_config_ids \
-            .filtered(lambda c: c.lang_id.code == user.lang)
-        orm_config = orm_config_of_right_lang[0] if orm_config_of_right_lang \
-            else config.omr_config_ids
+            partner = self.env['res.partner'].browse(vals.get('partner_id'))
+            if partner:
+                lang_of_partner = self.env['res.lang'].search([
+                    ('code', 'like', partner.lang)
+                ])
+                omr_config = config.get_config_for_lang(lang_of_partner)
```
ecino:
```suggestion
                omr_config = config.get_config_for_lang(lang_of_partner)[0:]
```
In order to force take the first config in case there would be multiple configs with no language (that could cause a bug)

ecino (approved):

PR #1235, opened 2019-12-16
CO-2894 report is now setted on the attachments

quentingigon:


partner_communication/views/ir_attachment_view.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <!-- Attachment -->
+        <record id="view_ir_attachment_form" model="ir.ui.view">
+            <field name="model">ir.attachment</field>
+            <field name="inherit_id" ref="base.view_attachment_form"/>
```
ecino: This is already done in file `communication_job_view.xml`, line 117. Please move the record in that file inside your file (it will override your content).

partner_communication/models/communication_attachment.py
```diff
@@ -56,7 +58,11 @@ def create(self, vals):
             })
             vals['attachment_id'] = attachment.id
 
+        vals['report_id'] = \
+            self.env['report']._get_report_from_name(vals['report_name']).id
```
ecino: You should move this at the beggining of the method and only if `report_id` is not already defined in vals.
Then you should add the value of `report_id` inside the vals of the attachment created at line 53.

ecino (changes requested):
You are setting the report_id value at attachment creation. Please provide a migration script that will set all report_id values for the already existing records!

partner_communication/migrations/10.0.1.2.0/post-migration.py
```diff
@@ -0,0 +1,34 @@
...
+        return
+    cr = env.cr
+
+    cr.execute("ALTER TABLE partner_communication_attachment "
+               "ADD COLUMN IF NOT EXISTS report_id INTEGER")
+
+    cr.execute("SELECT id FROM partner_communication_attachment")
+    communication_ids = cr.dictfetchall()
+
+    for communication_id in communication_ids:
```
ecino: You don't need all the `cr` and query thing. Here you want to update the communication attachments so I would just do this :
```
attachments = env['partner.communication.attachment'].search([
    ('report_name', '!=', False),
    ('communication_id', '!=', False)
])
for attachment in attachments:
    attachment.report_id = \
            env['report']._get_report_from_name(attachment.report_name)
```

partner_communication/migrations/10.0.1.2.0/post-migration.py
```diff
@@ -0,0 +1,34 @@
...
+    if not version:
+        return
+    cr = env.cr
+
+    cr.execute("ALTER TABLE partner_communication_attachment "
+               "ADD COLUMN IF NOT EXISTS report_id INTEGER")
+
+    cr.execute("SELECT id FROM partner_communication_attachment")
+    communication_ids = cr.dictfetchall()
+
```
ecino: Remove these lines

ecino (changes requested):

ecino (approved):
Thanks ! :+1:

PR #1236, opened 2019-12-16
CO-2896 Thank you letters migration v11

quentingigon:


ecino (changes requested):
In this commit, I moved the inheritance of `res.partner.title` into the module `sponsorship_compassion`. https://github.com/CompassionCH/compassion-modules/commit/809017c1f0b2c8b6af6d0bb28c706a2613337c24#diff-6b80f44fee4a90294b19ce6882919831L125

However now it fails because the field `gender` is not defined and `thankyou_letters` doesn't depend on `sponsorship_compassion`.

Therefore I propose to move again this class inheritance into the module `advanced_translation` which is a common parent of both modules.

ecino (approved):
:+1:

PR #1239, opened 2019-12-17
CO-2898 Survey phone migration v11

quentingigon:


survey_phone/__manifest__.py
```diff
@@ -30,19 +29,24 @@
...
     'sequence': 150,
     'author': 'Compassion CH',
     'license': 'AGPL-3',
     'website': 'http://www.compassion.ch',
-    'depends': ['survey', 'base_phone'],
+    'depends': ['survey',                       # oca_addons/survey
+                'base_phone',                   # oca_addons/connector-telephony
+                'survey',
+                'partner_contact_birthdate',    # oca_addons/partner_contact
+                'partner_contact_gender',    # oca_addons/partner_contact
```
ecino: Why did you add those two dependencies?

quentingigon: I added them because I needed some fields they were implementing, aka gender and birth-something. If they are in an other module, I will make the required changes

ecino: Ok, but the last one is conflicting with the `gender` field we added in `advanced_translation` module. So please just change `partner_contact_gender` to `advanced_translation`

quentingigon: Ok

ecino (approved):

PR #1243, opened 2019-12-19
CO-2904 answer to alias

quentingigon:


crm_request/models/request.py
```diff
@@ -72,15 +74,19 @@ def action_reply(self):
...
-        if self.partner_id:
-            partner = self._get_partner_alias(
-                self.partner_id, parseaddr(self.email_from)[1]
+        if original_partner:
+            alias_partner = self._get_partner_alias(
+                original_partner, parseaddr(self.email_from)[1]
             )
+            partner = alias_partner
+            if alias_partner == original_partner:
+                partner = original_partner
```
ecino, line 83: These two lines don't do anything and are not necessary.

ecino (changes requested):
@eicher31 If I understand correctly the change, it means that the partner assigned to the request will be the alias contact instead of the main partner. Will this be a problem or not? If we suppose that the alias is archived, does it mean that we will face the issue where the answer is not sent because the partner is archived?

@eicher31 Please test the changes.

crm_request/models/request.py
quentingigon: I don't think they are not necessary as they permit to simplify a bit the code after, when filtering messages (line 87-89)

ecino (approved):
I will test in devel

PR #1244, opened 2019-12-19
CO-2836 PDF storage migration

josueBulle:
https://github.com/CompassionCH/compassion-modules/pull/1244

- Future correspondence created from the app will have the source 'app'
- Correspondences from S2B_generator and import_letter_history do not store the PDF anymore.
- Instead, they store the sponsor's text and attachments separately.
- PDF are then (re-)generated for GMC and upon request using the template's generator. (templates may need to be fixed)

Migration of correspondence with following requirements:
  - direction is 's2b'
  - source is 'website' or 'compassion'
  - has a template_id

What it does:
  - Moves the `original_text` to the `english_text` field
  - Extracts the sponsor's text from the multi-paged PDF
    - Detects and ignores references in the pages
    - Removes line breaks '\n' and stores the text in `original_text`
  - Extract images from the PDF
    - Filters images with their size to ignores backgrounds
    - stores them in original_attachment_ids
  - Delete the PDF from the database (A copy is still present in the NAS)

Migration is splitted in job of 10 correspondences
Jobs are executed in the 'root.sbc_compassion_migration' channel. So it is possible to limit the number of concurrent jobs and add a waiting time between jobs to limit impact on system responsiveness.


sbc_compassion/models/correspondence.py
```diff
@@ -446,6 +462,15 @@ def create(self, vals):
             else:
                 raise UserError(
                     _('You can only attach tiff or pdf files'))
+
+        if 'original_attachment_ids' in vals:
+            vals['original_attachment_ids'] = [(0, 0, {
```
ecino: Normally the value of `original_attachment_ids` should be set correctly when calling the `create` method. It is not the responsibility of the object to convert and format the values, it should be done directly properly anywhere in the code when calling create/write methods on the records.

sbc_compassion/models/correspondence_s2b_generator.py
```diff
@@ -167,16 +174,20 @@ def generate_letters_job(self):
...
                 'template_id': self.s2b_template_id.id,
                 'direction': 'Supporter To Beneficiary',
-                'source': 'compassion',
+                'source': self.source,
                 'original_language_id': self.language_id.id,
                 'original_text': text,
-            })
+            }
+            if self.image_ids:
+                vals['original_attachment_ids'] = self.image_ids
```
ecino: Here you should format the value correctly and remove the code from the create method.

sbc_compassion/controllers/download_correspondence.py
```diff
@@ -0,0 +1,38 @@
...
+from odoo.addons.web.controllers.main import content_disposition
+
+from werkzeug.exceptions import BadRequest, NotFound
+
+_logger = logging.getLogger(__name__)
+
+
+class RestController(http.Controller):
+
+    @http.route('/web/pdf/correspondence', type='http', auth='public', methods=['GET'])
```
ecino: Why is this new route needed?

Is it a problem that the route is public?

sbc_compassion/migrations/10.0.1.6.0/post-migration.py
```diff
@@ -0,0 +1,196 @@
...
+        ('source', '=', 'compassion'),
+        ('letter_format', '=', 'pdf'),
+        ('direction', '=', 'Supporter To Beneficiary'),
+        ('template_id', '!=', False),
+        ('store_letter_image', '=', True)
+    ])
+
+    templates_backgrounds_sizes = retrieve_background_sizes(env)
+
+    correspondences = correspondences.filtered(lambda x: x.id in [27788, 108590])
```
ecino: I think this was for your testing purposes. However I suggest that we split the recordset into small batches (let's say 10 correspondences) and launch the migration of these batches into job queues functions, which are asynchronous.

This way, the migration won't block the launching of the server and it can be done gradually while we can still use Odoo.

ecino (changes requested):

josueBulle:
Dans certains cas, j'ai observé des bouts de textes extraits aux mauvais endroits. Je ne sais pas comment résoudre ce problème.

ecino:
> Dans certains cas, j'ai observé des bouts de textes extraits aux mauvais endroits. Je ne sais pas comment résoudre ce problème.

Did you resolve the issue?

josueBulle:
I added these lines in the requirements.txt for travis because I needed these specific branches.
However, they are only needed for the migration and I don't think we should keep them after the migration.

```
pillow
git+https://github.com/pdfminer/pdfminer.six@master
git+https://github.com/mstamy2/PyPDF2@master
```

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -405,3 +405,33 @@ def is_zip(name):
     """
     ext = os.path.splitext(name)[1]
     return ext.lower() == '.zip'
+
+
+def find_child(env, code):
```
ecino, line 410: What is this for? Seems that the function is not called?

sbc_compassion/tools/import_letter_functions.py
```diff
@@ -405,3 +405,33 @@ def is_zip(name):
...
+    :param env: Odoo's environment
+    :param code: the child's code or local_id
+    :return: The child, when found
+    """
+    field = len(code) == 9 and 'code' or 'local_id'
+    child = env['compassion.child'].search([(field, '=', code)])
+    return child
+
+
+def find_partner(env, code, email):
```
ecino, line 423: What is this for? Seems that the function is not called?

sbc_compassion/tools/import_letter_functions.py
josueBulle: it is used in https://github.com/CompassionCH/compassion-switzerland/pull/972

I thought the function was too long and I could extract some parts in this file
(The same way it was done with func.analyze_attachment)

ecino (approved):

PR #1245, opened 2019-12-20
CO-2883 Childpool filters

quentingigon:


child_compassion/wizards/global_child_search.py
```diff
@@ -123,13 +123,19 @@ class GlobalChildSearch(models.TransientModel):
...
-        self.advanced_criteria_used = (self.state_selected and
-                                       self.state_selected != 'Available') or \
-            self.chronic_illness or self.holding_gp_ids or \
+
+        advanced_criteria_used = self.chronic_illness or self.holding_gp_ids or \
             self.father_alive or self. mother_alive or \
             self.physical_disability or self.completion_date_after or \
             self.completion_date_before or self.local_id
 
+        if advanced_criteria_used:
```
ecino: I don't see why doing the check this way is better than having all the conditions at the same place like how it was done before? I think it's more confusing now...

child_compassion/wizards/global_child_search.py
```diff
@@ -469,6 +476,34 @@ def data_to_json(self, mapping_name=None):
             for key, val in json_result.copy().items():
                 if not val:
                     del json_result[key]
+        if mapping_name == "advanced_search":
```
ecino, line 475: Please add a comment explaining why we need this "workaround".

ecino (changes requested):
Some flake8 issues:
```
child_compassion/wizards/global_child_search.py:136:43: E126 continuation line over-indented for hanging indent
child_compassion/wizards/global_child_search.py:139:5: E303 too many blank lines (2)
child_compassion/wizards/global_child_search.py:471:5: E303 too many blank lines (2)
child_compassion/wizards/global_child_search.py:480:24: E271 multiple spaces after keyword
```

child_compassion/wizards/global_child_search.py
quentingigon: It is necessary because before if we selected for example chronic_illness and mother_alive but the state was 'Available', this would make a simple search instead of an advanced search. I'll make it a bit more simple though

ecino (changes requested):
Two flake8 issues left:
```
child_compassion/wizards/global_child_search.py:128:15: E126 continuation line over-indented for hanging indent
child_compassion/wizards/global_child_search.py:466:5: E303 too many blank lines (2)
```

ecino (approved):

PR #1252, opened 2020-01-07
CO-2836 mobile feedback: add language & date

josueBulle:
Uses the language code given in request parameters


ecino (changes requested):
Please add the language field in the tree view and also in the search view to be able to search and to group by the language field.

PR #1254, opened 2020-01-07
CO-2905 last message of partner quoted when reply on request

quentingigon:


crm_request/models/request.py
```diff
@@ -84,9 +94,10 @@ def action_reply(self):
             if messages:
                 # Put quote of previous message in context for using in
                 # mail compose message wizard
-                message = messages[0]
-                ctx['reply_quote'] = message.get_message_quote()
-                ctx['message_id'] = message.id
+                message = _get_last_message_of_partner(messages, self.partner_id)
```
ecino:
```suggestion
                message = messages.filtered(lambda m: m.author_id == self.partner_id)[-1:]
```

ecino (changes requested):
I think this should work the same in much shorter way.

crm_request/models/request.py
quentingigon: Yes totally, thank you

ecino:
crm_request/models/request.py:87:89: E501 line too long (90 > 88 characters)
:laughing:

crm_request/models/request.py
```diff
@@ -84,9 +84,11 @@ def action_reply(self):
             if messages:
                 # Put quote of previous message in context for using in
                 # mail compose message wizard
-                message = messages[0]
-                ctx['reply_quote'] = message.get_message_quote()
-                ctx['message_id'] = message.id
+                message = messages.filtered(
```
ecino, line 87: In fact by looking more carefully I see that the message are arleady correctly filtered at line 81! Therefore you don't need to filter again and the only required change is to replace `message = messages[0]` as it was done before by `message = messages[-1:]`

ecino (changes requested):

crm_request/models/request.py
quentingigon: I'm not sure because at the line 81, we filter the messages on two conditions -> m.author_id == self.partner_id OR condition2. So if condition2 is true, we could have in messages a message not sent by the partner but by an employee of Compassion

PR #1264, opened 2020-01-21
CO-2984 - option to see all sponsorships in the apps

josueBulle:
The option is added in the res.partner form view under the "Sponsorships" tab
The letters sent to paids-only sponsorships are correctly sent as the official correspondent

@ecino FYI the records for the tiles (in tile_type_data.xml) do not have the noupdate attribute, it might be a mistake


ecino (approved):

ecino:
@josueBulle It is good that we let update tile_type_data. This will only change the default templates but not the active tiles.

PR #1265, opened 2020-01-22
CO-2854 Communication expeditor fix

quentingigon:


quentingigon:
Tested with production API key and working !

partner_communication/models/communication_job.py
```diff
@@ -664,7 +666,8 @@ def _send_mail(self):
                 'attachment_ids': [(6, 0, self.ir_attachment_ids.ids)],
                 'auto_delete': False,
                 'reply_to': (self.email_template_id.reply_to or
-                             self.user_id.email)
+                             self.user_id.email),
+                'email_from': self.env['res.users'].browse(self.env.uid).email
```
ecino: Why not using `self.user_id.email` here as well?
```suggestion
                'email_from': self.user_id.email
```

ecino (approved):

PR #1266, opened 2020-01-22
CO-2979 Requests from app in interaction resume v10

quentingigon:
…ble in the interaction resume of the partner


mobile_app_connector/models/request.py
```diff
@@ -65,6 +67,50 @@ def mobile_contact_us(self, parameters, **params):
...
+        """
+        mail = self.env['mail.mail'].create({
+            'state': 'sent',
+            # 'recipient_ids': [(4, contact_id)],
+            'subject': subject,
+            'body_html': body,
+            'author_id': partner.id,
+            'email_from': partner.email,
+            'sent_date': datetime.today(),
+            'mail_message_id': self.env['mail.message'].create({
```
ecino, line 92: I believe that at line 66 of the code, you can catch the message that was created in the claim (support request) when the method `message_post` is called by doing `message = claim.message_post(...)`.
Then here you could use it and link to it instead of creating a new duplicate message. `message_id` would be passed to the function as parameter `message_id = message.id`
```suggestion
            'mail_message_id': message_id
```

mobile_app_connector/models/request.py
```diff
@@ -65,6 +67,50 @@ def mobile_contact_us(self, parameters, **params):
...
+        :return: None
+        """
+        mail = self.env['mail.mail'].create({
+            'state': 'sent',
+            # 'recipient_ids': [(4, contact_id)],
+            'subject': subject,
+            'body_html': body,
+            'author_id': partner.id,
+            'email_from': partner.email,
+            'sent_date': datetime.today(),
```
ecino: Should use the Odoo function for that, by importing it (`from odoo import fields`)
```suggestion
            'sent_date': fields.Datetime.today(),
```

mobile_app_connector/models/request.py
```diff
@@ -65,6 +67,50 @@ def mobile_contact_us(self, parameters, **params):
...
+            'recipient_address': partner.email,
+            'partner_id': partner.id
+        })
+
+        config = self.env.ref('partner_communication.default_communication')
+        self.env['partner.communication.job'].sudo().create({
+            'partner_id': partner.id,
+            'config_id': config.id,
+            'auto_send': False
+        })
```
ecino: I think creating a mail tracking and communication.job is not required. Could you test if it works without it?

ecino (changes requested):

ecino:
Please add a migration script which would scan the existing requests that were made from the app and create an associated mail with it.

mobile_app_connector/models/request.py
quentingigon: I tried this and the emails did not appear in the interaction resume. I think it's because the mail.messages are different (different models, etc)

mobile_app_connector/models/request.py
quentingigon: It works! I was sure it was necessary..

ecino (approved):

PR #1267, opened 2020-01-22
CO-2979 Requests from app in interaction resume v11

quentingigon:


ecino:
@quentingigon I think these are the exact same changes as v10, right? In that case I can simply close the pull request as I will merge back all v10 changes in the v11 branches.

quentingigon:
> @quentingigon I think these are the exact same changes as v10, right? In that case I can simply close the pull request as I will merge back all v10 changes in the v11 branches.

Yes it is ! And the v10 version is more up-to-date than this one also

PR #1268, opened 2020-01-22
AP-329 Fix app crashes when user has no sponsorship

josueBulle:
Apps expected an empty list of children in that case
+ Fix stack trace in console when the mobile_app endpoint raises an exception
+ Fix Wordpress cron occasional failures


ecino (approved):
Thanks!

PR #1284, opened 2020-02-05
Simplify invoice payment from app

ecino:
When I tested I had several weird issues from invoices changed and not properly updated. Since it was some complex cases that would have even further complexified the code, I decided to change the procedure and simplify it.
@josueBulle can you review these changes?

### DIFF
--- a/mobile_app_connector/models/account_invoice.py
+++ b/mobile_app_connector/models/account_invoice.py
@@ -7,8 +7,12 @@
 #    The licence is in the file __manifest__.py
 #
 ##############################################################################
+import logging
+
 from odoo import models, api, fields, _
 from collections import defaultdict
+
+_logger = logging.getLogger(__name__)
 
 
 class AccountInvoice(models.Model):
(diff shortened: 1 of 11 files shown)


mobile_app_connector/models/account_invoice.py
```diff
@@ -119,50 +62,55 @@ def mobile_post_invoice(self, json_data, **parameters):
             invoice_lines_values.append(l_vals)
```
josueBulle, line 66: With the 50.- sponsorships, the staff will have to split the invoice.lines such that the 8CHF end in the general fund.
(In the cases where the new invoice does not match any existing open invoice)

mobile_app_connector/models/account_invoice.py
```diff
@@ -119,50 +62,55 @@ def mobile_post_invoice(self, json_data, **parameters):
...
+                ('partner_id', '=', wrapper.partner_id),
+                ('state', '=', 'open'),
+                ('amount_total', '=', invoice.amount_total),
+            ], order='date asc', limit=1)
+            sp_invoice_contracts = sponsorship_invoice.invoice_line_ids.mapped(
+                'contract_id')
+            new_invoice_contracts = invoice.invoice_line_ids.mapped('contract_id')
+            common_sponsorships = sp_invoice_contracts & new_invoice_contracts
+            if common_sponsorships and len(common_sponsorships) == \
+                    len(sp_invoice_contracts):
```
josueBulle, line 92: You will never match the 50.- (42+8) sponsorships here. They have 2 lines in the original invoice (with the same contract_id) but only one in the new invoice. (the & operator computes a set)
But maybe you did it on purpose...

josueBulle:
In the end, invoices are matched only when the partner is exactly paying once for all his sponsorships and nothing else (no gifts, no fund donations, only 42.- sponsorships).

So I thought, maybe we should have only one product in the app for paying all the sponsorships (e.g. with the amount of the oldest open invoice) instead of letting the sponsor pay for specific ones.
It's just a thought

mobile_app_connector/models/account_invoice.py
ecino: Yes it will match, the set contains unique elements and I'm comparing the related sponsorships. So even if one invoice have multiple lines, if they all are related to the same sponsorship, the resulting set will contain only one element.

mobile_app_connector/models/account_invoice.py
ecino: Yes, I think we can live with that.

ecino:
> In the end, invoices are matched only when the partner is exactly paying once for all his sponsorships and nothing else (no gifts, no fund donations, only 42.- sponsorships).
> 
> So I thought, maybe we should have only one product in the app for paying all the sponsorships (e.g. with the amount of the oldest open invoice) instead of letting the sponsor pay for specific ones.
> It's just a thought

That's a good thought. I will add a task for this change and we will discuss it with David and Stephane.

mobile_app_connector/models/account_invoice.py
josueBulle: my bad, I thought `.mapped('contract_id')` was a 1 to 1 mapping

PR #1286, opened 2020-02-10
CO-3054 Mobile app connector migration v11

quentingigon:
I tested the most I could, and especially the mappings, using the app when needed.


ecino:
@josueBulle Can you review the code?

josueBulle:
ok, will do

mobile_app_connector/models/app_hub.py
```diff
@@ -175,9 +172,9 @@ def _fetch_wordpress_tiles(self, **pagination):
             start = int(pagination.get('start', 0))
             number_mess = int(pagination.get('limit', 1000))
             offset = (start % number_mess) * self.LIMIT_WORDPRESS_TILES
-            wp_mapping = WPPostMapping(self.env)
+            print(len(available_posts))
```
josueBulle: Fix this line

josueBulle (commented):
- I could not send a contact request. I get 400: bad request with both iOS and android apps, **I was not logged in**.
[21/Feb/2020 10:56:01] "POST /mobile-app-api/contact_us HTTP/1.1" 400

- I then created a test user in odoo and tried to login:
KeyError: ('ir.model.access', <function IrModelAccess.check at 0x7f116f1b4158>, 9, 'compassion.mapping', 'read', True, ('fr_CH',))

- The link to create an account within the app (/fr_CH/registration) leads to 404.
however the same link without the locales code part works

If you have any idea

mobile_app_connector/static/mappings/wp_post_mapping.json
```diff
@@ -0,0 +1,53 @@
...
+          }
+        }
+      }
+    },
+    "Title": {
+      "field": "name"
+    },
+    "ActionText": {
+      "field": "image_url"
+    },
```
josueBulle: The posts in the app have the image_url in the tile title

josueBulle: Screenshot_20200221-115329

quentingigon:
> * I then created a test user in odoo and tried to login:
>   KeyError: ('ir.model.access', <function IrModelAccess.check at 0x7f116f1b4158>, 9, 'compassion.mapping', 'read', True, ('fr_CH',))
> 
> * The link to create an account within the app (/fr_CH/registration) leads to 404.
>   however the same link without the locales code part works

I could create an account from the application by removing fr_CH from url or replacing it by en_US. After that, I just had to do some fixes in module to make the account working. 
I still don't know how to make it accept different locales

quentingigon:
>     * I could not send a contact request. I get 400: bad request with both iOS and android apps, **I was not logged in**.
>       [21/Feb/2020 10:56:01] "POST /mobile-app-api/contact_us HTTP/1.1" 400

I can't access this route, even if I am connected. Always error 400

mobile_app_connector/models/compassion_child_pictures.py
```diff
@@ -41,12 +40,20 @@ def get_app_json(self, multi=False):
...
+                data.append(child.data_to_json("mobile_app_child_pictures"))
         return data
+
+    @api.multi
+    def data_to_json(self, mapping_name=None):
+        res = super().data_to_json(mapping_name)
+        if not res:
+            res = {}
+        res['Date'] = datetime.datetime.strptime(
+            res['Date'], '%Y-%m-%d %H:%M:%S').strftime('%d-%m-%Y %H:%M:%S')
```
josueBulle: res['Date'] = datetime.datetime.strptime(res['Date'], '%Y-%m-%d').strftime('%d-%m-%Y %H:%M:%S')
to fix "does not match format" error

mobile_app_connector/models/request.py
```diff
@@ -54,7 +53,7 @@ def mobile_contact_us(self, parameters, **params):
             'email_from': email,
             'subject': subject,
             'code':
-                self.env.ref('crm_claim_code.sequence_claim_app')
+                self.env.ref('sequence_claim_app')
                     .sudo()._next(),
```
josueBulle: 'mobile_app_connector.sequence_claim_app'
Otherwise it can't find the record and raises an exception

josueBulle (changes requested):
Login & blogs' title are fixed, thanks.
**I had to add `.sudo()` in models/mobile_feedback.py to be able to send feedback from the app**

Now, I want to test child-related features. I created a sponsorship through sms_sponsorship but the child infos cannot be retrieved:
`Associated object not found using mapping Compassion Child, JSON Key ThingsILike, JSON value ['Ball Games']`
And I cannot terminate the sponsorship because:
`column compassion_child_ble.comments does not exist`

It doesn't look like it's related to this PR but I'm stuck and can't test further

ecino:
> Login & blogs' title are fixed, thanks.
> **I had to add `.sudo()` in models/mobile_feedback.py to be able to send feedback from the app**
> 
> Now, I want to test child-related features. I created a sponsorship through sms_sponsorship but the child infos cannot be retrieved:
> `Associated object not found using mapping Compassion Child, JSON Key ThingsILike, JSON value ['Ball Games']`
> And I cannot terminate the sponsorship because:
> `column compassion_child_ble.comments does not exist`
> 
> It doesn't look like it's related to this PR but I'm stuck and can't test further

This is due to GMC sending unexpected values in some fields, here in hobbies. The workaround is to add this value Ball Games in the child.hobby.csv data file, or try with another child that has expected values in its information.

column compassion_child_ble.comments does not exist comes from a recent change (we added this field), and I think it should be fixed if you update module child_compassion.

mobile_app_connector/static/mappings/compassion_child_mapping.json
```diff
@@ -0,0 +1,119 @@
...
+    },
+    "ChildNeedKey": {
+      "field": "local_id"
+    },
+    "CommunityDescription": {
+      "to_json_conversion": false
+    },
+    "Country": {
+      "from_json_conversion": "project_id.country_id.name"
+    },
```
josueBulle: Fatal error in iOS app because of a null value in 'Country'
I think it is caused by this line which has two relations (two dots) ?

I don't really know how mappings works but shouldn't it be something like
`      "field": "project_id.country_id.name",
      "search_relational_record": true`

josueBulle: Also, the hierarchy of the mobile app menu is still broken.
And I have a dozen access error while loading the hub (and being logged in).

Maybe we can debug those together.

quentingigon: I fixed the hierarchy of the mobile app menu in odoo and I am looking at the mapping

josueBulle:
Thank you, it fixed the mapping issue:
```
    "Country": {
      "field": "project_id.country_id",
      "to_json_conversion": "odoo_value.name"
    },
    "CountryCode": {
      "field": "project_id.country_id",
      "to_json_conversion": "odoo_value.code"
    },
```
And  this new security rule in advanced_translation fixed the hub:
`access_advanced_translation_portal,Read access on translations,model_ir_advanced_translation,base.group_portal,1,0,0,0`


There are a few `self.env['mobile.app.settings'].get_param()` call that must be corrected and then I think we are good

quentingigon:
Would the security rule added in advanced translation be more at its place in mobile_app_switzerland with the other rule giving access to the jobs to portal_users ?

PR #1287, opened 2020-02-11
CO-3005 modification to controller. 

quentingigon:
A second PR on compassion-switzerland implements second half of fix. When getting the sponsorship using browse, its parameters were then inaccessible

### DIFF
--- a/sms_sponsorship/controllers/main.py
+++ b/sms_sponsorship/controllers/main.py
@@ -142,8 +142,9 @@ class SmsSponsorshipWebsite(Controller, FormControllerMixin):
            auth='public', website=True, noindex=['robots', 'meta', 'header'])
     def step2_confirm_sponsorship(self, sponsorship_id=None, **kwargs):
         """ SMS step2 controller. Returns the sponsorship registration form."""
-        sponsorship = request.env['recurring.contract'].sudo().browse(
-            sponsorship_id)
+        sponsorship = request.env['recurring.contract'].sudo().search([
+            ('id', '=', sponsorship_id)
+        ])
         if sponsorship.sms_request_id.state == 'step2' or \
                 sponsorship.state in ['active', 'waiting']:
             # Sponsorship is already confirmed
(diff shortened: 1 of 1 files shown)


ecino:
Is it because sponsorship_id is a string? Then, could we just make browse(int(sponsorship_id)) instead? And maybe check that sponsorship_id is not None.

ecino:
https://github.com/CompassionCH/compassion-switzerland/pull/1004

quentingigon:
I will try it

quentingigon:
> Is it because sponsorship_id is a string? Then, could we just make browse(int(sponsorship_id)) instead?

No, I tried all possibilities and with browse(), I had no access to its parameters

sms_sponsorship/controllers/main.py
```diff
@@ -142,20 +142,24 @@ def sms_change_child(self, child_request_id):
...
+                return self.make_response(
+                    'recurring.contract',
+                    model_id=sponsorship and sponsorship.id,
+                    **kwargs
+                )
+            else:
+                raise NotFound()
         else:
-            raise NotFound()
+            raise Exception("Please provide a sponsorship_id")
```
ecino: raise NotFound() was better (for user interface), however you can add a line to log an error in the logs
```
import logging

_logger = logging.getLogger(__name__)

_logger.error("No sponsorship_id was given to the route")

File: sms_sponsorship/controllers/main.py [hunk 2]
```diff
@@ -142,20 +142,24 @@ def sms_change_child(self, child_request_id):
...
+                # Sponsorship is already confirmed
+                return self.sms_registration_confirmation(sponsorship.id, **kwargs)
+            if sponsorship.state == "draft":
+                return self.make_response(
+                    'recurring.contract',
+                    model_id=sponsorship and sponsorship.id,
+                    **kwargs
+                )
+            else:
+                raise NotFound()
```
ecino, line 165: Add a log
```
_logger.error("No valid sponsorship found for given id : %s", sponsorship_id)
```

ecino (changes requested):

ecino (approved):

PR #1288, opened 2020-02-11
v11 fixes

josueBulle:
- Fix logging feature port to v11 in cms_form
- file upload in cms_form
- super() in python 3
- Fix warning: use hasclass() instead of @class in xml xpath
- Fix advanced_translation migration:
  - Refactor buggy and complex get_date and added some tests
  - Fixed translations in child_switzerland, babel does not use the same patterns as strftime


josueBulle:
finally, I'm not sure it is a good idea to move the translations lines in csv (but they still need the fixes)

ecino (changes requested):
The tests are broken.

josueBulle:
I think the tests fail on the branch v11 too, it might not be related to this pr

ecino (approved):

PR #1291, opened 2020-02-17
CO-2963 fix for email_cc

quentingigon:


ecino:
I think we won't fix this, as the module `partner_communication` doesn't depend on contact module (where we can link contact aliases to partners). This fix should therefore be moved in the switzerland modules, but in switzerland modules we always use sendgrid, so this fix doesn't make sense.

PR #1292, opened 2020-02-17
CO-2885 Merger les communications rules avec le module support

SylvainLosey:
La migration et l'envoi de mail à l'air de fonctionner correctement mais à controller

### DIFF
--- a/partner_communication/__manifest__.py
+++ b/partner_communication/__manifest__.py
@@ -28,41 +28,39 @@
 
 # pylint: disable=C8101
 {
-    'name': 'Partner Communication',
-    'version': '11.0.2.0.0',
-    'category': 'Other',
-    'author': 'Compassion CH',
-    'license': 'AGPL-3',
-    'website': 'http://www.compassion.ch',
-    'depends': [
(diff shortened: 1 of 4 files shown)


ecino:
@josueBulle Can you review the code please?

josueBulle:
I think there is an issue with the revision module. Or I don't remember how to use variables in the template.
Just to be sure: when we reply to a `crm.request`, we can only choose a template, not a config. Is it how it is supposed to be ?

(Also, I'm not sure if it's relevant but maybe the `partner.communication.config` generated by the migration script should be defined in xml. So that when we install modules creating `mail.templates` but don't run this very specific migration, the related configs still exist)

SylvainLosey:
> Just to be sure: when we reply to a `crm.request`, we can only choose a template, not a config. Is it how it is supposed to be ?

Yes, when replying to a `crm.request` the interface / functionnality is the same. This PR is meant to facilitate changes to support `email.template` by allowing them to be modified by their corresponding `partner.communication.config`

eicher31 (approved):
Looks good to me!

PR #1295, opened 2020-02-24
CO-2997 some small changes to make partner_communication_switzerland …

quentingigon:
Those changes are needed for the migration of partner_communication_switzerland

### DIFF
--- a/message_center_compassion/tools/onramp_connector.py
+++ b/message_center_compassion/tools/onramp_connector.py
@@ -42,14 +42,14 @@ class OnrampConnector(object):
             connect_url = config.get('connect_url')
             api_key = config.get('connect_api_key')
             if connect_url and api_key:
-                OnrampConnector.__instance._connect_url = connect_url
-                OnrampConnector.__instance._api_key = api_key
+                OnrampConnector._connect_url = connect_url
+                OnrampConnector._api_key = api_key
                 session = requests.Session()
                 session.params.update({
                     'api_key': api_key,
                     'gpid': 'CH'
                 })
-                OnrampConnector.__instance._session = session
+                OnrampConnector._session = session
             else:
                 raise UserError(
                     _('Please give connect_url and connect_api_key values '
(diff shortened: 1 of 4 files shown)


message_center_compassion/tools/onramp_connector.py
```diff
@@ -42,14 +42,14 @@ def __new__(cls):
             connect_url = config.get('connect_url')
             api_key = config.get('connect_api_key')
             if connect_url and api_key:
-                OnrampConnector.__instance._connect_url = connect_url
-                OnrampConnector.__instance._api_key = api_key
+                OnrampConnector._connect_url = connect_url
+                OnrampConnector._api_key = api_key
```
ecino: Are you sure about this? Is it a change from python3? I want to make sure we access the correct instance of the class (this is a Singleton pattern).

quentingigon: I went through the debugger and it always told me that OnrampConnector had no __instance parameter. The program crashed if I didn't do that. But it may be due to something else

quentingigon: So... I put it back and now everything seems to be working.......

ecino:
I have the following error when updating the module label with your changes:
```
"External ID conflict, dynamic_label already refers to a `ir.actions.report` record, you can't define a `ir.ui.view` record with this ID." while parsing None:2, near
<t t-name="label.dynamic_label">
```
Did you test it? Do we have to uninstall/install ? Otherwise maybe we should provide a migration script.

ecino:
No, it doesn't work. When I try creating a communication for a sponsorship, New Dossier, I get the error : `KeyError: 'label.report_label'`

You are not doing it right I think with the id references. Didn't it work just to change the reference in compassion-switzerland as I suggested in a comment and leave the ids as before here ? I will test that.

quentingigon:
Maybe it will work but the names and ids of the report, record and template do not respect the naming convention found across the different reports, where the id of report and record are the same and the id of the template is the same also, minus "report_" at the beginning. I changed those just to apply the same logic for all reports. But if it introduces too many bugs or changes, we can surely do something like you said

PR #1297, opened 2020-02-24
CO-3097 Move module settings to general settings 

SylvainLosey:
- For now all the subsettings that link to the Compassion settings page are still here, but it might be best to group them in a single Settings menu link, as is done in other modules such as Invoicing
- The OCA module connector-telephony/base_phone [adds its settings to all settings block](https://github.com/OCA/connector-telephony/blob/0eda4f60383b1bd7a3d0f4e97f8263996fcbd9b6/base_phone/views/res_config_settings.xml#L15) which is not ideal. I'm not sure how to handle this as this behavior changes in v12.


child_compassion/__manifest__.py
```diff
@@ -60,7 +60,8 @@
         'views/gmc_message_view.xml',
         'views/compassion_reservation_view.xml',
         'views/field_office_disaster_alert_view.xml',
-        'views/res_config_view.xml',
+        'views/availability_management_view.xml',
```
ecino: You should put this after compassion_settings_view, because it depends on it and should therefore be loaded after.

ecino (changes requested):

sponsorship_tracking/views/settings_view.xml
```diff
@@ -5,34 +5,62 @@
...
+                        <div class="o_setting_left_pane"/>
+                        <div class="o_setting_right_pane">
+                            <div class="row mt16">
+                                <label class="col-md-5 o_light_label"
+                                       for="sub_fr"/>
+                                <field name="sub_fr"/>
+                            </div>
+                            <div class="row mt16">
+                                <label class="col-md-5 o_light_label"
+                                       for="sub_fr"/>
```
josueBulle: `sub_de` => `sub_fr`

josueBulle (commented):
Visually it looks great but
As discussed, you need to migrate all the settings `set_values` and `get_value` methods (Now multiple fields are not writing new values on save).

(I used this regexp to find the use of old named settings that will raise an exception when executed:)
`env\[.((?!res.config).)*?.settings.\]`

ecino (approved):

PR #1299, opened 2020-02-27
CO-2821 Changes for report compassion upgrade

quentingigon:


partner_communication/models/communication_job.py
```diff
@@ -390,8 +396,14 @@ def refresh_text(self, refresh_uid=False):
                 fields = self.env['mail.compose.message'].with_context(
                     lang=lang).get_generated_fields(
                     job.email_template_id, [job.id])
+                body = job.report_id.render_qweb_html([job.id])
```
ecino: Here it looks like you do a second pass of templating, is this really needed? I would probably remove the call of mail.compose.message wizard if that is possible because I have the feeling that the operation is now duplicated which is not optmial for performance.

partner_communication/models/communication_job.py
```diff
@@ -345,6 +345,12 @@ def send(self):
         to_print = todo.filtered(lambda j: j.send_mode == 'physical')
         for job in todo.filtered(lambda j: j.send_mode in ('both',
                                                            'digital')):
+            origin = self.env.context.get('origin')
+            # if we print first in a communication with send_mode == both
+            if origin == "both_print" and job.send_mode == 'both':
+                job.send_mode = 'digital'
+                return job._print_report()
```
ecino, line 352: This will update the send_mode to `done` so you might want to call _print_report before setting the send_mode to digital.

partner_communication/models/communication_job.py
```diff
@@ -390,8 +396,14 @@ def refresh_text(self, refresh_uid=False):
...
                     lang=lang).get_generated_fields(
                     job.email_template_id, [job.id])
+                body = job.report_id.render_qweb_html([job.id])
+                body_html = tools.html_sanitize(body[0],
+                                                sanitize_attributes=True,
+                                                sanitize_style=True,
+                                                strip_classes=True)
                 job.write({
-                    'body_html': fields['body_html'],
+                    'body_html': fields['body_html'] if len(fields['body_html']) > 50
```
ecino: I don't understand what you are trying to do here.

ecino (changes requested):

partner_communication/models/communication_job.py
quentingigon: I modified a bit the print_report function to especially correct that (with the origin parameter in context)

partner_communication/models/communication_job.py
quentingigon: I removed the call to get_generated_fields and as you said it wasn't needed

partner_communication/models/communication_job.py
quentingigon: It was sh*t, I changed it

ecino:
`partner_communication/models/communication_job.py:391:13: F841 local variable 'lang' is assigned to but never used`

partner_communication/models/communication_job.py
```diff
@@ -387,12 +393,14 @@ def refresh_text(self, refresh_uid=False):
         for job in self:
             lang = self.env.context.get('lang_preview', job.partner_id.lang)
             if job.email_template_id and job.object_ids:
-                fields = self.env['mail.compose.message'].with_context(
-                    lang=lang).get_generated_fields(
-                    job.email_template_id, [job.id])
+                body = job.report_id.render_qweb_html([job.id])
```
ecino: I think this is wrong! You are using the Qweb report used for printing the communication and this will produce html with address of the partner and all layout for printing on a paper. However here we are only interested to render the content of the communication. It could as well be that the report is not defined for communications that we never print but only send by email! Please correct this. The goal is to use the mail.template associated with the communication in order to produce the content of the communication (both for the body and the subject.)

ecino (changes requested):

partner_communication/models/communication_job.py
quentingigon: I wanted to use the method get_generated_fields() but I don't understand how the body_html is generated and the value is always something like: 
**<p>&lt;p&gt;&amp;lt;p&amp;gt;&amp;lt;br&amp;gt;&amp;lt;/p&amp;gt;&lt;/p&gt;</p>**
I am blocked by this

quentingigon: The subject is also not generated

quentingigon: Like we discussed, it could come from a template error

partner_communication/models/communication_job.py
```diff
@@ -387,12 +393,14 @@ def refresh_text(self, refresh_uid=False):
...
+                body_html = tools.html_sanitize(body[0],
+                                                sanitize_attributes=True,
+                                                sanitize_style=True,
+                                                strip_classes=True)
                 job.write({
-                    'body_html': fields['body_html'],
-                    'subject': fields['subject'],
+                    'body_html': body_html,
+                    'subject': self.subject
```
ecino: I propose that you revert all those changes, as I never noticed such issues. If we notice problems in the future we can open another issue. But in general it's best to understand the root cause of a problem before trying something to "patch" an issue.

ecino (changes requested):

partner_communication/models/communication_job.py
quentingigon: ok

ecino (approved):

PR #1305, opened 2020-03-04
fixed support demand from public user

quentingigon:


mobile_app_connector/models/request.py
```diff
@@ -58,17 +65,15 @@ def mobile_contact_us(self, parameters, **params):
...
+            'language': self.sudo().detect_lang(question).lang_id.code
         })
         claim.message_post(body=question,
                            subject=_("Original request from %s %s ") %
                            (firstname, lastname))
-
-        if contact_id:
-            partner = self.env['res.partner'].browse(int(contact_id))
-            if partner.exists():
-                self.create_email_for_interaction_resume(subject, question, partner)
```
ecino, line 71: This should not removed, as the route can send directly the id of the logged partner! We won't ignore that and replace by an email search which could not find the partner.

mobile_app_connector/models/request.py
```diff
@@ -49,6 +54,8 @@ def mobile_contact_us(self, parameters, **params):
         else:
             contact_id = None
 
+        partner = self.sudo().env['res.partner'].search([('email', 'like', email)])
```
ecino:
```suggestion
        partner = self.sudo().env['res.partner'].search([('email', 'ilike', email)], limit=1)
```

mobile_app_connector/models/request.py
```diff
@@ -99,3 +104,29 @@ def create_email_for_interaction_resume(self, subject, body, partner):
...
+            return False
+        for lang in langs:
+            if lang.get("code") == code_lang:
+                language_name = lang.get("name")
+                break
+        if not language_name:
+            return False
+
+        return self.env['res.lang.compassion'].search(
+            [('name', '=ilike', language_name)], limit=1)
```
ecino: Remove this! It's already defined in crm_request module

ecino (changes requested):

mobile_app_connector/models/request.py
quentingigon: I thought it was the way to do it as it was already done in 2 modules, I will change it

ecino (changes requested):
mobile_app_connector/models/request.py:55:89: E501 line too long (96 > 88 characters)
mobile_app_connector/models/request.py:104:1: W391 blank line at end of file

quentingigon:
Still trying to make the mobile.app.writings appear in the app

quentingigon:
I created exactly the same writings as in devel but the images would not charge. By looking at the logs of the app in Android Studio, we found this error when the app tries to get the images:

**W/Glide: Load failed for 192.168.200.125:8069/web/image/correspondence.template/11/template_image with size [840x1172]
    class com.bumptech.glide.load.engine.GlideException: Failed to load resource**

The server received the GET request and responded with 200 but still no images.
My guess is that http links are not as well supported as https ones

quentingigon:
I did it !

mobile_app_connector/models/mobile_writing.py
```diff
@@ -109,7 +109,7 @@ def get_json(self):
             + "/"
         return {
             'templates': [
-                {'templateURL': base_url + str(t.id) + "/template_image",
+                {'templateURL': "http://" + base_url + str(t.id) + "/template_image",
```
ecino: The `web.external.url` parameter should include the 'http://' prefix. I think it's best to remove it from here and correcting your parameter in your database. In production we have https:// as well

mobile_app_connector/models/request.py
```diff
@@ -58,17 +65,15 @@ def mobile_contact_us(self, parameters, **params):
             'name': question,
             'categ_id': self.env['crm.claim.category'].sudo().search(
                 [('name', '=', source)]).id,
-            'partner_id': contact_id,
-            'user_id': False
+            'partner_id': contact_id or partner.id,
```
ecino: This will work as well
```suggestion
            'partner_id': partner.id,
```

ecino (changes requested):

mobile_app_connector/models/mobile_writing.py
quentingigon: I didn't know...

mobile_app_connector/models/request.py
quentingigon: ok!

ecino (approved):

PR #1309, opened 2020-03-06
CO-3130 modules of the repo are installable in v12.0

quentingigon:

PR #1310, opened 2020-03-06
12.0

josueBulle:
Replaces #1309


crm_compassion/models/event_compassion.py
```diff
@@ -370,22 +368,22 @@ def onchange_type(self):
...
-            event.hold_start_date = event.start_date - dt
+            event.hold_start_date = (event.start_date - dt).date()
             if not event.end_date or event.end_date < event.start_date:
                 event.end_date = event.start_date
 
     @api.onchange('end_date')
     @api.multi
     def onchange_end_date(self):
-        days_after = self.env['res.config.settings'].sudo().get_param(
+        days_after = self.env['demand.planning.settings'].get_param(
```
ecino: The class demand.planning.settings was removed, I think this change should be reverted

crm_compassion/models/event_compassion.py
```diff
@@ -454,8 +449,8 @@ def allocate_children(self):
             no_money_yield /= self.number_allocate_children
             yield_rate /= self.number_allocate_children
         expiration_date = self.end_date + \
-            timedelta(days=self.env['res.config.settings'].sudo()
-                      .get_param('days_hold_after_event'))
+            timedelta(days=self.env['demand.planning.settings'].
```
ecino: this should be reverted

ecino (changes requested):

PR #1311, opened 2020-03-09
Fixes for hr_attendance_management

quentingigon:
This is a fix proposition for 2 problems that can occur when using hr_attendance_management. 
It locks a field behind a Condition/Mutex because this field is accessed around the same time by the methods that depends on the same parameter -> _employee.period_ids_.
It should help with the KeyError problem and with the creation of attendances by the CRON

### DIFF
--- a/hr_attendance_management/models/hr_employee.py
+++ b/hr_attendance_management/models/hr_employee.py
@@ -126,9 +126,9 @@ class HrEmployee(models.Model):
             final_balance = None
 
             with cond:
-                _logger.info("waiting predicate")
+                # _logger.info("waiting predicate")
                 cond.wait_for(employee.condition())
-                _logger.info("Condition passed")
+                # _logger.info("Condition passed")
 
                 if employee.period_ids:
                     employee_history_sorted = \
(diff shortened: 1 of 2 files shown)


ecino:
@josueBulle Can you review?

josueBulle (commented):
It feels really hacky. And it will work only once. The new field `periods_computed` is never set to False.

hr_attendance_management/models/hr_employee.py
```diff
@@ -172,6 +184,18 @@ def _compute_balance(self, store=False):
...
                                    lost,
                                    employee.extra_hours_continuous_cap)
+        return True
+
+    def condition(self):
+        def predicate():
+            return self.periods_computed
+        if not predicate():
+            # _logger.info("Computing periods")
+            self._compute_periods()
```
josueBulle: Now the `_compute_periods` is called twice.
`api.depends` → `_compute_periods`
`api.depends` → `_compute_balance` → `condition` → `_compute_periods`

eicher31 (approved):

PR #1315, opened 2020-03-20
Printers

josueBulle:
CO-2895
CO-2900
CO-3184

merge first `report-print-send/printer_tray`


eicher31 (commented):
thx

ecino (approved):

ecino (changes requested):
@eicher31 Tests are broken and should be fixed. Flake8 issue doesn't come from this PR but from another change made in devel I think. It can however be corrected.

PR #1322, opened 2020-03-31
CO-3191 correction notification bug

quentingigon:


SylvainLosey (approved):

mobile_app_connector/models/partner_communication.py
```diff
@@ -45,6 +45,15 @@ def _get_default_vals(self, vals, default_vals=None):
...
 
+    @api.model
+    def create(self, vals):
+        res = super(CommunicationJob, self).create(vals)
+
+        if res.config_id and res.config_id.mobile_notification_send:
+            res['mobile_notification_send'] = True
+
+        return res
+
```
ecino: This has no effect and can be removed. It's already handled here https://github.com/CompassionCH/compassion-modules/blob/6170ebe7ace9de109b0b1472ec82d6e4b29a14a7/mobile_app_connector/models/partner_communication.py#L48 or did you experience a problem with the value not being copied to the job? The rest is the good fix (replacing self by job).

ecino (changes requested):

mobile_app_connector/models/partner_communication.py
quentingigon: I don't really remember why I did that. As you said it's already done in partner_communication so I will remove it here

ecino (approved):

PR #1324, opened 2020-04-07
New payment forms for v12.0

ecino:
@quentingigon Please review, replaces #1321


quentingigon (approved):

cms_form_compassion/forms/match_partner_form.py
```diff
@@ -8,183 +8,182 @@
...
+    partner_email = fields.Char("Email", required=True)
+    partner_phone = fields.Char("Phone", required=True)
+    partner_street = fields.Char("Street", required=True)
+    partner_zip = fields.Char("Zip", required=True)
+    partner_city = fields.Char("City", required=True)
+    partner_country_id = fields.Many2one(
+        "res.country", "Country", required=True, readonly=False
+    )
+    partner_state_id = fields.Many2one("res.country.state", "State", readonly=False)
+    partner_lang = fields.Selection(_lang_get, "Language")
```
quentingigon: should be "_lang_get"

cms_form_compassion/forms/payment_form.py
```diff
@@ -7,125 +7,62 @@
...
-            all_vals.update(extra_values)
-            return all_vals
 
-        def _edit_transaction_values(self, tx_values, form_vals):
-            """ Hook to setup custom values for payment. """
-            pass
+    def generate_invoice(self):
+        """Inherit this method in order to create the invoice that will
+        be used for the web payment."""
+        return self.env["account.invoice"]
```
quentingigon, line 59: is this correct ? Shouldn't it be self.env["account.invoice"].create() ?

sms_sponsorship/forms/sms_sponsorship_registration_form.py
```diff
@@ -9,174 +9,169 @@
...
-            if not self.main_object.sudo().origin_id:
-                fieldset.append(
-                    {
-                        "id": "origin",
-                        "title": _("How did you hear about Compassion?"),
-                        "fields": ["origin_text"],
-                    }
-                )
+class PartnerSmsRegistrationForm(models.AbstractModel):
+    _name = "cms.form.recurring.contract"
```
quentingigon, line 18: add description ?

quentingigon (commented):
Seems pretty good, I just had some comments

cms_form_compassion/forms/payment_form.py
ecino: This is the abstract form which purpose is to be inherited. Returning this will just return an empty recordset (using create would raise an error because you cannot create an invoice without any values).

sms_sponsorship/forms/sms_sponsorship_registration_form.py
ecino: Good point.

cms_form_compassion/forms/match_partner_form.py
ecino: I used lambda instead. The string would be wrong, the parameter should be a list of tuples.

PR #1327, opened 2020-04-15
Adapt to v12

eicher31:


quentingigon (commented):
The changes are good but there is one error in the tests apparently: 
**and (start_date - previous_end_date).days > 1**
where we compare a "datetime.datetime" and "datetime.date"

PR #1335, opened 2020-05-26
AP-347 Create account for sponsorships in creation

SylvainLosey:
Could not test completely

### DIFF
--- a/mobile_app_connector/forms/registration_form.py
+++ b/mobile_app_connector/forms/registration_form.py
@@ -346,16 +346,23 @@ class RegistrationSupporterForm(models.AbstractModel):
         if self.form_next_url() == "/registration/confirm":
             # form submitted
             partner_email = self._sanitize_email(extra_values["partner_email"])
+
             # Find sponsor given the e-mail
-            partner = (
+            matching_partner = (
                 self.env["res.partner"]
                 .sudo()
-                .search(
+                .search([("email", "ilike", partner_email)])
+            )
+
+            # Filter to partner that have sponsorships active or in creation
+            partner = matching_partner.filtered(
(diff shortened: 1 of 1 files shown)


mobile_app_connector/forms/registration_form.py
```diff
@@ -346,16 +346,23 @@ def form_before_create_or_update(self, values, extra_values):
...
+                .search([("email", "ilike", partner_email)])
+            )
+
+            # Filter to partner that have sponsorships active or in creation
+            partner = matching_partner.filtered(
+                lambda partner: self.env["recurring.contract"].search_count(
                     [
-                        ("email", "ilike", partner_email),
-                        ("has_sponsorships", "=", True),
+                        ("partner_id", "=", partner.id),
```
ecino, line 362: The sponsor may be the paying partner or the corresponding partner in the sponsorship.
```suggestion
                        "|", ("partner_id", "=", partner.id),
                        ("correspondent_id", "=", partner.id),
```

mobile_app_connector/forms/registration_form.py
```diff
@@ -346,16 +346,23 @@ def form_before_create_or_update(self, values, extra_values):
...
+                        ("state", "not in", ["cancelled", "terminated"]),
                     ]
                 )
+                > 0
```
ecino: This is not needed. If search_count is 0 it will automatically evaluate to False.

ecino (changes requested):

mobile_app_connector/forms/registration_form.py
SylvainLosey: Ah ... true :)

ecino (approved):

PR #1336, opened 2020-05-26
Better zip in forms

quentingigon:
Half of fix for better zips


quentingigon:
Other half is [HERE](https://github.com/CompassionCH/compassion-switzerland/pull/1091)

PR #1337, opened 2020-05-27
CO-3106 colors of manually logged emails and phone calls 

quentingigon:
The colors are the following:
- **purple** for logged phone calls
- **blue** for logged emails
- **green** for opened emails
- **red** for errors

There is still a bug with the tracking status of manually logged email -> it always is "error". After that, the task should be done

### DIFF
--- a/crm_compassion/wizards/partner_log_interaction_wizard.py
+++ b/crm_compassion/wizards/partner_log_interaction_wizard.py
@@ -31,7 +31,7 @@ class LogInteractionWizard(models.TransientModel):
         in order for it to appear in the interaction.resume view
         :return: True
         """
-        return self.env["mail.mail"].create(
+        mail = self.env["mail.mail"].create(
             {
                 "state": "sent",
                 "recipient_ids": [(4, self.partner_id.id)],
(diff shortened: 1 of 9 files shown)


ecino:
> There is still a bug with the tracking status of manually logged email -> it always is "error". After that, the task should be done

Can't we just force it to "Delivered" or "Manual" in that case? I don't know if it's a big change, if yes, could you create the Jira task? If not, you may quickly change it.

quentingigon:
The last commit fixes that issue, I should have specified it. It happened because the status is taken from the mail_tracking_email object that is created when a mail is created. It had no status, so the value was set to "error". The value is now set to "sent"

crm_compassion/wizards/partner_log_interaction_wizard.py
```diff
@@ -52,3 +53,14 @@ def log_interaction(self):
...
+        email = mail._send_prepare_values(partner=self.partner_id)
+        vals = mail._tracking_email_prepare(self.partner_id, email)
+        vals.update({
+            "state": "sent"
+        })
+        mail_tracking_obj = self.env['mail.tracking.email']
+        mail_tracking_obj.search([
+            ('mail_id', '=', mail.id)
+        ]).unlink()
+        mail_tracking_obj.sudo().create(vals)
```
eicher31, line 66: Shouldn't you return the mail?

quentingigon: It's not the mail that is created at the end, but a mail_tracking_email and it's automatically linked to the mail created at the beginning. And as it's "fake emails", no need to send or return them, creating them is enough

eicher31 (approved):

PR #1344, opened 2020-06-15
CO-3238 Fix bug with mappings

SylvainLosey:
The mapping_id field of many rows in gmc_action was empty, which led to the bug.

Need to update modules child_compassion, intervention_compassion, sbc_compassion and sponsorship_compassion. The migration sets noupdate to false to those fields and the related xml is now no_update=0, so the value of the missing rows will be updated with the value of the xml.

NB:
- Loading gmc_action.xml is very slow, normal
- Might need to run the update of the 4 modules twice (not sure when the query will be run depending on which module is updated first, didn't want to create 4 migrations)


ecino (changes requested):
I think you should leave the noupdate to 1 in the xml files, and replace your migration with the use of function `openupgrade.load_xml()` which can force to load xml data even if it has the noupdate flag. In that way we load it once to correct the data but afterwards we leave it to noupdate (to optimize update process of the modules).

Also you can move your migration in `post-migration` file.

SylvainLosey:
Thanks, I didn't know that was possible.

ecino (approved):

PR #1346, opened 2020-06-16
C0-3239 12.0 fix field mapping

wvmrtn:


child_compassion/static/mappings/child.json
```diff
@@ -107,15 +109,15 @@
     "SponsoredStatus": "sponsorship_status",
     "ThingsILike": {
       "field": "hobby_ids.name",
-      "search_relational_record": true
+      "search_relational_record": true,
```
ecino: You cannot have a final `,` : it will break the JSON.
```suggestion
      "search_relational_record": true
```

child_compassion/static/mappings/child.json
```diff
@@ -107,15 +109,15 @@
...
     },
     "ThingsILikeAge1To2": {
       "field": "hobby_ids.name",
-      "search_relational_record": true
+      "search_relational_record": true,
```
ecino:
```suggestion
      "search_relational_record": true
```

child_compassion/static/mappings/child.json
```diff
@@ -107,15 +109,15 @@
...
     },
     "ThingsILikeToDoAge3Plus": {
       "field": "hobby_ids.name",
-      "search_relational_record": true
+      "search_relational_record": true,
```
ecino:
```suggestion
      "search_relational_record": true
```

ecino (changes requested):

ecino (approved):

PR #1354, opened 2020-06-24
Various fixes on printer

SylvainLosey:
- Use BytesIO instead of StringIO
- Unpack return of render_qweb_pdf to correctly use the pdf data and document type
- Remove the condition leaving the communication on pending when the context 'both_print' is present

### DIFF
--- a/partner_communication/models/communication_job.py
+++ b/partner_communication/models/communication_job.py
@@ -10,7 +10,7 @@
 import logging
 import threading
 from html.parser import HTMLParser
-from io import StringIO, BytesIO
+from io import BytesIO
 
 from reportlab.lib.colors import white
 from reportlab.lib.units import mm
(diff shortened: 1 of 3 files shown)


partner_communication/models/communication_job.py
```diff
@@ -781,8 +781,6 @@ def _print_report(self):
             state = "done"
             if job.need_call == "after_sending":
                 state = "call"
-            elif origin == "both_print":
-                state = "pending"
```
ecino: The code is right and should not be changed. It's the view in xml that is wrong.

ecino (changes requested):

partner_communication/models/communication_job.py
SylvainLosey: I pushed a corrected version with this change. However I find it a bit weird because as far as I can see `both_print` is only set in this XML, and as such I don't see when the conditions related to `both_print` will be triggered if it is removed.

ecino: Yes in fact we will need to put it back in the view in a correct way. But for now I think I will already merge this.

PR #1357, opened 2020-06-25
CO-3288 fix relation not found for FCP without translation

wvmrtn:
Translation of missing relational fields coming in next PR.


wvmrtn:
Closed PR since new fix on translation of missing fields has been done in another PR

PR #1370, opened 2020-07-03
CO-3291 manual translate gmc values

ecino:
This allows staff to translate values received by GMC if the translations are not automatically available.

### DIFF
--- a/child_compassion/models/child_compassion.py
+++ b/child_compassion/models/child_compassion.py
@@ -499,7 +499,7 @@ class CompassionChild(models.Model):
                 break
             except RelationNotFound as e:
                 self.fetch_missing_relational_records(
-                    e.field_relation, e.value, e.json_name
+                    e.field_relation, e.field_name, e.value, e.json_name
                 )
 
         # Update household
@@ -520,7 +520,8 @@ class CompassionChild(models.Model):
             return {}
         return super().data_to_json(mapping_name)
 
-    def fetch_missing_relational_records(self, field_relation, values, json_name):
+    def fetch_missing_relational_records(self, field_relation,
+                                         field_name, values, json_name):
         """ Fetch missing relational records in various languages.
 
         Method used to catch missing values for household duties, hobbies,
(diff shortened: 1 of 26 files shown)


child_compassion/models/child_compassion.py
```diff
@@ -552,18 +554,22 @@ def fetch_missing_relational_records(self, field_relation, values, json_name):
...
             if hasattr(relation_obj, "value"):
+                # Useful for connect.multipicklist objects
                 search_vals.insert(0, "|")
                 search_vals.append(("value", "=", value))
             search_count = relation_obj.search_count(search_vals)
             # if not exist, then create it
             if not search_count:
                 value_record = (
-                    relation_obj.create({"name": value, "value": value})
+                    relation_obj.create({field_name: value, "value": value})
```
wvmrtn, line 567: Need to check again if relation_obj has attribute "value" if we want to create a new record with "value": value. Otherwise, create record with column "name" only.

child_compassion/models/project_compassion.py
```diff
@@ -617,18 +618,22 @@ def fetch_missing_relational_records(self, field_relation, values, json_name):
...
             if hasattr(relation_obj, "value"):
+                # Useful for connect.multipicklist objects
                 search_vals.insert(0, "|")
                 search_vals.append(("value", "=", value))
             search_count = relation_obj.search_count(search_vals)
             # if not exist, then create it
             if not search_count:
                 value_record = (
-                    relation_obj.create({"name": value, "value": value})
+                    relation_obj.create({field_name: value, "value": value})
```
wvmrtn, line 631: Same comment as same place than in child_compassion.py

child_compassion/models/child_compassion.py
ecino: This is not needed here because the record creation will work and there will only be a warning in the console telling that the field "value" was ignored. It's not like in the search where it raises an error in the search method.

PR #1382, opened 2020-07-13
CO-3361 fix datetime zip filename

quillasp:
Another datetime in filename string conversion error in file `b2s_image.py`


Theyiot (approved):
LGTM

ecino (approved):

PR #1384, opened 2020-07-14
CO-3346 Can't create events from calendar view

SylvainLosey:
I found the origin of the bug but I'm not sure on how to proceed, as the fix would require heavy refactoring.

Here is the problem:
- The Event Calendar View (id `view_events_calendar`) declares what field are used as start and stop of the event model for display in the calendar, currently `start_date` and `end_date`
- The calendar_view.js of the core web module maps the fields declared above, and calendar_controller.js uses the names of the mapping to set in the context "default_[name of the mapped field]" to the value selected when creating the event from the calendar view.
- The `calendar.event` has a computed `start_date` field. When creating an instance of a model, Odoo checks in the context for a "default_[name of the field]" (odoo/models.py l. 1153), finds the value set previously by the JS for `default_start_date`, and uses it for this field.
- As it is a computed field, the inverse method is called to set the `start` and `stop` fields, but it returns Null as this is not the expected behavior, resulting in the error that originated this task.

In short, the name used to store the start and stop datetimes (`start_date` and `end_date`) of  `crm.event.compassion` conflicts with the one used in `calendar.event` through the context set in JS. Renaming `start_date` and `end_date` to `start` and `stop` (or any other name without conflict) solves the bug. The problem is that those fields are used heavily throughout the codebase, and as the name isn't unique refactoring to a new name everywhere will be error prone. 

Therefore I would like to make sure this is the fix I should use before continuing. I do not see another obvious one as it would require the modification of base odoo modules.

### DIFF
--- a/crm_compassion/models/event_compassion.py
+++ b/crm_compassion/models/event_compassion.py
@@ -210,7 +210,7 @@ class EventCompassion(models.Model):
                 )
 
     def compute_hold_start_date(self, start=None):
-        delta = self.env["res.config.settings"].get_param("days_allocate_before_event")
+        delta = self.env["res.config.settings"].sudo().get_param("days_allocate_before_event")
         return (start if start else self.start_date.date()) - timedelta(days=delta)
 
     @api.multi
--- a/crm_compassion/models/event_compassion.py
+++ b/crm_compassion/models/event_compassion.py
@@ -114,7 +114,7 @@ class EventCompassion(models.Model):
         default=0,
     )
     planned_sponsorships = fields.Integer(
-        "Expected sponsorships", track_visibility="onchange", required=True
+        "Expected sponsorships", track_visibility="onchange", required=True, default=0
     )
     lead_id = fields.Many2one(
         "crm.lead", "Opportunity", track_visibility="onchange", readonly=False
@@ -209,6 +209,10 @@ class EventCompassion(models.Model):
                     _("The hold start date must " "be before the event starting date !")
                 )
 
+    def compute_hold_start_date(self, start=None):
+        delta = self.env["res.config.settings"].get_param("days_allocate_before_event")
+        return (start if start else self.start_date.date()) - timedelta(days=delta)
+
     @api.multi
     @api.depends("staff_ids")
     def _compute_users(self):
@@ -237,6 +241,11 @@ class EventCompassion(models.Model):
         elif event_name[-2:] == event_year[-2:]:
             vals["name"] = event_name[:-2]
 
+        # Compute hold_start_date from vals if it hasn't been set
+        if not vals.get("hold_start_date"):
+            hold_start_date = self.compute_hold_start_date(start=vals["start_date"])
+            vals["hold_start_date"] = hold_start_date
+
         event = super().create(vals)
 
         # Analytic account and Origin linked to this event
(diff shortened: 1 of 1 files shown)


SylvainLosey:
Idea for another fix, much easier but more hackish: add a compute field called start for example that returns start_date. Use it only in the XML declaration of the calendar

crm_compassion/models/event_compassion.py
```diff
@@ -393,26 +403,21 @@ def onchange_type(self):
...
-        )
-        dt = timedelta(days=days_allocate_before_event)
-        for event in self.filtered("start_date"):
-            event.hold_start_date = (event.start_date - dt).date()
-            if not event.end_date or event.end_date < event.start_date:
-                event.end_date = event.start_date
-
-    @api.onchange("end_date")
+    def onchange_start(self):
+        """ Update stop and hold_start as soon as v is changed """
```
Theyiot: Minor error but `as start is changed` instead of `as v is changed`

crm_compassion/models/event_compassion.py
```diff
@@ -114,7 +114,7 @@ class EventCompassion(models.Model):
         default=0,
     )
     planned_sponsorships = fields.Integer(
-        "Expected sponsorships", track_visibility="onchange", required=True
+        "Expected sponsorships", track_visibility="onchange", required=True, default=0
```
Theyiot, line 117: Is this modification mandatory for the fix to work ? Does it solve anything else ?

crm_compassion/models/event_compassion.py
```diff
@@ -217,6 +217,10 @@ def _compute_users(self):
                 lambda u: not u.share
             )
 
+    def compute_hold_start_date(self, start=None):
+        delta = self.env["res.config.settings"].get_param("days_allocate_before_event") or 10
```
Theyiot: Why not using directly `delta = self.env["res.config.settings"].get_values()["days_allocate_before_event"]`, which already takes care to fill in the dictionary returned by `get_values()` with the default value 10 if the parameter does not exist ?

Theyiot (changes requested):
Overall the code looks good to me, there are just a few small corrections that can be done.

SylvainLosey:
Oh sorry I should have made it clearer that the code isn't ready (WIP). I'm not sure what fix to use between renaming the discussed fields (like proposed in my first comment and started but not finished), and creating compute fields (like proposed in the second comment). I'm waiting on the input from Ema or Stéphane to know how to go forward.

ecino:
Can't you check when the `calendar.event` object is created in the stack trace and you would be able to remove "default_start_date" and "default_end_date" at that point so that it doesn't conflict in the creation?

crm_compassion/models/event_compassion.py
SylvainLosey: It is, either we need to set a default or make this fields non required, because it can be created from the calendar view that doesn't have this info

SylvainLosey:
This works too, thanks for the idea !

crm_compassion/models/event_compassion.py
```diff
@@ -209,6 +209,10 @@ def _check_hold_start_date(self):
                     _("The hold start date must " "be before the event starting date !")
                 )
 
+    def compute_hold_start_date(self, start=None):
+        delta = self.env["res.config.settings"].get_param("days_allocate_before_event")
```
ecino:
```suggestion
        delta = self.env["res.config.settings"].sudo().get_param("days_allocate_before_event")
```

ecino (changes requested):
There might be access rights issues if we access settings without sudo.

ecino (approved):

PR #1400, opened 2020-07-20
CO-3378 FIX Support replies aren't set to resolved when replied to

SylvainLosey:
Was introduced by CO-3240. Each time the write method of crm.claim was called with an user_id set, the state was set back to "waiting for support".

I deleted this code which solves the bug, but I'm not sure why it was here so it might introduce another bug. Would be good if Joel could review this.

### DIFF
--- a/crm_request/models/request.py
+++ b/crm_request/models/request.py
@@ -248,9 +248,6 @@ class CrmClaim(models.Model):
         - Push partner to associated mail messages
         """
 
-        if values.get("user_id"):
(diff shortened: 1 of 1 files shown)


ecino (commented):
It was so that when someone Assigns a new request, it goes automatically to the state "waiting support". Could you check that this still works ? If not you should do this not in the write method but specifically in the "assign" method. Or you could add a condition that it moves the state only if previous state was "new"

ecino (approved):

PR #1401, opened 2020-07-21
CO-3192 remplacement de l'état "appeler" en une activité selon les mê…

quillasp:
…mes critères

### DIFF
--- a/partner_communication/models/communication_job.py
+++ b/partner_communication/models/communication_job.py
@@ -99,7 +99,7 @@ class CommunicationJob(models.Model):
     sent_date = fields.Datetime(readonly=True, copy=False)
     state = fields.Selection(
         [
-            ("call", _("Call partner")),
+            # ("call", _("Call partner")),
             ("pending", _("Pending")),
             ("done", _("Done")),
             ("cancel", _("Cancelled")),
@@ -239,16 +239,17 @@ class CommunicationJob(models.Model):
             vals["object_ids"] = str(vals["partner_id"])
 
         same_job_search = [
-            ("partner_id", "=", vals.get("partner_id")),
-            ("config_id", "=", vals.get("config_id")),
-            (
-                "config_id",
-                "!=",
-                self.env.ref(
-                    "partner_communication.default_communication").id,
-            ),
-            ("state", "in", ("call", "pending")),
-        ] + self.env.context.get("same_job_search", [])
+                              ("partner_id", "=", vals.get("partner_id")),
+                              ("config_id", "=", vals.get("config_id")),
+                              (
+                                  "config_id",
+                                  "!=",
+                                  self.env.ref(
+                                      "partner_communication.default_communication").id,
+                              ),
+                              # ("state", "in", ("call", "pending")),
+                              ("state", "=", "pending"),
+                          ] + self.env.context.get("same_job_search", [])
         job = self.search(same_job_search)
         if job:
             job.object_ids = job.object_ids + "," + vals["object_ids"]
@@ -270,13 +271,6 @@ class CommunicationJob(models.Model):
             job.refresh_text()
         else:
             job.set_attachments()
-
-        # Check if phonecall is needed
-        if (
-                job.need_call == "before_sending"
-                or job.config_id.need_call == "before_sending"
-        ):
-            job.state = "call"
 
         if job.body_html or job.send_mode == "physical":
             job.count_pdf_page()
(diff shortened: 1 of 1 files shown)


partner_communication/models/communication_job.py
```diff
@@ -239,16 +238,16 @@ def create(self, vals):
...
+                              ("partner_id", "=", vals.get("partner_id")),
+                              ("config_id", "=", vals.get("config_id")),
+                              (
+                                  "config_id",
+                                  "!=",
+                                  self.env.ref(
+                                      "partner_communication.default_communication").id,
+                              ),
+                              ("state", "=", "pending"),
+                          ] + self.env.context.get("same_job_search", [])
```
ecino: Please check and restore the indentation

partner_communication/models/communication_job.py
```diff
@@ -378,6 +370,25 @@ def write(self, vals):
         if vals.get("body_html") or vals.get("send_mode") == "physical":
             self.count_pdf_page()
 
+        # if the call must be done after the sending, we unlink all activities
+        # associated with the job (as for the moment, there is only one activity type)
+        if self.need_call == "after_sending" and self.state == "pending":
+            for activity in self.activity_ids:
+                activity.unlink()
```
ecino: There is no need to loop through the records, you can directly call the method on the recordset.
```suggestion
            self.activity_ids.unlink()
```

partner_communication/models/communication_job.py
```diff
@@ -408,21 +419,18 @@ def send(self):
...
                 job.refresh_text()
 
+        # if the call must be done after the sending, an activity is scheduled
+        if self.need_call == "after_sending":
+            self.activity_schedule('mail.mail_activity_data_call',
+                                   summary="call_" + self.need_call,
+                                   user_id=self.user_id.id,
+                                   note="Call {} at (phone) {}"
+                                   .format(self.partner_id.name,
+                                           self.partner_phone))
```
ecino: You should move this inside the for loop and use `job` instead of `self`. This code will fail if you send multiple communications at once, because you cannot access `self.need_call` on a recordset.

partner_communication/models/communication_job.py
```diff
@@ -745,7 +741,13 @@ def _send_mail(self):
...
         if email.state == "sent":
             if self.need_call == "after_sending":
-                final_state = "call"
+                # same as in the send method
+                self.activity_schedule('mail.mail_activity_data_call',
+                                       summary="call_" + self.need_call,
+                                       user_id=self.user_id.id,
+                                       note="Call {} at (phone) {}"
+                                       .format(partner.name,
+                                               partner.phone))
```
ecino: This has to be removed I think, it is already handled in your code in the `send` method, regardless if the communication is sent by email or printed.

partner_communication/models/communication_job.py
```diff
@@ -783,7 +785,13 @@ def _print_report(self):
...
             state = "done"
             if job.need_call == "after_sending":
-                state = "call"
+                # same as in the send method
+                job.activity_schedule('mail.mail_activity_data_call',
+                                      summary="call_" + job.need_call,
+                                      user_id=job.user_id.id,
+                                      note="Call {} at (phone) {}"
+                                      .format(job.partner_id.name,
+                                              job.partner_id.phone))
```
ecino: This can be removed and should be handled directly from the `send` method.

ecino (changes requested):

partner_communication/models/communication_job.py
```diff
@@ -744,9 +739,7 @@ def _send_mail(self):
 
         final_state = "pending"
         if email.state == "sent":
-            if self.need_call == "after_sending":
-                final_state = "call"
-            else:
+            if self.need_call == "before_sending":
                 final_state = "done"
         return final_state
```
ecino: I think all this code is no more needed, because the final state will always be "done" now. It can be just:
```
return "done"
```

ecino (changes requested):

partner_communication/models/communication_job.py
```diff
@@ -378,6 +370,24 @@ def write(self, vals):
...
+        # (there's only one activity type associated with the communication job)
+        if ((self.need_call == "before_sending"
+             and self.state == "pending")
+                and not self.activity_ids):
+            self.activity_schedule('mail.mail_activity_data_call',
+                                   summary="call_" + self.need_call,
+                                   user_id=self.user_id.id,
+                                   note="Call {} at (phone) {}"
+                                   .format(self.partner_id.name,
+                                           self.partner_phone))
```
ecino: Maybe you should put all this code inside a if condition to execute it only if the `need_call` or `state` fields values are changed. Otherwise it will execute this each time we write any field values in the communication. Then you have to loop on all communications, otherwise it will fail. If you don't understand, we can discuss this today.
```python
        if "need_call" in vals or "state" in vals:
            for job in self:
                if job.need_call == "afer_sending" ....
                        ....
```

partner_communication/models/communication_job.py
ecino: Please have a look at this comment as well.

ecino (changes requested):

partner_communication/models/communication_job.py
ecino: This is not yet correct. You must have the loop in the write method.

partner_communication/models/communication_job.py
quillasp: Apparently the `final_state` can be pending if the `need_call` is "after_sending". Should I really remove this part and put "done" by default ?

partner_communication/models/communication_job.py
quillasp: This part of code is inside the send method so I'm not really sure what you mean. Should I use the `job` in `todo.filtered(lambda j: j.send_mode in ("both", "digital"))` or should I use `self` instead ? 

For this PR, I just translated the code to act as if the state call still exist, but create an activity instead. So there might be logic errors before my attempt to do the task and they'll be corrected with this PR.

partner_communication/models/communication_job.py
quillasp: Is this the write method you mentioned when you talked about the lines 422 to 429 which are in the send method ? Still, I followed your advice and check if the `need_call` and `state` fields values changed and then did a for loop inside `self`.

partner_communication/models/communication_job.py
ecino: yes

ecino (approved):

PR #1414, opened 2020-07-30
Co 3210 notify staff letter state unsuccessful

quillasp:
I'm still wondering why PyCharm indents some part of the code even though I didn't edit anything there... Hence the identation "fix".

### DIFF
--- a/sbc_compassion/models/correspondence.py
+++ b/sbc_compassion/models/correspondence.py
@@ -302,12 +302,12 @@ class Correspondence(models.Model):
         for letter in self:
             if letter.sponsorship_id and letter.communication_type_ids:
                 letter.name = (
-                    letter.communication_type_ids[0].name
-                    + " ("
-                    + letter.sponsorship_id.partner_id.ref
-                    + " - "
-                    + letter.child_id.local_id
-                    + ")"
+                        letter.communication_type_ids[0].name
+                        + " ("
+                        + letter.sponsorship_id.partner_id.ref
+                        + " - "
+                        + letter.child_id.local_id
+                        + ")"
                 )
             else:
                 letter.name = _("New correspondence")
@@ -442,8 +442,8 @@ class Correspondence(models.Model):
     def _compute_is_final_letter(self):
         for letter in self:
             letter.is_final_letter = (
-                "Final Letter" in letter.communication_type_ids.mapped("name")
-                or letter.sponsorship_state != "active"
+                    "Final Letter" in letter.communication_type_ids.mapped("name")
+                    or letter.sponsorship_state != "active"
             )
 
     @api.multi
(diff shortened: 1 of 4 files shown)


sbc_compassion/models/correspondence.py
```diff
@@ -76,6 +76,14 @@ class Correspondence(models.Model):
...
         related="sponsorship_id.child_id", store=True, readonly=False
     )
+    user_id = fields.Many2one(
+        "res.users",
+        string="Assigned to",
+        default=lambda self: self.env.user,
+        domain=[("share", "=", False)],
+        readonly=False,
+        required=True
+    )
```
ecino: You should not add this field in the model, as we don't want to assign each letter individually to a user. A lot of letters imported automatically would be assigned to the user Administrator and therefore nobody will be notified when the letter is unsucessful.

Instead, you should add a setting in the sbc_settings.py class that will set a user for every letters, and it will be read from there when we assign the notification.

ecino (changes requested):

sbc_compassion/models/correspondence.py
```diff
@@ -517,11 +517,23 @@ def create(self, vals):
     def write(self, vals):
         """ Keep track of state changes. """
         if "state" in vals:
+            if vals["state"] == "Translation check unsuccessful":
+                settings_config = self.env["res.config.settings"].create({})
```
ecino:
```suggestion
                settings_config = self.env["res.config.settings"]
```

sbc_compassion/models/correspondence.py
```diff
@@ -517,11 +517,23 @@ def create(self, vals):
...
+                for c in self:
+                    self._make_activity(c, vals["state"],
+                                        settings_config.letter_responsible.id)
```
ecino:
```suggestion
                                        settings_config.get_param("letter_responsible"))
```

sbc_compassion/models/correspondence.py
```diff
@@ -973,3 +985,14 @@ def json_to_data(self, json, mapping_name=None):
             odoo_data["page_ids"] = pages or False
 
         return odoo_data
+
+    ##########################################################################
+    #                            PRIVATE METHODS                             #
+    ##########################################################################
+
+    def _make_activity(self, correspondence, state, user_id):
+        correspondence.activity_schedule('mail.mail_activity_data_call',
```
ecino:
```suggestion
        correspondence.activity_schedule('mail.mail_activity_data_todo',
```

ecino:
```suggestion
    def _make_activity(self, state, user_id):
        self.ensure_one()
        correspondence.activity_schedule('mail.mail_activity_data_call',
```

sbc_compassion/models/correspondence.py
```diff
@@ -973,3 +985,14 @@ def json_to_data(self, json, mapping_name=None):
...
+                                         summary=state,
+                                         user_id=user_id,
+                                         note="{} has state {}"
+                                         .format(correspondence.name, state))
```
ecino:
```suggestion
                                         note=f"Letter has state {state}")
```

sbc_compassion/models/correspondence.py
ecino:
```suggestion
                    c._make_activity(
                        vals["state"], settings_config.letter_responsible.id)
```

ecino (changes requested):

ecino:
Various code suggestions, if you don't understand something please come and we can discuss it.

ecino (approved):

PR #1417, opened 2020-07-31
CO-3218 The account of the partner is considered for an automated gift

Theyiot:
Following how it were done here: https://github.com/CompassionCH/bank-payment/blob/10.0-payment-cancel/account_payment_order/models/account_payment_line.py#L108, the code now fetches the first bank account of the user and links it to the corresponding invoice.

### DIFF
--- a/sponsorship_compassion/wizards/generate_gift_wizard.py
+++ b/sponsorship_compassion/wizards/generate_gift_wizard.py
@@ -85,6 +85,7 @@ class GenerateGiftWizard(models.TransientModel):
                 invoice_date = self.invoice_date
             inv_data = self._setup_invoice(contract, invoice_date)
             invoice = self.env["account.invoice"].create(inv_data)
+            invoice.partner_bank_id = contract.partner_id.bank_ids[0].id
             invoice.action_invoice_open()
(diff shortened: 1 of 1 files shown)


sponsorship_compassion/wizards/generate_gift_wizard.py
```diff
@@ -85,6 +85,7 @@ def generate_invoice(self):
                 invoice_date = self.invoice_date
             inv_data = self._setup_invoice(contract, invoice_date)
             invoice = self.env["account.invoice"].create(inv_data)
+            invoice.partner_bank_id = contract.partner_id.bank_ids[0].id
```
ecino: This has the risk of failing if there are no bank accounts on partner (which could be the case for non LSV/DD sponsors). 
```suggestion
            invoice.partner_bank_id = contract.partner_id.bank_ids[:1].id
```

ecino (changes requested):

sponsorship_compassion/wizards/generate_gift_wizard.py
Theyiot: Nice catch ! Thank you

ecino (approved):

PR #1424, opened 2020-08-07
use planned sponsorship from futur events

davidwul:
CO-3430

retrieve planned sponsorship info from futur events.
change the field to be a computed field

### DIFF
--- a/crm_compassion/models/crm_lead.py
+++ b/crm_compassion/models/crm_lead.py
@@ -11,13 +11,14 @@
 from odoo.addons.crm.models.crm_lead import CRM_LEAD_FIELDS_TO_MERGE
 
 from odoo import api, models, fields
-
+import datetime
 
 class CrmLead(models.Model):
     _inherit = "crm.lead"
 
     planned_sponsorships = fields.Integer(
         "Expected new sponsorships", track_visibility="onchange"
+        ,compute="_compute_planned_sponsorship"
     )
     event_ids = fields.One2many(
         "crm.event.compassion", "lead_id", "Events", readonly=False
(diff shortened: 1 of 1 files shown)


crm_compassion/models/crm_lead.py
```diff
@@ -11,13 +11,14 @@
...
 from odoo import api, models, fields
-
+import datetime
 
 class CrmLead(models.Model):
     _inherit = "crm.lead"
 
     planned_sponsorships = fields.Integer(
         "Expected new sponsorships", track_visibility="onchange"
+        ,compute="_compute_planned_sponsorship"
```
ecino: If we store it, we can search with this, order and group by, which could be helpful. It will also display the kanban view faster because it won't need to compute all values when opening the view.
```suggestion
        ,compute="_compute_planned_sponsorship", store=True
```

crm_compassion/models/crm_lead.py
```diff
@@ -54,6 +55,16 @@ def create_event(self):
...
 
+    @api.multi
+    @api.depends("event_ids")
+    def _compute_planned_sponsorship(self):
+        for lead in self:
+            future_planned_sponsorships=0
+            for e in lead.event_ids:
+                if e.start_date > datetime.datetime.now():
+                    future_planned_sponsorships+=e.planned_sponsorships
+            lead.planned_sponsorships=future_planned_sponsorships
```
ecino: This is just flake8 formatting issues, that requires putting spaces around the operators.
```suggestion
            future_planned_sponsorships = 0
            for e in lead.event_ids:
                if e.start_date > datetime.datetime.now():
                    future_planned_sponsorships += e.planned_sponsorships
            lead.planned_sponsorships = future_planned_sponsorships
```

ecino (changes requested):

crm_compassion/models/crm_lead.py
```diff
@@ -54,6 +55,16 @@ def create_event(self):
             "context": context,
         }
 
+    @api.multi
+    @api.depends("event_ids")
```
ecino: Trigger recomputation as well if we change the planned sponsorships in a linked event.
```suggestion
    @api.depends("event_ids", "event_ids.planned_sponsorships")
```

ecino (changes requested):

ecino (approved):

PR #1427, opened 2020-08-17
CO-3422 Adding the last date at which a sponsor wrote a letter for a given contract

Theyiot:
In parallel to: https://github.com/CompassionCH/compassion-switzerland/pull/1217.

### DIFF
--- a/sbc_compassion/models/correspondence.py
+++ b/sbc_compassion/models/correspondence.py
@@ -12,6 +12,7 @@ import logging
 import re
 import threading
 import uuid
+from datetime import datetime
 from io import BytesIO
 
 from odoo import fields, models, api, _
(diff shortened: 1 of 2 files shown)


sbc_compassion/models/correspondence.py
```diff
@@ -12,6 +12,7 @@
 import re
 import threading
 import uuid
+from datetime import datetime
```
ecino:
```suggestion
```

sbc_compassion/models/correspondence.py
```diff
@@ -487,9 +488,11 @@ def create(self, vals):
         if vals.get("store_letter_image", True) is False:
             vals["letter_image"] = False
 
+        contract = self.env["recurring.contract"].browse(vals["sponsorship_id"])
+        contract.last_sponsor_letter = datetime.now()
```
ecino: We only want to mark supporter letters and not update the date when we receive a child letter.
```suggestion
        contract = self.env["recurring.contract"].browse(vals["sponsorship_id"])
        if vals.get("direction") == "Supporter To Beneficiary":
            contract.last_sponsor_letter = fields.Date.today()
```

ecino (changes requested):

ecino (approved):

PR #1428, opened 2020-08-18
AP-364 Resolve bug with mobile app account creation

nfluckiger:


Theyiot (approved):
Looks good to me.

mobile_app_connector/forms/registration_form.py
```diff
@@ -351,12 +351,21 @@ def form_before_create_or_update(self, values, extra_values):
...
+                    "&",
+                    ("email", "=ilike", partner_email),
+                    "|", ("active", "=", True),
+                         ("active", "=", False)
+                     ])
             )
 
+            if not matching_partner.active:
+                matching_partner.write({"email": matching_partner.contact_id.email})
+                matching_partner = matching_partner.contact_id
```
ecino, line 364: Shouldn't we here replace the principal mail address with the matched one as well? Or I don't understand what you are doing (maybe adding a comment would be helpful).
```suggestion
                matching_partner.write({"email": matching_partner.contact_id.email})
                matching_partner = matching_partner.contact_id
                matching_partner.write({"email": partner_email})
```

nfluckiger: the method update the main partner email in all case, so we don't have to do it.

PR #1433, opened 2020-08-26
Co-3443 Adding verification to avoid multiple notification

nfluckiger:


mobile_app_connector/models/account_invoice.py
```diff
@@ -146,8 +146,13 @@ def action_invoice_paid(self):
             has_app = self.env['firebase.registration'].search_count([
                 ('partner_id', '=', partner.id)
             ])
+            previous_notification = self.env["firebase.notification"].search([
+                '&', ("res_model", "=", self._name),
+                ("res_id", "=", self._uid)
```
ecino: - `&` is the default join operator.
- res_id should be the record id and not the current user id. 

```suggestion
                ("res_model", "=", self._name),
                ("res_id", "=", invoice.id)
```

mobile_app_connector/models/account_invoice.py
```diff
@@ -178,6 +183,9 @@ def send_mobile_notification(self):
                 "partner_ids": [(6, 0, partner.ids)],
                 "title": _("You gave CHF %s.- for %s") % (amount, for_text),
                 "body": _("Thank you for your generosity!"),
+                "res_model": self.env['ir.model'].name_search(
+                    self._name, operator='=')[0][0],
+                "res_id": self._uid
```
ecino:
```suggestion
                "res_model": self._name,
                "res_id": self.id
```

mobile_app_connector/models/firebase_notification.py
```diff
@@ -36,6 +36,9 @@ class FirebaseNotification(models.Model):
     fundType = fields.Many2one("product.product", "Fund product", readonly=False)
     child_id = fields.Many2one("compassion.child", "Child", readonly=False)
 
+    res_model = fields.Many2one("ir.model")
```
ecino: I  think it's more simple to store directly the name of the related model instead of the id in the table `ir.model`
```suggestion
    res_model = fields.Char()
```

mobile_app_connector/models/partner_communication.py
```diff
@@ -82,6 +82,9 @@ def send(self):
                     "destination": job.mobile_notification_destination,
                     "topic": job.mobile_notification_topic,
                     "partner_ids": [(4, job.partner_id.id)],
+                    "res_model": self.env["ir_model"].search(
+                        "name", 'like', object._name).id,
+                    "res_id": object._uid
```
ecino: I would rather link it directly to the communication instead of the related objects.
```suggestion
                    "res_model": self._name,
                    "res_id": job.id
```

ecino (changes requested):

mobile_app_connector/models/account_invoice.py
ecino: Instead of putting the logic here, you could move it inside the `create` method of `firebase.notification` so that the behaviour is global : whenever a notification is created, if one already existed attached to a same object in odoo, it won't create it.

mobile_app_connector/models/firebase_notification.py
```diff
@@ -36,9 +36,21 @@ class FirebaseNotification(models.Model):
...
     #                             PUBLIC METHODS                             #
     ##########################################################################
+    @api.model
+    def create(self, vals):
+        previous_notification = self.env["firebase.notification"].search([
+            '&', ("res_model", "=", self._name),
+            ("res_id", "=", self._uid)
+        ])
+        if previous_notification:
+            return False
```
ecino:
```suggestion
        previous_notification = self.env["firebase.notification"].search([
            ("res_model", "=", vals["res_model"]),
            ("res_id", "=", vals["res_id"])
        ])
        if previous_notification:
            return previous_notification
```

mobile_app_connector/models/firebase_notification.py
```diff
@@ -36,9 +36,21 @@ class FirebaseNotification(models.Model):
     fundType = fields.Many2one("product.product", "Fund product", readonly=False)
     child_id = fields.Many2one("compassion.child", "Child", readonly=False)
 
+    res_model = fields.Char()
```
ecino:
```suggestion
    res_model = fields.Char(required=True)
```

mobile_app_connector/models/firebase_notification.py
```diff
@@ -36,9 +36,21 @@ class FirebaseNotification(models.Model):
...
+    res_id = fields.Integer()
```
ecino:
```suggestion
    res_id = fields.Integer(required=True)
```

mobile_app_connector/models/account_invoice.py
```diff
@@ -178,6 +179,8 @@ def send_mobile_notification(self):
                 "partner_ids": [(6, 0, partner.ids)],
                 "title": _("You gave CHF %s.- for %s") % (amount, for_text),
                 "body": _("Thank you for your generosity!"),
+                "res_model": self._name,
+                "res_id": self._uid
```
ecino:
```suggestion
                "res_id": self.id
```

ecino (changes requested):

mobile_app_connector/models/firebase_notification.py
```diff
@@ -36,9 +36,21 @@ class FirebaseNotification(models.Model):
...
+    res_id = fields.Integer(required=True)
+
     ##########################################################################
     #                             PUBLIC METHODS                             #
     ##########################################################################
+    @api.model
+    def create(self, vals):
+        previous_notification = self.env["firebase.notification"].search([
+            '&', ("res_model", "=", self._name),
+            ("res_id", "=", self._uid)
```
ecino: This is still not changed according to what I asked previously.

ecino (changes requested):

mobile_app_connector/models/firebase_notification.py
```diff
@@ -36,9 +36,21 @@ class FirebaseNotification(models.Model):
...
+    res_id = fields.Integer(required=True)
+
     ##########################################################################
     #                             PUBLIC METHODS                             #
     ##########################################################################
+    @api.model
+    def create(self, vals):
+        previous_notification = self.env["firebase.notification"].search([
+            ("res_model", "=", self._name),
+            ("res_id", "=", self._uid)
```
ecino:
```suggestion
            ("res_model", "=", vals["res_model"]),
            ("res_id", "=", vals["res_id"])
```

ecino (changes requested):
Please move your changes from `mobile_app_connector/models/firebase_notification.py` to `firebase_connector/models/firebase_notification.py`

mobile_app_connector/models/firebase_notification.py
```diff
@@ -36,10 +36,12 @@ class FirebaseNotification(models.Model):
     fundType = fields.Many2one("product.product", "Fund product", readonly=False)
     child_id = fields.Many2one("compassion.child", "Child", readonly=False)
 
+    res_model = fields.Char(required=True)
+    res_id = fields.Integer(required=True)
+
```
ecino: those two fields also need to be moved in `firebase_connector` module.

ecino (changes requested):

ecino (approved):

PR #1435, opened 2020-08-27
AP-363 Implementing pagination in Odoo for application

Theyiot:
Pagination is a feature that reduces the size of the messages that are sent to the application, hence increasing the performances. There was a base of code in _Odoo_ but that required quite a few changes in order to work properly with the applications.

This _PR_ is composed of three parts:
- Modifications in _Odoo_ (here)
- [Modifications in _Android_](https://github.com/CompassionCH/compassion_uk_android/pull/126)
- [Modifications in _iOS_](https://github.com/CompassionCH/Compassion_ios/pull/119)

### DIFF
--- a/mobile_app_connector/models/app_hub.py
+++ b/mobile_app_connector/models/app_hub.py
@@ -8,8 +8,11 @@
 ##############################################################################
 import logging
 import random
+import json
 from collections import defaultdict
 from dateutil.relativedelta import relativedelta
+from dateutil.parser import parse
+import datetime
 
 from odoo import api, models, fields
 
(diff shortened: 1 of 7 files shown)


mobile_app_connector/models/app_hub.py
```diff
@@ -219,7 +240,6 @@ def _construct_hub_message(
             + "/mobile-app-api"
         )
         hub_endpoint = "/hub/{}?start={}&limit={}"
-        self._assign_order(messages)
```
ecino, line 222: Is the function still used? If not, it can be removed. But was it useful for ordering the tiles? Don't you need it anymore here?

mobile_app_connector/models/res_partner.py
```diff
@@ -27,9 +27,24 @@ class GetPartnerMessage(models.Model):
...
     ##########################################################################
+    @api.model_create_multi
+    def create(self, vals_list):
+        partners = super().create(vals_list)
+
+        for partner_id in partners:
+            app_messages = self.env["mobile.app.messages"].create({
+                "partner_id": partner_id.id
+            })
+            partner_id.app_messages = app_messages
```
ecino, line 45: Why would you create app.messages for every new partners ? A lot of them don't even have the app installed!

mobile_app_connector/models/app_hub.py
```diff
@@ -98,29 +101,48 @@ def mobile_get_message(self, partner_id, **pagination):
...
-                        self.env.ref("mobile_app_connector.tile_subtype_gi7").id,
-                    ),
-                    ("create_date", "!=", False),
-                ]
+
+        app_messages = partner.app_messages
+
+        start = int(pagination.get("start", 0))
+        limit = int(pagination.get("limit", 10))
+        if start != 0:
```
ecino, line 109: Maybe we should not have automatic creation of app.messages for every partner and just check here if messages are available:
```suggestion
        if start != 0 and app_messages:
```

ecino: Also, you don't check for last refresh date? Maybe if the date is too old it would be good to still load the hub.

mobile_app_connector/models/app_hub.py
Theyiot: Yes, it is still used. I just changed when it was called (checking it won't change the behaviour), otherwise the tiles would not be sorted when they are stored. This lead to only prayers and donations tiles at the beginning for a few pages.

mobile_app_connector/models/res_partner.py
Theyiot: Because it minimizes a lot the risk to include a bug in the code but for most of the users (as they don't have the app), only their id will be stored, so the table will still keep a really small size. But yeah, a better solution would be to create a migration script to add one for all users that utilizes the app and to add an entry in the table at the creation of such a user. Otherwise, we could add an entry when trying to access it for a user if there is none.

mobile_app_connector/models/app_hub.py
Theyiot: Yes, if we don't add to every partner an app.message entry, then this is required. We could potentially use this to create an entry for partner that does not have one yet.
For the refresh date, this can be easily implemented, but is there nothing in the app for a refresh of the hub after a certain amount of time ?

ecino (approved):

PR #1439, opened 2020-09-03
AP-363 add options for limiting hub refresh

ecino:
- Refresh only after 3 days or any key event happening
- Events include info of child fetched, new letter, new sponsorship
- Add option to force refresh from the app with a parameter
- Add option to manually force refresh with a boolean field


Theyiot (approved):
Nice additions, this is a cool improvements for the users.

PR #1443, opened 2020-09-08
CO-3219 Changing SMS request state when contract are validated manually by SDS

Theyiot:


sms_sponsorship/models/recurring_contract.py
```diff
@@ -179,6 +179,13 @@ def contract_waiting(self):
         self._post_payment_first_month()
         return True
 
+    @api.multi
+    def contract_active(self):
+        super().contract_active()
+        if self.sms_request_id:
+            self.sms_request_id.state = "step2"
```
ecino: This should be done in `contract_waiting` method instead of `contract_active`.

```suggestion
        self.mapped("sms_request_id").write({"state": "step2"})
```

ecino (changes requested):

sms_sponsorship/models/recurring_contract.py
Theyiot: Changes are done

sms_sponsorship/models/recurring_contract.py
```diff
@@ -179,6 +179,12 @@ def contract_waiting(self):
         self._post_payment_first_month()
         return True
 
+    @api.multi
+    def contract_waiting(self):
```
ecino: The method is already defined just above.

ecino (changes requested):

sms_sponsorship/models/recurring_contract.py
Theyiot: Yeah, right...

ecino (approved):

PR #1446, opened 2020-09-10
CO-3447 Removing writing_language field and adding default reading_language

Theyiot:
The field `writing_language` was not used and so is removed from now on. The field `reading_language` is now set by default to be a match between the language in which the children write and one of the spoken languages of the correspondent. If there is no such match, then the preferred language of the correspondent is chosen.
A migration script is also included in that _PR_ to fill the field `reading_language` everywhere it is possible. There are some cases where a `res.lang.compassion` record of the spoken languages has no `lang_id` (`res.lang`), which makes the match to the default language of the correspondent (a `res.lang` field) impossible. This could be done in a future task with a new migration, to make this field available for all users.

### DIFF
--- a/sbc_compassion/migrations/12.0.1.0.4/post-migration.py
+++ b/sbc_compassion/migrations/12.0.1.0.4/post-migration.py
@@ -1,0 +1,48 @@
+from openupgradelib import openupgrade
+
+
+def _update_reading_language(cr, sponsorship_id, lang):
+    cr.execute(
+        "UPDATE recurring_contract "
+        "SET reading_language = %s "
+        "WHERE id = %s" % (lang, sponsorship_id)
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if not version:
+        return
+
+    sponsorships_to_fill = env["recurring.contract"].search([
+        ("reading_language", "=", False)
+    ])
+
+    message_obj = env["gmc.message"]
+    action_id = env.ref("sponsorship_compassion.create_sponsorship").id
+
+    for sponsorship in sponsorships_to_fill:
+        correspondent_lang_id = False
+        for lang in sponsorship.correspondent_id.spoken_lang_ids:
+            if sponsorship.child_id.correspondence_language_id.id == lang.id:
+                _update_reading_language(
+                    env.cr,
+                    sponsorship.id,
+                    lang.id
+                )
+                continue
+            if sponsorship.correspondent_id.lang == lang.lang_id.code:
+                correspondent_lang_id = lang.id
+        if not sponsorship.reading_language and correspondent_lang_id:
+            _update_reading_language(
+                env.cr,
+                sponsorship.id,
+                correspondent_lang_id
+            )
+
+        message_obj.create({
+            "partner_id": sponsorship.correspondent_id.id,
+            "child_id": sponsorship.child_id.id,
+            "action_id": action_id,
+            "object_id": sponsorship.id,
+        }).process_messages()
--- a/sbc_compassion/migrations/12.0.1.0.4/post-migration.py
+++ b/sbc_compassion/migrations/12.0.1.0.4/post-migration.py
@@ -5,7 +5,7 @@ def _update_reading_language(cr, sponsorship_id, lang):
     cr.execute(
         "UPDATE recurring_contract "
         "SET reading_language = %s "
-        "WHERE id = %s" % (lang, sponsorship_id)
+        "WHERE id = %s", [lang, sponsorship_id]
     )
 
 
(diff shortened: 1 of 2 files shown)


sbc_compassion/migrations/12.0.1.0.4/post-migration.py
```diff
@@ -0,0 +1,48 @@
+from openupgradelib import openupgrade
+
+
+def _update_reading_language(cr, sponsorship_id, lang):
+    cr.execute(
+        "UPDATE recurring_contract "
+        "SET reading_language = %s "
+        "WHERE id = %s" % (lang, sponsorship_id)
```
ecino: The execute method already takes parameters for string replacement and should be used to avoid possible sql injections.
```suggestion
        "WHERE id = %s", [lang, sponsorship_id]
```

sbc_compassion/migrations/12.0.1.0.4/post-migration.py
```diff
@@ -0,0 +1,48 @@
...
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if not version:
+        return
+
+    sponsorships_to_fill = env["recurring.contract"].search([
+        ("reading_language", "=", False)
```
ecino: We can limit to child sponsorships that are not yet finished.
```suggestion
        ("reading_language", "=", False),
        ("child_id", "!=", False),
        ("state", "not in", ["terminated", "cancelled"])
```

sbc_compassion/migrations/12.0.1.0.4/post-migration.py
```diff
@@ -0,0 +1,48 @@
...
+            if sponsorship.correspondent_id.lang == lang.lang_id.code:
+                correspondent_lang_id = lang.id
+        if not sponsorship.reading_language and correspondent_lang_id:
+            _update_reading_language(
+                env.cr,
+                sponsorship.id,
+                correspondent_lang_id
+            )
+
+        message_obj.create({
```
ecino: Please exclude those for which the language wasn't fixed.

sbc_compassion/models/contracts.py
```diff
@@ -104,6 +97,32 @@ def _compute_write_for_birthday_alert(self):
...
+                )
+            if sponsorship.correspondent_id.lang == lang.lang_id.code:
+                correspondent_lang_id = lang.id
+        if not sponsorship.reading_language and correspondent_lang_id:
+            self._update_reading_language(
+                sponsorship.id,
+                correspondent_lang_id
+            )
+
+        return sponsorship
```
ecino: You could do it before the actual create to avoid multiple sql transactions and simplify the code:

```suggestion
        if "child_id" in vals and "correspondent_id" in vals and "reading_language" not in vals:
            child = self.env["compassion.child"].browse(vals["child_id"])
            correspondent = self.env["res.partner"].browse(vals["correspondent_id"])
            correspondent_lang_id = False
            for lang in correspondent.spoken_lang_ids:
                if child.correspondence_language_id == lang:
                    vals["reading_language"] = lang.id
                    break
                if correspondent.lang == lang.lang_id.code:
                    correspondent_lang_id = lang.id
            else:
                vals["reading_language"] = correspondent_lang_id
        return super().create(vals)
```

sbc_compassion/models/contracts.py
```diff
@@ -161,3 +180,13 @@ def contract_active(self):
...
+
+    ##########################################################################
+    #                            PRIVATE METHODS                             #
+    ##########################################################################
+    def _update_reading_language(self, sponsorship_id, lang):
+        self.env.cr.execute(
+            "UPDATE recurring_contract "
+            "SET reading_language = %s "
+            "WHERE id = %s" % (lang, sponsorship_id)
+        )
```
ecino: This wouldn't be necessary with my proposed variant.

ecino (changes requested):
Did you also check everywhere that the field `writing_language` is not used?

Theyiot:
> Did you also check everywhere that the field `writing_language` is not used?

Yes, it was the only usage in the entire project (except for to `.po` files that can be ignored).

sbc_compassion/migrations/12.0.1.0.4/post-migration.py
```diff
@@ -0,0 +1,51 @@
...
+    for sponsorship in sponsorships_to_fill:
+        correspondent_lang_id = False
+        for lang in sponsorship.correspondent_id.spoken_lang_ids:
+            if sponsorship.child_id.correspondence_language_id.id == lang.id:
+                _update_reading_language(
+                    env.cr,
+                    sponsorship.id,
+                    lang.id
+                )
+                continue
```
ecino, line 35: In this case you may not store the `correspondent_lang_id` value and you won't create the message update. So maybe you should add here `correspondent_lang_id = lang.id` before the `continue` statement.

ecino (changes requested):

sbc_compassion/migrations/12.0.1.0.4/post-migration.py
Theyiot: Checking for the existence of the variable `reading_language` should also do the trick without changing the rest of the code !

ecino (approved):

PR #1447, opened 2020-09-10
CO-3284 Pending state is now replaced by an activity

Theyiot:
Previously, when an unmatched partner was created, its state was put under `pending` until someone would check that it really did not exist (check here: https://github.com/CompassionCH/compassion-switzerland/pull/1229). Now, an activity is created instead and assigned to someone determined beforehand using a general setting.
A delay of one week is given to the person responsible to do the task.

### DIFF
--- a/cms_form_compassion/__manifest__.py
+++ b/cms_form_compassion/__manifest__.py
@@ -43,8 +43,7 @@
         "queue_job",  # oca_addons/queue
         "base_automation",  # source/addons
         "link_tracker",  # source/addons
-        "base_location",  # oca_addons/partner-contact
-        "message_center_compassion"
+        "base_location"  # oca_addons/partner-contact
     ],
     "data": [
         "data/ir.config_parameter.xml",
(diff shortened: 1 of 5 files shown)


cms_form_compassion/__manifest__.py
```diff
@@ -43,14 +43,16 @@
         "queue_job",  # oca_addons/queue
         "base_automation",  # source/addons
         "link_tracker",  # source/addons
-        "base_location"  # oca_addons/partner-contact
+        "base_location",  # oca_addons/partner-contact
+        "message_center_compassion"
```
ecino: I wouldn't make it depend just to place the settings view. Please put this settings in some existing view from the existing dependencies or base views.

ecino (changes requested):

cms_form_compassion/__manifest__.py
Theyiot: This is done by adding the setting in the base ones.

ecino (approved):

PR #1457, opened 2020-10-01
CO-3105 Interaction resume now considers all sent paper communications

Theyiot:
Some paper communications did not appear in _interaction resume_ previously because they were in a special state `call` even though they were printed and sent. Instead of checking the state, we now check if the database field `sent_date` is set: if so, we know for sure that the letter has been sent and should appear in the interaction resume. Otherwise, it need to be sent before appearing in the view.


ecino:
I don't know why this task was already there, but it has been corrected by removing the `call` state. The change is no longer required.

PR #1465, opened 2020-11-06
CO-3104 - notification-invoice-paid

jonathantarabbia:
the notification is only sent when the payment is already made by the partner from the app and only when there are not an existing contract

### DIFF
--- a/mobile_app_connector/models/account_invoice.py
+++ b/mobile_app_connector/models/account_invoice.py
@@ -136,7 +136,8 @@ class AccountInvoice(models.Model):
                 invoice.send_mobile_notification()
 
             # when a contract exist before than this invoice created, so the invoice.origin contains sponsorship
-            if invoice.invoice_type == 'sponsorship' and 'sponsorship' not in invoice.origin:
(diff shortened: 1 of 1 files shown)


mobile_app_connector/models/account_invoice.py
```diff
@@ -156,6 +134,30 @@ def action_invoice_paid(self):
             if invoice.invoice_type in ('gift', 'fund') and has_app \
                     and not invoice.avoid_thankyou_letter:
                 invoice.send_mobile_notification()
+
+            # when a contract exist before than this invoice created, so the invoice.origin contains sponsorship
+            if invoice.invoice_type == 'sponsorship' and 'sponsorship' not in invoice.origin:
```
ecino: I think this is the test we are looking for if we want to check only for mobile app invoices.
```suggestion
            if invoice.invoice_type == 'sponsorship' and ('android' in invoice.origin or 'ios' in invoice.origin):
```

ecino (changes requested):

ecino (approved):

PR #1478, opened 2020-11-26
CO-3543

davidwul:
Not possible to remove only emails or attendees from the calendar.event.
So I just commented the partner_id to prevent him to be added to the calendar.event.


ecino (changes requested):
Correction good, but syntax errors producing travis failure. We could also simply remove the line but maybe it's better to keep it to remember this was producing the problem (and avoid reintroducing it) ?
```
crm_compassion/models/event_compassion.py:495:13: E128 continuation line under-indented for visual indent
crm_compassion/models/event_compassion.py:495:89: E501 line too long (108 > 88 characters)
```

PR #1494, opened 2021-02-19
Country stats

davidwul:
New module that fetch country statistics from the worldbank and store it


res_country_statistics/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- coding: utf-8 -*-
+
```
ecino:
```suggestion
```

res_country_statistics/__manifest__.py
```diff
@@ -0,0 +1,65 @@
...
+    'data': [
+        'views/views.xml',
+        'data/res.country.indicator.csv',
+        'data/update_all_country_stat.xml',
+        'security/ir.model.access.csv',
+    ],
+    # only loaded in demonstration mode
+    'demo': [
+        'demo/demo.xml',
+    ],
```
ecino: This doesn't exist and should be removed.
```suggestion
```

res_country_statistics/models/models.py
```diff
@@ -0,0 +1,90 @@
...
+    def _compute_latest_stats(self):
+        indicators = self.env['res.country.indicator'].search([])
+        for record in self:
+            stat = self.env['res.country.stat'].search([('country_id', '=', record.id)])
+            country_statistics = []
+            for i in indicators:
+                res = stat.search([('indicator_id', '=', i.id)], order='year desc', limit=1)
+                if res:
+                    country_statistics.append(res.id)
+            self.statistics_ids = country_statistics
```
ecino: Better like this (more understandable + bug fix in assignation):
```suggestion
        indicators = self.env["res.country.indicator"].search([])
        for country in self:
            country_stats = self.env["res.country.stat"]
            for i in indicators:
                country_stats += self.env["res.country.stat"].search([
                    ("country_id", "=", country .id),
                    ("indicator_id", "=", i.id)
                ], order="year desc", limit=1)
            country.statistics_ids = country_stats
```

ecino (changes requested):

ecino:
@davidwul Did you see my review? :)

res_country_statistics/models/models.py
ecino: What do you think of this change?

ecino (approved):

PR #1501, opened 2021-03-11
CO-3532: Change invoice_type to invoice_category

coup1234:


sbc_compassion/models/correspondence_template.py
```diff
@@ -129,6 +129,7 @@ class CorrespondenceTemplate(models.Model):
         copy=True,
         readonly=False,
     )
+    website_published = fields.Boolean(default = True)
```
ecino: This has nothing to do with the task?

sponsorship_compassion/models/account_invoice.py
```diff
@@ -20,15 +20,16 @@ class AccountInvoice(models.Model):
...
         [
             ("sponsorship", "Sponsorship"),
             ("gift", "Gift"),
             ("fund", "Fund donation"),
             ("other", "Other"),
         ],
-        compute="_compute_invoice_type",
+        compute="_compute_invoice_category",
         store=True,
+        oldname="invoice_category"
```
ecino:
```suggestion
        oldname="invoice_type"
```

ecino (changes requested):

sbc_compassion/models/correspondence_template.py
coup1234: Yes, it has nothing to do.  I can do a commit to remove it.

ecino (approved):

PR #1503, opened 2021-03-19
CO-3582 explicitly specify state for child after release

JonathanGuerne:
When a sponsorship end while in draft state the desire target state for the child after release should be passed to the function `child_realease()`. The desire state in this case is `'N'` (consigned)

### DIFF
--- a/sponsorship_compassion/models/contracts.py
+++ b/sponsorship_compassion/models/contracts.py
@@ -592,7 +592,7 @@ class SponsorshipContract(models.Model):
                 if contract.child_id.sponsor_id == contract.correspondent_id:
                     child = contract.child_id.with_context({})
                     child.child_unsponsored()
-                    child.child_released()
+                    child.child_released('N')
         return super().unlink()
 
     ##########################################################################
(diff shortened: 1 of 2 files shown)


sponsorship_compassion/models/contracts.py
```diff
@@ -592,7 +592,7 @@ def unlink(self):
                 if contract.child_id.sponsor_id == contract.correspondent_id:
                     child = contract.child_id.with_context({})
                     child.child_unsponsored()
-                    child.child_released()
+                    child.child_released('N')
```
ecino, line 595: I think the call should simply be removed, what do you think?
```suggestion
```

JonathanGuerne: After a few test it seems removing the line doesn't work properly. 
- The hold is in the expected state ("Consignment Hold") but is not available for new sponsorships. 
- The sponsorship is fully deleted. It doesn't appear of the tree view as a cancelled sponsorship (which was the case before)

ecino: Ok but I fear that we will lose the link from the child to the hold because of this line: https://github.com/CompassionCH/compassion-modules/blob/be41600918ccdcb383f8b07b83e61953ee952b14/child_compassion/models/child_compassion.py#L674
The `hold_id` shouldn't be set to False in this case.

PR #1509, opened 2021-03-25
AP-378 add language to registration objects

JonathanGuerne:
Server changes to handle firebase registration language. 
a notification must be created with a language. on send each registration is filter with the notification language

### DIFF
--- a/firebase_connector/models/firebase_notification.py
+++ b/firebase_connector/models/firebase_notification.py
@@ -66,6 +66,11 @@ class FirebaseNotification(models.Model):
         string="Partner read status of the notification",
         readonly=True,
     )
+    language = fields.Selection([('fr_CH', 'French'),
+                                 ('de_DE', 'German'),
+                                 ('it_IT', 'Italian'),
+                                 ('en_US', 'English')], required=True)
+
     res_model = fields.Char()
     res_id = fields.Integer()
     test_mode = fields.Boolean()
@@ -142,6 +147,11 @@ class FirebaseNotification(models.Model):
                     [("partner_id", "=", False)]
                 )
 
+            # filter registration based on language
+            if notif.language:
+                registration_ids = registration_ids.filtered(
+                    lambda reg: not reg.language or reg.language == notif.language)
+
             kwargs.update({
                 "notification_id": str(notif.id),
                 "title": notif.title, "body": notif.body
@@ -160,7 +170,7 @@ class FirebaseNotification(models.Model):
                         self.env["firebase.notification.partner.read"].create([
                             {"partner_id": partner.id, "notification_id": notif.id}
                             for partner in registration_ids[i: i + split].exists()
-                            .mapped("partner_id")
+                                .mapped("partner_id")
                         ])
                         self.env.cr.commit()
                 notif.sent = status_ok
--- a/firebase_connector/models/firebase_notification.py
+++ b/firebase_connector/models/firebase_notification.py
@@ -66,11 +66,7 @@ class FirebaseNotification(models.Model):
         string="Partner read status of the notification",
         readonly=True,
     )
-    language = fields.Selection([('fr_CH', 'French'),
-                                 ('de_DE', 'German'),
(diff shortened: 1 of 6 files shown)


firebase_connector/models/firebase_notification.py
```diff
@@ -66,6 +66,11 @@ class FirebaseNotification(models.Model):
         string="Partner read status of the notification",
         readonly=True,
     )
+    language = fields.Selection([('fr_CH', 'French'),
+                                 ('de_DE', 'German'),
+                                 ('it_IT', 'Italian'),
+                                 ('en_US', 'English')], required=True)
+
```
ecino: You should use available languages instead of hardcoding them. See https://github.com/CompassionCH/compassion-modules/blob/cb9bb3102ab811a1b0f4d5ff727ccac637526ed1/crm_request/models/request.py#L37 for an example

firebase_connector/models/firebase_notification.py
```diff
@@ -142,6 +147,11 @@ def send(self, **kwargs):
                     [("partner_id", "=", False)]
                 )
 
+            # filter registration based on language
+            if notif.language:
+                registration_ids = registration_ids.filtered(
+                    lambda reg: not reg.language or reg.language == notif.language)
+
```
ecino, line 155: Why not adding the filter directly in the search method above at lines 147 and 143?

firebase_connector/models/firebase_registration.py
```diff
@@ -29,6 +29,7 @@ class FirebaseRegistration(models.Model):
     registration_id = fields.Char("Firebase Registration ID", required=True, index=True)
     partner_id = fields.Many2one("res.partner", "Partner", readonly=False)
     partner_name = fields.Char(related="partner_id.name", readonly=True)
+    language = fields.Char()
```
ecino, line 32: Please make it also visible in the views

mobile_app_connector/models/firebase_registration.py
```diff
@@ -125,8 +125,14 @@ def mobile_register(self, json_data, **params):
         firebase_id = params["firebaseId"]
         operation = params["operation"]
         partner_id = params.get("supId", None)
+        language = params.get("language", None)
```
ecino: Better yet, for one line assignation:
```suggestion
        language = params.get("language") or None
```
You can do the same change for `partner_id` field.

mobile_app_connector/models/firebase_registration.py
```diff
@@ -125,8 +125,14 @@ def mobile_register(self, json_data, **params):
...
+
         if partner_id == "":
             partner_id = None
+
+        if language == "":
+            language = None
+
```
ecino:
```suggestion
```

ecino (changes requested):

firebase_connector/models/firebase_notification.py
JonathanGuerne: Simply because there is less repetition and I also find it a bit clearer when handling notification with no language specified (for example old notification still running when the changes will be published). Same goes for handling registration with no set language. 

But I am open to the idea of doing it differently

mobile_app_connector/models/firebase_registration.py
JonathanGuerne: Thanks for the tip. I'm curious, how is it better actually ? more efficient computing vise ?

ecino: It's not more efficient but it saves you some lines after (see my comment below for lines 132-135 that you can remove completely.

firebase_connector/models/firebase_notification.py
```diff
@@ -66,6 +66,7 @@ class FirebaseNotification(models.Model):
         string="Partner read status of the notification",
         readonly=True,
     )
+    language = fields.Selection("_get_lang")
```
ecino, line 69: Well, `_get_lang` should be a method defined in your class. I think this alone won't work.

ecino (approved):

PR #1514, opened 2021-04-20
CO-3613, CO-3531 : Pass payment to draft so that it can be reconcile …

coup1234:
…if a sponsorship is terminated.

- auto_cancel for payment on the website

### DIFF
--- a/cms_form_compassion/models/invoice.py
+++ b/cms_form_compassion/models/invoice.py
@@ -1,0 +1,15 @@
+##############################################################################
+#
+#    Copyright (C) 2021 Compassion CH (http://www.compassion.ch)
+#    @author: Robin Berguerand <robin.berguerand@gmail.com>
+#
+#    The licence is in the file __manifest__.py
+#
+##############################################################################
+from odoo import models, fields
+
+
+class AccountInvoice(models.Model):
+    _inherit = "account.invoice"
+    auto_cancel_no_transaction = fields.Boolean(default=False, help='If true, cancel the invoice if the linked payment '
+                                                        'transaction is cancelled',oldname='cancel_invoice')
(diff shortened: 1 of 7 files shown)


ecino (changes requested):
Changes from `website_compassion` need to be moved in `cms_form_compassion` module.

cms_form_compassion/models/invoice.py
```diff
@@ -0,0 +1,15 @@
...
+#    The licence is in the file __manifest__.py
+#
+##############################################################################
+from odoo import models, fields
+
+
+class AccountInvoice(models.Model):
+    _inherit = "account.invoice"
+    auto_cancel_no_transaction = fields.Boolean(default=False, help='If true, cancel the invoice if the linked payment '
+                                                        'transaction is cancelled',oldname='cancel_invoice')
```
ecino, line 15: It doesn't look like "cancel_invoice" was an existing field. We should either remove this old field if we find it or remove this attribute.
```suggestion
                                                        'transaction is cancelled')
```

sponsorship_compassion/models/move_line.py
```diff
@@ -35,16 +35,17 @@ def split_payment_and_reconcile(self):
...
+                    _(
                     "This can only be done if one move line can be split "
                     "to cover the reconcile difference"
+                    )
                 )
-            )
 
         # Edit move in order to split payment into two move lines
         move.button_cancel()
+        move_line.payment_id.action_draft()
```
ecino, line 48: Test and verify if this corrects CO-3613, and if we should restore the payment state after move.write or move.post.

ecino: After some more talk and testing we confirmed that this is an issue and would leave the payment in draft state after the process.
The problem is that we cannot delete payments if they were already posted, and we cannot cancel them neither otherwise it deletes the associated move.

One solution would be to still cancel it and create another move replacing it. We must however test that this doesn't break the rest of the process.

PR #1515, opened 2021-04-20
CO-3615: Change partner_birthdate to partner_birthdate_date so that i…

coup1234:
…t matchs res_partner field.


ecino (commented):
Did you test the changes, does it solve our issues?

coup1234:
Yes, I fill the form "MuskathlonRegistrationForm"  and then check in the backend that the birthdate is correctly set. There is a pull resquest in compassion_switzerland as well.

PR #1523, opened 2021-05-07
CO-3601 check expiration date is not reach on write + tests

JonathanGuerne:


child_compassion/models/compassion_hold.py
```diff
@@ -205,6 +205,9 @@ def create(self, vals):
 
     @api.multi
     def write(self, vals):
+        if self.expiration_date < datetime.now() and "expiration_date" in vals:
+            raise UserError(_("The expiration date as been reach and thus can't be changed."))
```
ecino:
```suggestion
        if "expiration_date" in vals and self.filtered(lambda h: h.expiration_date < datetime.now()):
            raise UserError(_("The expiration date as been reach and thus can't be changed."))
```

ecino (changes requested):

ecino (approved):

PR #1524, opened 2021-05-07
CO-3600 add 1 week to hold expiration date if sms request cancelled

JonathanGuerne:
- if sms request is cancelled 
- if sms request change child


ecino (changes requested):
Unfortunately here you apply the change for every sms requests. But it can happen that children are not taken from the website. We want to do this only if the child was taken from the website and put back on the website. Thus, the change should be made in compassion-switzerland modules.

PR #1526, opened 2021-05-25
Fix payment cancel

ecino:
Extract CO-3613 from #1514 
Move code into recurring_contract and small other fixes around write method in contracts.


sponsorship_compassion/models/contracts.py
```diff
@@ -535,20 +535,22 @@ def write(self, vals):
 
         super().write(vals)
 
-        try:
-            if not testing:
-                # We want to avoid having dangling sponsorships
-                self.env.cr.commit()  # pylint: disable=invalid-commit
-            if updated_correspondents:
+        if updated_correspondents and not testing:
```
JonathanGuerne, line 538: would this be an issue If we add a new test case related to correspondents change ?

ecino: No this commit doesn't really change the behaviour, we only make sure that if we sent some info to GMC, we are in sync in Odoo and we don't lose this information in case the following code fails.

JonathanGuerne (approved):

PR #1528, opened 2021-05-27
CO-3640 move commitment_number update on write + tests

JonathanGuerne:
**Insight**
- commitment number are doubled with a couple of partner
- such mistake are rare but does still appear
- after investigation they all have one thing in common : change of partner and correspondent
- commitment number was updated trough an `onchange` function

**proposition**
- change the `onchange`call (with odoo api) to be call, if needed, on the `write`
  -  More reliable (not only call on user interface changes)


ecino (approved):

ecino:
Can you fix the conflicts please?

PR #1535, opened 2021-06-17
CO-3661 cron to send notification on expiring intervention

JonathanGuerne:
- create a new cron
- called daily on intervention
- if expiration of one of the intervention is tomorrow and state is "On Hold" will send a notif


intervention_compassion/models/compassion_intervention.py
```diff
@@ -849,6 +850,20 @@ def json_to_data(self, json, mapping_name=None):
             json["ICP"] = json["ICP"].split("; ")
         return super().json_to_data(json, mapping_name)
 
+    @api.model
+    def send_notif_before_expiration(self):
+        threshold_date = fields.datetime.today() + relativedelta(days=1)
```
ecino: 1 day only is risky, if it's the weekend or that person is not working that day... it should be at least 7 days.

intervention_compassion/models/compassion_intervention.py
```diff
@@ -849,6 +850,20 @@ def json_to_data(self, json, mapping_name=None):
...
+        for intervention in self.filtered(lambda x: x.state == "on_hold" and x.expiration_date <= threshold_date):
+            for user in intervention.user_id:
```
ecino: `user_id` is a Many2one field, you don't need a loop here.

ecino (changes requested):

ecino (changes requested):
Instead of doing a CRON, you can do a Base Automation which supports creating an activity. Then you don't need any code method to perform the action.

intervention_compassion/models/compassion_intervention.py
```diff
@@ -10,6 +10,7 @@
 
 import logging
 import time
+from dateutil.relativedelta import relativedelta
```
ecino: This is no more needed.

ecino (approved):

PR #1536, opened 2021-06-18
CO-3669 add lang to thank_letter_config

JonathanGuerne:
- add lang field
- if lang is specified then config should match partner's lang
- if lang is not specified it will act as a default for the related min amount
- config are sorted to prevent config with non-specified lang to override config with lang 
   - lang = False is always the first

### DIFF
--- a/thankyou_letters/models/thankyou_config.py
+++ b/thankyou_letters/models/thankyou_config.py
@@ -39,7 +39,9 @@ class ThankYouConfig(models.Model):
         # Cover the case where the total_amount is smaller that all min_
         # donation amount.
 
-        filtered_config = self.filtered(lambda x: x.lang is False or invoice_lines.partner_id.lang == x.lang)
+        filtered_config = self.filtered(
+            lambda x: x.lang is False or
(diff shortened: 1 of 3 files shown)


thankyou_letters/models/thankyou_config.py
```diff
@@ -36,9 +38,12 @@ def for_donation(self, invoice_lines):
         assert len(self) > 0, "There should be at least one Thank you configuration."
         # Cover the case where the total_amount is smaller that all min_
         # donation amount.
-        config = self[0]
+
+        filtered_config = self.filtered(lambda x: x.lang is False or invoice_lines.partner_id.lang == x.lang)
```
ecino: If there are multiple invoice_lines this line will fail.
```suggestion
        filtered_config = self.filtered(
            lambda x: x.lang is False or
            invoice_lines.mapped("partner_id").lang == x.lang)
```

ecino (changes requested):

ecino (approved):

PR #1537, opened 2021-06-18
CO-3675 revert default correspondent feature + tests

JonathanGuerne:
(+ clean left tagged in test files )

### DIFF
--- a/sponsorship_compassion/models/contracts.py
+++ b/sponsorship_compassion/models/contracts.py
@@ -501,6 +501,9 @@ class SponsorshipContract(models.Model):
 
         new_sponsorship = super().create(vals)
 
+        if new_sponsorship.partner_id and not new_sponsorship.correspondent_id:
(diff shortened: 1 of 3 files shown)


sponsorship_compassion/models/contracts.py
```diff
@@ -501,6 +501,9 @@ def create(self, vals):
 
         new_sponsorship = super().create(vals)
 
+        if new_sponsorship.partner_id and not new_sponsorship.correspondent_id:
+            new_sponsorship.correspondent_id = new_sponsorship.partner_id
+
```
ecino: This was previously done in an `onchange` method, which I think is better, so that the user sees before saving that the correspondent is set. Otherwise he will think the correspondent is empty and will manually set it, which is what we want to avoid.

ecino (changes requested):

ecino (approved):

PR #1542, opened 2021-06-30
CO-3679 add a cron to remind user of unaddressed support request

JonathanGuerne:
Use activity schedule to send a mail to user with unaddressed support issue
cron is executed each week

### DIFF
--- a/crm_request/models/request.py
+++ b/crm_request/models/request.py
@@ -310,3 +310,26 @@ class CrmClaim(models.Model):
                 force_send=True,
                 email_values={"email_to": request.email_origin},
             )
+
+    @api.model
+    def cron_reminder_request(self):
+        """ Periodically sends a reminder to unaddressed request (new, waiting for support)"""
+
+        new_stage_id = self.env.ref("crm_claim.stage_claim1").id
+        wait_support_stage_id = self.env.ref("crm_request.stage_wait_support").id
+
+        request_to_notify = self.search([
+            ("stage_id", "in", [new_stage_id, wait_support_stage_id]),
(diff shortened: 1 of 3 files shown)


crm_request/models/request.py
```diff
@@ -310,3 +310,26 @@ def send_holiday_answer(self):
...
+    @api.model
+    def cron_reminder_request(self):
+        """ Periodically sends a reminder to unaddressed request (new, waiting for support)"""
+
+        new_stage_id = self.env.ref("crm_claim.stage_claim1").id
+        wait_support_stage_id = self.env.ref("crm_request.stage_wait_support").id
+
+        request_to_notify = self.search([
+            ("stage_id", "in", [new_stage_id, wait_support_stage_id]),
+            ("user_id", "!=", False)
```
eicher31: To avoid sending duplicate notification the same day, I would filter request by older than some number day (setting in the module configuration).

crm_request/models/request.py
```diff
@@ -310,3 +310,26 @@ def send_holiday_answer(self):
...
+
+        for req in request_to_notify:
+            req.activity_schedule(
+                "mail.mail_activity_data_todo",
+                summary=_("A support request require your attention"),
+                note=_("The request {} you were assigned to requires"
+                       " your attention.".
+                       format(req.code)
+                       ),
+                user_id=req.user_id.id
```
eicher31, line 336: Is the mail template adding URL link to the request? If not, I think it would be a good feature to add it.

eicher31 (changes requested):

eicher31 (approved):

PR #1544, opened 2021-06-30
CO-3692 test if the hold is already expired

JonathanGuerne:
- already expired hold were sometime trigger by gmc call
- if the hold is release the child will be released to

### DIFF
--- a/child_compassion/models/compassion_hold.py
+++ b/child_compassion/models/compassion_hold.py
@@ -420,6 +420,13 @@ class CompassionHold(models.Model):
     def beneficiary_hold_removal(self, commkit_data):
         data = commkit_data.get("BeneficiaryHoldRemovalNotification")
         hold = self.search([("hold_id", "=", data.get("HoldID"))])
+
+        # avoid realising a hold (and related child) that has already been released
+        if hold and hold.state == "expired":
+            logger.warning("Received Beneficiary Hold Removal order from GMC for already expired hold.")
+            # return empty list because no hold were created or modified
+            return []
+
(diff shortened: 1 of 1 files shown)


child_compassion/models/compassion_hold.py
```diff
@@ -420,6 +420,13 @@ def postpone_no_money_cron(self):
     def beneficiary_hold_removal(self, commkit_data):
         data = commkit_data.get("BeneficiaryHoldRemovalNotification")
         hold = self.search([("hold_id", "=", data.get("HoldID"))])
+
+        # avoid realising a hold (and related child) that has already been released
+        if hold and hold.state == "expired":
+            logger.warning("Received Beneficiary Hold Removal order from GMC for already expired hold.")
+            # return empty list because no hold were created or modified
+            return []
```
ecino: I think returning the hold id would actually let us open the hold from the message which can be handy. It doesn't change anything else on the logic I think.
```suggestion
            return hold.ids
```

ecino (changes requested):

child_compassion/models/compassion_hold.py
JonathanGuerne: I didn't knew about this feature so yes good suggestion. 
I will have to add a commit on compassion-switzerland because if the super() call return some hold it might create the unplanned exit communication (that's why I returned an empty list)

JonathanGuerne:
Changes on compassion_switzerland are more complex than I expected

ecino (approved):

PR #1551, opened 2021-07-26
CO-3693 add warning message when no child were found with given criteria

JonathanGuerne:
- raise an error in the controller if no child that could match the criteria where found
- if and error was raise by the controller it displays a warning message to the user for 5seconds

- new sponsor button is removed
  - it only clear name, last name and email field. it doesn't seem very useful
  -  the name of the button is confusing
    - the button could be kept but with an other text (like "clear")

### DIFF
--- a/sms_sponsorship/webapp/src/i18n.js
+++ b/sms_sponsorship/webapp/src/i18n.js
@@ -18,7 +18,7 @@ i18n
                     "waitingForChild": "we are looking for a child waiting for a sponsor, please wait a few seconds...",
                     "waitingForOtherChild": "we are looking for a child matching your request, please wait a few seconds...",
                     "ageYears": " years",
-                    "notMatchCriteria": "Could not found a child with given criteria.",
+                    "notMatchCriteria": "Could not find a child with given criteria.",
                     // ChildCard.js
                     "cardTitle": "Sponsor a child",
                     "age": "Age",
@@ -94,7 +94,7 @@ i18n
                     "waitingForChild": "Merci de patienter quelques secondes pendant que nous recherchons un enfant en attente d'un parrain...",
                     "waitingForOtherChild": "Nous recherchons un enfant qui correspond à vos critères. Merci de patienter quelques secondes...",
                     "ageYears": " ans",
-                    "notMatchCriteria": "Could not found a child with given criteria.",
+                    "notMatchCriteria": "Impossible de trouver un enfant avec les critères donnés.",
                     // ChildCard.js
                     "cardTitle": "Parrainer un enfant",
                     "age": "Age",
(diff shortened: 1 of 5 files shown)


sms_sponsorship/webapp/src/i18n.js
```diff
@@ -35,7 +36,6 @@ i18n
                     "emailConfirm": "Confirm your email address",
                     "sponsorshipPlus": "Sponsorship plus",
                     "otherChild": "Other child",
-                    "newSponsor": "New sponsor",
```
ecino, line 38: Why is this string removed?

sms_sponsorship/webapp/src/i18n.js
```diff
@@ -94,6 +94,7 @@ i18n
                     "waitingForChild": "Merci de patienter quelques secondes pendant que nous recherchons un enfant en attente d'un parrain...",
                     "waitingForOtherChild": "Nous recherchons un enfant qui correspond à vos critères. Merci de patienter quelques secondes...",
                     "ageYears": " ans",
+                    "notMatchCriteria": "Could not found a child with given criteria.",
```
ecino: This should be in French.

sms_sponsorship/webapp/src/i18n.js
```diff
@@ -164,6 +164,7 @@ i18n
                     "waitingForChild": "Bitte warte ein paar Sekunden, wir suchen ein Kind, das auf einen Paten wartet ...",
                     "waitingForOtherChild": "Wir suchen ein Kind, das zu dir passt, bitte warte ein paar Sekunden ...",
                     "ageYears": " Jahre",
+                    "notMatchCriteria": "Could not found a child with given criteria.",
```
ecino: Could you enter the translations please? Deepl should work fine for that (maybe check German with Andre).

sms_sponsorship/webapp/src/i18n.js
```diff
@@ -234,6 +234,7 @@ i18n
                     "waitingForChild": "Stiamo cercando un bambino in attesa di un sostenitore, per favore aspetta qualche secondo ...",
                     "waitingForOtherChild": "Stiamo cercando un bambino che corrisponda alla tua richiesta, per favore aspetta qualche secondo ...",
                     "ageYears": " anni",
+                    "notMatchCriteria": "Could not found a child with given criteria.",
```
ecino: Use deepl to put it in Italian.

ecino (changes requested):

sms_sponsorship/webapp/src/i18n.js
JonathanGuerne: This text was related to the "new sponsor" button that I propose to remove
it only reset the content of the form. If you found it mandatory I would suggest to change the button's text as it is a bit confusing imo

ecino: OK

ecino (approved):

PR #1559, opened 2021-08-10
CO-3646 Add field for date when we can answer requests

ecino:
- Include a message field to be used in the template so that people can edit the message
  directly from the holiday object.

### DIFF
--- a/crm_request/models/holiday_closure.py
+++ b/crm_request/models/holiday_closure.py
@@ -17,6 +17,7 @@ class HolidayClosure(models.Model):
     _name = "holiday.closure"
     _description = "Holiday closure"
     _inherit = "translatable.model"
+    _rec_name = "holiday_name"
 
     start_date = fields.Date("Start of holiday", required=True)
     end_date = fields.Date("End of holiday", required=True)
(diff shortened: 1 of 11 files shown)


ecino:
@eicher31 Do you think this change would solve as well task CO-2931 which requested that we can change the template? Maybe I can just add the user which will sign the mail.

crm_request/models/holiday_closure.py
```diff
@@ -15,10 +17,34 @@ class HolidayClosure(models.Model):
...
+
+    start_date = fields.Date("Start of holiday", required=True)
+    end_date = fields.Date("End of holiday", required=True)
+    reply_date = fields.Date("Date of first reply")
+    holiday_name = fields.Char("Name of holiday", required=True, translate=True)
+    holiday_message = fields.Html(
+        default=lambda s: s._default_message(), translate=True, sanitize=False,
+        help="Use ${holiday_name}, ${start_date}, ${end_date} and ${reply_date} "
+             "to replace in the text with the holiday name, start date, end date "
+             "or the date at which we will be able to answer again.")
```
eicher31, line 32: Should add an example of the display date, so that they can adapt the text.

eicher31 (approved):

eicher31:
> @eicher31 Do you think this change would solve as well task CO-2931 which requested that we can change the template? Maybe I can just add the user which will sign the mail.

How do you define the user who respond? I didn't find it in the template.

eicher31 (approved):

PR #1561, opened 2021-08-11
CS-503 update sponsorship number if partner.church_id changes

JonathanGuerne:


sponsorship_compassion/models/res_partner.py
```diff
@@ -257,6 +257,10 @@ def write(self, vals):
         if "firstname" in vals and "preferred_name" not in vals:
             vals["preferred_name"] = vals["firstname"]
         res = super().write(vals)
+
+        if "church_id" in vals:
+            self.update_number_sponsorships()
```
ecino: Wouldn't we need to update the church number rather than itself?
```suggestion
            self.mapped("church_id").update_number_sponsorships()
```

JonathanGuerne: `update_number_sponsorship` takes care of that (with an override). 
But it could be a bit quicker and easier to understand. I will apply those changes

ecino (approved):

PR #1562, opened 2021-08-13
CO-3148 fix issues w/ hr attendance

JonathanGuerne:
- remove break of attendance unlink if necessary 
- Test and correct attence from old period modification
- Test attendance day w/ no calendar
- Reactive old test


eicher31 (approved):
Ok for me!

PR #1566, opened 2021-08-17
CO-3667 FIX Claim answer quote for alias contacts

ecino:
It seems to work but I would be glad if you can verify that the code change doesn't breakup other cases. Thanks.


eicher31:
Seem to be ok!

PR #1568, opened 2021-08-27
CS-525 Fix wrong saluations in support email

JCompassion:


crm_request/models/mail_compose_message.py
```diff
@@ -40,7 +40,7 @@ def onchange_template_id(self, template_id, composition_mode, model, res_id):
         """
         Append the quote of previous e-mail to the body of the message.
         """
-        result = super().onchange_template_id(
+        result = super(MailComposer, self.with_context(lang=self.env.context.get('salutation_language'))).onchange_template_id(
```
ecino:
```suggestion
        result = super(MailComposer, self.with_context(
            lang=self.env.context.get('salutation_language', self.env.lang))
        ).onchange_template_id(
```

ecino (changes requested):

ecino (approved):

PR #1570, opened 2021-09-06
CO-3740 CO-3760 FIX GMC Mappings to prevent data loss

ecino:
- Allows to perform clever searches into relational records when
  receiving a set of values for the relation
- Allows to attach new relations instead of always replacing all of them
- Fixes disaster alerts mappings


eicher31 (approved):
:+1:

JCompassion (approved):

PR #1572, opened 2021-09-08
CO-3739 Change OAuth Provider for GMC

JCompassion:
When this commit is included in the codebase, the odoo.conf file have to
be updated otherwise message to GMC will fail. See related Jira task for
modifications to be applied.

### DIFF
--- a/message_center_compassion/models/ir_http.py
+++ b/message_center_compassion/models/ir_http.py
@@ -20,8 +20,10 @@ _logger = logging.getLogger(__name__)
 
 try:
     from jwt import JWT, jwk
-except ImportError:
+    from jwt.exceptions import JWTDecodeError
+except ImportError as e:
     _logger.error("Please install python jwt")
+    raise e
 
 
 class IrHTTP(models.AbstractModel):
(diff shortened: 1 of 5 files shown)


message_center_compassion/tools/onramp_connector.py
```diff
@@ -124,12 +124,9 @@ def get_gmc_token(cls):
                     "connect_token_server in your Odoo configuration file."
                 )
             )
-        params_post = "grant_type=client_credentials&scope=read+write"
+        params_post = "grant_type=client_credentials&scope=connect/read-write"
```
ecino, line 127: Will this be retro-compatible?

message_center_compassion/models/ir_http.py
```diff
@@ -20,8 +20,10 @@
 
 try:
     from jwt import JWT, jwk
-except ImportError:
+    from jwt.exceptions import JWTDecodeError
+except ImportError as e:
     _logger.error("Please install python jwt")
+    raise e
```
ecino, line 26: If you raise the error we won't be able to run odoo even when the module is not installed I think.

JCompassion: I though that since communication with the US is a hard requirement, it should be good to let Odoo crash if the requirements are not met instead of letting it run. It's up to you :)

ecino (approved):

PR #1574, opened 2021-09-10
CS-493 Emails info@ handled when employee in holiday

JCompassion:
When an employee gets a reply from a partner during his/her holidays the request is set to new and unassigned to this employee so that someone else can take it on.

Added a new template for fast reply to inform partner of the situation.
/!\ This template needs to be translated!

### DIFF
--- a/crm_request/__manifest__.py
+++ b/crm_request/__manifest__.py
@@ -7,7 +7,7 @@
     "license": "AGPL-3",
     "website": "http://www.compassion.ch",
     "category": "CRM",
-    "version": "12.0.1.1.0",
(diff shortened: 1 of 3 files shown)


JCompassion:
Strange that @ecino's request is included in this PR.. My branch was up-to-date. Does it matter and shall I do something about it?

ecino:
That's nothing to worry about. It's because you created your branch from the branch 12.0 I think and this merge is not yet included in the devel branch. You can leave it like this, it won't affect anything.

ecino (approved):

PR #1576, opened 2021-09-28
CS-523 Update French Text

duvenhage:
- replace "centre d'accueil" with "centre de développement de l’enfant", singular and plural


ecino (approved):
The change is fine but the existing translations are not overwritten in the production database, even when we change the .po files.

We should still apply a query to update the database, for example:

```
update ir_translation
set value = replace(value, 'centre d''accueil', 'centre de développement de l'enfant')
where value like '%centre d''accueil%'
```

PR #1585, opened 2021-11-19
CO-3827 : sick leave approved is not linked to the attendance-day

TheRaphael0000:
fix the _attendence_intervals call + change the attendance day update

### DIFF
--- a/hr_attendance_management/models/hr_holidays.py
+++ b/hr_attendance_management/models/hr_holidays.py
@@ -37,7 +37,7 @@ class HrLeave(models.Model):
                 ]
             )
 
-            rd.sudo().attendance_day_ids = att_days
+            rd.attendance_day_ids = att_days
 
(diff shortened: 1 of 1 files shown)


hr_attendance_management/models/hr_holidays.py
```diff
@@ -37,7 +37,7 @@ def _compute_att_day(self):
                 ]
             )
 
-            rd.sudo().attendance_day_ids = att_days
+            self.attendance_day_ids = att_days
```
ecino: I don't think this change is valid. Why would you change the whole recordset at each iteration? Can you test if your fix works by reverting this change and only keeping the next at line 55? Otherwise please explain me what is changed here.

TheRaphael0000:
Seems that only the sudo() call was causing the issue

ecino (approved):

PR #1590, opened 2021-12-06
CO-3834: Changing correspondent of sponsorship doesn't work

TheRaphael0000:
- Refactor the workflow to handle each record one at the time
- Fix different issues in the current code

### DIFF
--- a/sponsorship_compassion/models/contracts.py
+++ b/sponsorship_compassion/models/contracts.py
@@ -535,6 +535,8 @@ class SponsorshipContract(models.Model):
         if "parent_id" in vals:
             self.mapped("parent_id").write({"sub_sponsorship_id": False})
 
+        super().write(vals)
+
         if "reading_language" in vals:
             (self - updated_correspondents)._on_language_changed()
 
@@ -559,7 +561,7 @@ class SponsorshipContract(models.Model):
             self.on_change_partner_correspondent_id()
             self.auto_correspondent_id()
 
-        return super().write(vals)
+        return True
 
     @api.multi
     def unlink(self):
--- a/sponsorship_compassion/models/contracts.py
+++ b/sponsorship_compassion/models/contracts.py
@@ -10,6 +10,7 @@
 
 import logging
 import os
+import json
 from datetime import datetime, date
 
 from dateutil.relativedelta import relativedelta
@@ -526,7 +527,6 @@ class SponsorshipContract(models.Model):
                 if record.global_id:
                     old_correspondents = record.correspondent_id
                     record._remove_correspondent(old_correspondents)
-                    self.correspondent_id.update_number_sponsorships()
 
             for child in self.mapped("child_id"):
                 child.child_sponsored(vals["correspondent_id"])
@@ -534,8 +534,6 @@ class SponsorshipContract(models.Model):
         # Change the sub_sponsorship_id value in the previous parent_id
         if "parent_id" in vals:
             self.mapped("parent_id").write({"sub_sponsorship_id": False})
-
-        super().write(vals)
 
         if "reading_language" in vals:
             (self - updated_correspondents)._on_language_changed()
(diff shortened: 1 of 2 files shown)


TheRaphael0000:
The sponsorship cancellation seem to work
But I couldn't test the upsert since GMC return me errors that I don't understand (may be due to my development environment) : 

```
{
    "Beneficiary_GlobalID": "09201919",
    "Code": 5000,
    "Commitment_ID": null,
    "CorrespondentCommitmentID": null,
    "CorrespondentSupporterGlobalID": null,
    "HoldID": "",
    "Message": "Beneficiary Hold order not matched",
    "SponsorSupporterGlobalID": "06450834",
    "StartDate": null
}
```

sponsorship_compassion/models/contracts.py
```diff
@@ -1110,98 +1098,65 @@ def _on_language_changed(self):
...
+                    "object_id": self.id,
                 }
             )
-        messages.process_messages()
-        for i in range(0, len(messages)):
-            if messages[i].state == "success":
-                cancelled_sponsorships += sponsorships[i]
+        message.process_messages()
+
+        import json
```
ecino: Imports should all be at the beginning of the file.

sponsorship_compassion/models/contracts.py
```diff
@@ -522,10 +522,12 @@ def write(self, vals):
...
         if "correspondent_id" in vals:
-            old_correspondents = self.mapped("correspondent_id")
-            updated_correspondents = self._on_change_correspondant(
-                vals["correspondent_id"]
-            )
+            for record in self:
+                if record.global_id:
+                    old_correspondents = record.correspondent_id
+                    record._remove_correspondent(old_correspondents)
+                    self.correspondent_id.update_number_sponsorships()
```
ecino: This should be moved inside the `_remove_correspondent` method, just before the commit.

sponsorship_compassion/models/contracts.py
```diff
@@ -1110,98 +1098,65 @@ def _on_language_changed(self):
...
+            {
+                "action_id": create_action.id,
+                "child_id": self.child_id.id,
+                "partner_id": new_correspondents.id,
+                "object_id": self.id,
+            }
+        )
+        message.process_messages()
+        if not (message.state == "success"):
+            self.message_post(body=message.failure_reason, subject=_("The sponsorship is no more active!"))
```
ecino: This is not useful because the raise just after will delete this message (by rolling back any changes).

sponsorship_compassion/models/contracts.py
```diff
@@ -560,8 +545,11 @@ def write(self, vals):
             old_partners.update_number_sponsorships()
 
         if "correspondent_id" in vals:
-            self.mapped("correspondent_id").update_number_sponsorships()
-            old_correspondents.update_number_sponsorships()
+            new_correspondents = vals["correspondent_id"]
+            new_correspondents = self.env["res.partner"].browse(new_correspondents)
+            for record in self:
+                record._add_correspondent(new_correspondents)
+                record.correspondent_id.update_number_sponsorships()
```
ecino: You should move this inside `_add_correspondent` method, just before the commit.

ecino (changes requested):

sponsorship_compassion/models/contracts.py
```diff
@@ -574,7 +559,7 @@ def write(self, vals):
             self.on_change_partner_correspondent_id()
             self.auto_correspondent_id()
 
-        return True
+        return super().write(vals)
```
ecino: Now you changed the order of execution by moving the super call at the end. It means that the changes in the sponsorship will only be processed after all our code which in my opinion is wrong for the other methods that were called after the super like `_on_language_changed` and `update_number_sponsorships` at line 542. Those two should still be after the super.

ecino (changes requested):

ecino (approved):

PR #1592, opened 2021-12-13
CO-3834 Improvements

ecino:
- Add state "odoo_failure" in GMC messages in order to differentiate errors at our side
  and being able to replay the message without calling back GMC if we already received
  a success from them.
- Change further the procedure of changing correspondent to make sure we can replay the failed
  messages at any step.
- Other fixes regarding data.


TheRaphael0000 (approved):
I didn't test it since it wouldn't work on my local config, but seems fine to me.

ecino:
Thanks for reviewing it!

PR #1598, opened 2021-12-17
hotfix: add a browser user-agent to the image fetch to bypass restrictions

TheRaphael0000:


child_compassion/models/child_pictures.py
```diff
@@ -9,13 +9,18 @@
...
+from urllib.request import urlopen, Request
 
 from odoo import models, fields, api, _
 from odoo.http import request
 
 logger = logging.getLogger(__name__)
 
+# This User-Agent simulate a browser, so that the fetch is not blocked
+HEADERS = {
+    "User-Agent": "Mozilla/5.0 (Windows; U; Windows NT 5.1; en-US; rv:1.9.0.7) Gecko/2009021910 Firefox/3.0.7"
```
TheRaphael0000, line 21: We can use other user agent (this one i've found it on Internet)
Here is mine for example :
Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/96.0.4664.93 Safari/537.36

PR #1601, opened 2022-01-06
CS-656 Cannot see campaign results on mass mailing

TheRaphael0000:
- Add handle UTMs when generating letters
- I had to refix the timezone problem that I've fixed a while back (I don't know why only one part was correctly commited)

### DIFF
--- a/mobile_app_connector/models/compassion_correspondence.py
+++ b/mobile_app_connector/models/compassion_correspondence.py
@@ -221,9 +221,16 @@ class CompassionCorrespondence(models.Model):
             )
         )
         # Use the latest preview for generating the letter
-        gen[:1].generate_letters()
+        # update utms
+        utm_source = params.get("utm_source")
+        utm_medium = params.get("utm_medium")
+        utm_campaign = params.get("utm_campaign")
+        utms = self.env["utm.mixin"].get_utms(utm_source, utm_medium, utm_campaign)
+
+        gen[:1].generate_letters(utms)
         # Delete old previews
         gen[1:].unlink()
+
         return {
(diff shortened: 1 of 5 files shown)


mobile_app_connector/models/compassion_correspondence.py
```diff
@@ -221,9 +221,16 @@ def mobile_send_letter(self, *params, **parameters):
             )
         )
         # Use the latest preview for generating the letter
-        gen[:1].generate_letters()
+        # update utms
+        utm_source = params.get("utm_source")
+        utm_medium = params.get("utm_medium")
+        utm_campaign = params.get("utm_campaign")
+        utms = self.env["utm.mixin"].get_utms(utm_source, utm_medium, utm_campaign)
```
ecino, line 228: The method `get_utms` is defined in module `mass_mailing_switzerland` and is therefore not available from here. I think you should change and assume that the utms are already transmitted as a dictionary (call the `get_utms` method before calling the generator).
```suggestion
        utms = params.get("utms")
```

ecino: Please also update the comment at line 184 to mention that the params can contain this `utms` value and how it should be passed in order to work.

ecino (changes requested):

ecino:
Or maybe you should just move the method `def get_utms` into the module `cms_form_compassion` which is both inherited by `mobile_app_connector` and `website_compassion`. Maybe it will be easier this way, otherwise it will be tricky to convert the parameters from the javascript of the `website_compassion`module.

ecino (approved):

PR #1603, opened 2022-01-10
CO-3842 When printing a communication, it doesn't take all user speci…

TheRaphael0000:
…fic printing settings.

- Refactor the code
- Use the behaviour function instead of the attributes of ir.action.report to set the trays

### DIFF
--- a/partner_communication/models/communication_job.py
+++ b/partner_communication/models/communication_job.py
@@ -851,7 +851,7 @@ class CommunicationJob(models.Model):
             return next((v for v in source if v), False)
 
         print_options["output_tray"] = get_first((
-            self[0:].printer_output_tray_id.system_name,
+            self[:1].printer_output_tray_id.system_name,
             config_lang.printer_output_tray_id.system_name,
             behaviour["output_tray"]
         ))
--- a/partner_communication/models/communication_job.py
+++ b/partner_communication/models/communication_job.py
@@ -789,7 +789,7 @@ class CommunicationJob(models.Model):
             if job.attachment_ids:
                 # Print letter
                 print_name = name[:3] + " " + (job.subject or "")
-                output_tray = job.print_letter(print_name)
+                output_tray = job.print_letter(print_name)["output_tray"]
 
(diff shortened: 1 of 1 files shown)


partner_communication/models/communication_job.py
```diff
@@ -825,55 +825,48 @@ def print_letter(self, print_name, **print_options):
...
+
+        # The get the print options in the following order of priority:
+        # - partner.communication.job (only for output bin)
+        # - partner.communication.config
+        # - ir.actions.report (behaviour: which can be specific for the user currently logged in)
+        def get_first(source):
+            return next((v for v in source if v), False)
+
+        print_options["output_tray"] = get_first((
+            self[0:].printer_output_tray_id.system_name,
```
ecino: Did you want to take the first record of the set? I think it would be like this:
```suggestion
            self[:1].printer_output_tray_id.system_name,
```

partner_communication/models/communication_job.py
```diff
@@ -825,55 +825,48 @@ def print_letter(self, print_name, **print_options):
...
-                doc_format=report.report_type,
-                action=behaviour['action'],
-                input_tray=input_tray,
-                output_tray=output_tray,
-                **print_options
-            )
-        return output_tray
+            printer.print_document(report.report_name, to_print[0], **print_options)
+
+        return print_options
```
ecino, line 869: By changing the return type, don't you break anything else in the code? Did you check all calls to this method so that they continue to behave as expected?

ecino (changes requested):

partner_communication/models/communication_job.py
TheRaphael0000: Yes my bad, what I've wrote does nothing

partner_communication/models/communication_job.py
TheRaphael0000: Yes, I've checked with the PyCharm Odoo plugin and the only part that were using the output I've changed it

ecino (approved):

PR #1604, opened 2022-01-11
CO-3859 Sponsorships started automatically

TheRaphael0000:
Only allow to change the sponsor (on GMC side) when the contract is already active (on GMC side)

### DIFF
--- a/sponsorship_compassion/models/contracts.py
+++ b/sponsorship_compassion/models/contracts.py
@@ -544,8 +544,8 @@ class SponsorshipContract(models.Model):
             self.mapped("partner_id").update_number_sponsorships()
             old_partners.update_number_sponsorships()
 
-        if "correspondent_id" in vals and self.state in ["active"]:
-            for record in self:
+        if "correspondent_id" in vals:
+            for record in self.filtered(lambda x: x.state in ["active"]):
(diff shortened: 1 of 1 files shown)


TheRaphael0000:
I'm not sure about this one

sponsorship_compassion/models/contracts.py
```diff
@@ -544,7 +544,7 @@ def write(self, vals):
             self.mapped("partner_id").update_number_sponsorships()
             old_partners.update_number_sponsorships()
 
-        if "correspondent_id" in vals:
+        if "correspondent_id" in vals and self.state in ["active"]:
             for record in self:
```
ecino: `self.state` wouldn't work when the recordset has more than one sponsorship.

```suggestion
        if "correspondent_id" in vals:
            updated_correspondents = self.filtered("global_id")
            for record in updated_correspondents:
```

ecino (changes requested):

sponsorship_compassion/models/contracts.py
TheRaphael0000: Sorry didn't saw your suggestion, we can commit yours instead

ecino (approved):

PR #1607, opened 2022-01-19
CO-3867 Simplify letter import

TheRaphael0000:
Remove old barcodes detection methods and implement new ones. Also remove the template detection and language detection.
The new version only uses:
- fitz
- pillow
- pyzbar
 
These were already imported. Opencv and PymuPDF are not used for this process, but they maybe are used elsewhere so I did not remove the dependency in the requirement file.

- add a new way of analyzing the letters
- remove all the old processing files and simplify/refactor the process
- simplify strategies update debug code
- use generator instead of implementing directly the function


sbc_compassion/models/correspondence_template.py
```diff
@@ -408,24 +407,6 @@ def _compute_template_data(self):
             template.qrcode_y_max = template.page_height * float(
                 config_obj.get_param("qrcode_y_max")
             )
-            if template.pattern_image:
```
ecino, line 411: I think you can also remove the field `pattern_image` from the model and all its usages, because it was only used for template detection if I remember correctly.

sbc_compassion/models/correspondence_template.py
```diff
@@ -408,24 +407,6 @@ def _compute_template_data(self):
...
-                    raise UserError(
-                        _(
-                            "The pattern could not be detected in given "
-                            "template image."
-                        )
-                    )
-                else:
-                    pattern_keypoints = res[0]
-
-                template.nber_keypoints = pattern_keypoints.shape[0]
```
ecino, line 428: You should also be able to remove this field from the model and all its usages (`nber_keypoints`)

sbc_compassion/models/correspondence_template.py
```diff
@@ -408,24 +407,6 @@ def _compute_template_data(self):
             template.qrcode_y_max = template.page_height * float(
                 config_obj.get_param("qrcode_y_max")
             )
-            if template.pattern_image:
-                # pattern detection
-                res = pr.patternRecognition(
-                    template_cv_image,
-                    template.pattern_image,
-                    template.get_pattern_area(),
```
ecino, line 416: The method `get_pattern_area` should also be removed.

sbc_compassion/models/import_letters_history.py
```diff
@@ -214,7 +137,6 @@ def button_save(self):
 
     @api.multi
     def button_review(self):
-        """ Returns a form view for import lines in order to browse them """
```
ecino: I don't think it's an improvement to remove all the comments from the code. It's ok when the function name is self explanatory and the code easy to understand but in general commenting the code is a good practice.

sbc_compassion/models/import_letters_history.py
```diff
@@ -233,57 +155,62 @@ def onchange_config(self):
             for field, val in list(config.get_correspondence_metadata().items()):
                 setattr(self, field, val)
 
-    ##########################################################################
-    #                             PRIVATE METHODS                            #
-    ##########################################################################
+    def local_generator(self):
+        d = set(self.data)
```
ecino: Please add a comment explaining this method and what it's used for. Should it be private (`local_generator`)?

ecino (changes requested):

sbc_compassion/models/import_letters_history.py
ecino: Please use human understandable variable names.

ecino: Specify in the comments the return type

sbc_compassion/models/import_letters_history.py
```diff
@@ -233,57 +155,62 @@ def onchange_config(self):
...
-                # case with normal format (PDF,TIFF)
-                elif func.check_file(attachment.name) == 1:
-                    logger.debug(f"Analyzing file {progress}/{self.nber_letters}")
-                    self._analyze_attachment(file_data, attachment.name)
-                    progress += 1
-                else:
-                    raise UserError(_("Only zip/pdf files are supported."))
-            else:
-                raise UserError(_("Two files are the same"))
+        for i, n, file in generator():
```
ecino: Please use human readable variable names. Single characters are not a good practice and make it difficult to understand the code.

sbc_compassion/models/import_letters_history.py
```diff
@@ -233,57 +155,62 @@ def onchange_config(self):
...
+            partner_code, child_code, preview = read_barcode.letter_barcode_detection_pipeline(pdf_data)
+        except Exception as e:
+            logger.error(f"Couldn't import file {file_name} : \n{traceback.format_exc()}")
+            return
+
+        partner = self.env["res.partner"].search([("ref", "=", partner_code)], limit=1)
+        if partner:
+            partner_id = partner.id
+        else:
+            partner_id = None
```
ecino: This doesn't seem required. If no partner is found, `partner.id` will return False.

sbc_compassion/models/import_letters_history.py
```diff
@@ -233,57 +155,62 @@ def onchange_config(self):
...
+
+        child = self.env["compassion.child"].search(["|", ("code", "=", child_code), ("local_id", "=", child_code)], limit=1)
+        if child:
+            child_id = child.id
+        else:
+            child_id = None
```
ecino: This isn't needed like above for partner.id

sbc_compassion/models/import_letters_history.py
```diff
@@ -233,57 +155,62 @@ def onchange_config(self):
...
+
+        data = {
+            "import_id": self.id,
+            "partner_id": partner_id,
```
ecino:
```suggestion
            "partner_id": partner.id,
```

sbc_compassion/models/import_letters_history.py
```diff
@@ -233,57 +155,62 @@ def onchange_config(self):
...
+            "child_id": child_id,
```
ecino:
```suggestion
            "child_id": child.id,
```

sbc_compassion/tools/read_barcode.py
```diff
@@ -0,0 +1,125 @@
...
+        print(file)
+        file_data = open(file, "rb").read()
+        o = letter_barcode_detection_pipeline(file_data)
+        print(o[0], o[1])
+        output.append(o)
+
+    success = sum([o[0] is not None for o in output])
+    total = len(output)
+    print(f"{success} / {total} = {success/total}")
+    print(f"{time.perf_counter() - t} sec.")
```
ecino: Was it used for testing? I have the feeling this can be removed.

sbc_compassion/views/import_letters_history_view.xml
```diff
@@ -8,29 +8,27 @@
         <field name="arch" type="xml">
             <form string="Imports">
                 <header>
-                    <field name="state" widget="statusbar" statusbar_visible="draft,open,done"/>
+                    <field name="state" widget="statusbar" statusbar_visible="draft,pending,open,done"/>
                     <field name="import_completed" invisible="1"/>
                     <button name="button_save" string="Import Letters" states="ready" type="object" class="oe_highlight"/>
-                    <button name="button_import" string="Load Files" type="object" states="draft"/>
+                    <button name="button_import" string="Load Files" type="object"/>
```
ecino: Why removing the state filter? I think the button makes no sense when the import is not in draft state.

sbc_compassion/views/import_letters_history_view.xml
```diff
@@ -8,29 +8,27 @@
         <field name="arch" type="xml">
             <form string="Imports">
                 <header>
-                    <field name="state" widget="statusbar" statusbar_visible="draft,open,done"/>
+                    <field name="state" widget="statusbar" statusbar_visible="draft,pending,open,done"/>
```
ecino: I prefer to hide the pending state (except when it is in this state), it makes it more clean. So I think we can revert to as it was before.

sbc_compassion/views/import_letters_history_view.xml
```diff
@@ -8,29 +8,27 @@
...
                 </header>
                 <sheet>
                     <h1>Imported Letters</h1>
                     <group>
-                        <label for="data" string="Files" states="draft"/>
-                        <div states="draft">
-                            <field name="data" widget="many2many_binary"/>
-                        </div>
+                        <label for="data" string="Files"/>
+                        <field name="data" widget="many2many_binary"/>
```
ecino: I would leave the state filter, as we remove the data when the import is done no?

sbc_compassion/views/import_letters_history_view.xml
```diff
@@ -8,29 +8,27 @@
...
                         <field name="is_encourager"/>
                         <field name="physical_attachments"/>
                         <field name="attachments_description"/>
-                        <field name="template_id" states="draft,open"/>
+                        <field name="template_id"/>
                     </group>
                     <group>
                         <button name="button_review" string="Review Imports" type="object" states="open" colspan="2"/>
-                        <field name="import_line_ids" states="open,ready" domain="[('id','=','-1')]">
+                        <field name="import_line_ids" domain="[('id','=','-1')]">
```
ecino: I would also leave the state filter, why did you remove all these filters?

ecino (changes requested):

ecino:
You can remove python-opencv and opencv-contrib-python from `requirements.txt` file, as they are no longer used within this repository

sbc_compassion/models/correspondence_template.py
```diff
@@ -408,24 +381,6 @@ def _compute_template_data(self):
...
-                            "The pattern could not be detected in given "
-                            "template image."
-                        )
-                    )
-                else:
-                    pattern_keypoints = res[0]
-
-                template.nber_keypoints = pattern_keypoints.shape[0]
 
     def _get_cv2_image(self):
```
ecino: This method can be removed

sbc_compassion/models/correspondence_template.py
```diff
@@ -119,7 +113,6 @@ class CorrespondenceTemplate(models.Model):
         copy=True,
         readonly=False,
     )
-    nber_keypoints = fields.Integer("Number of key points")
```
ecino, line 122: In fact you can also remove all fields above at lines 94-115 (and all their usages)!
You can remove the model `correspondence.lang.checkbox` and all usages (the security accesses too).

sbc_compassion/models/correspondence_template.py
```diff
@@ -443,7 +398,7 @@ def _verify_template(tpl):
     """
     width = tpl.page_width
     height = tpl.page_height
-    if tpl.template_image and tpl.pattern_image:
```
ecino, line 446: You can remove the whole method. Remove also the caller method `def _verify_template`

ecino (changes requested):

sbc_compassion/data/correspondence_template_data.xml
```diff
@@ -159,24 +127,6 @@
...
-        </record>
-        <record id="qrcode_y_min" model="ir.config_parameter">
-            <field name="key">qrcode_y_min</field>
-            <field name="value">0.0</field>
-        </record>
-        <record id="qrcode_y_max" model="ir.config_parameter">
-            <field name="key">qrcode_y_max</field>
-            <field name="value">0.3</field>
-        </record>
-
```
ecino, line 179: We can also remove all `bluecorner_*` fields and usages (here just below).

sbc_compassion/models/correspondence_template.py
```diff
@@ -172,40 +113,16 @@ def _compute_template_image(self):
     ##########################################################################
     @api.model
     def create(self, vals):
-        template = super().create(vals)
-        template._compute_template_data()
-        return template
+        return super().create(vals)
```
ecino: In this case the inheritance doesn't do anything and can simply be removed.

sbc_compassion/models/correspondence_template.py
```diff
@@ -172,40 +113,16 @@ def _compute_template_image(self):
...
 
     @api.multi
     def write(self, vals):
         super().write(vals)
-        if "template_image" in vals or "pattern_image" in vals:
```
ecino, line 182: In this case the inheritance doesn't do anything and can simply be removed.

ecino (changes requested):

ecino (approved):

PR #1608, opened 2022-01-20
CO-3874 changing correspondant terminates the sponsorship

TheRaphael0000:
I keep having this message from GMC, but my change seems to make sense:

```
Beneficiary is in Ineligible state
```

Is this a problem with my developpment environment (database state: sponsorships/mappings) or is there an error ?


TheRaphael0000:
After testing with a new child hold, this change seem to fix the problem on my side.

ecino (approved):

PR #1609, opened 2022-01-21
CO-3864 remove all suspension proccess for the invoices

TheRaphael0000:
Remove code related these invoice editions

### DIFF
--- a/sponsorship_compassion/models/contracts.py
+++ b/sponsorship_compassion/models/contracts.py
@@ -77,7 +77,6 @@ class SponsorshipContract(models.Model):
         default=lambda self: self._get_standard_lines(), readonly=False
     )
     preferred_name = fields.Char(related="child_id.preferred_name")
-    suspended_amount = fields.Float(compute="_compute_suspended_amount")
 
     child_id = fields.Many2one(
         "compassion.child",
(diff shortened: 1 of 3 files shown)


sponsorship_compassion/models/contracts.py
```diff
@@ -623,53 +623,8 @@ def suspend_contract(self):
                 message_type="comment",
             )
 
-        # Change invoices if config tells to do so.
```
ecino, line 626: In fact you can completely remove this method and the code that calls it. We no longer need anything from that. You can also remove the code that inherits `def suspend_funds` in `sponsorship_compassion/models/project_compassion.py` at line 48.

ecino: You can also remove field `suspended_amount` and its compute method.

sponsorship_compassion/models/contracts.py
```diff
@@ -623,53 +623,8 @@ def suspend_contract(self):
...
-            "name": "Replacement of sponsorship (fund-suspended)",
-        }
-        rec = self.env["account.analytic.default"].account_get(product.id)
-        if rec and rec.analytic_id:
-            vals["account_analytic_id"] = rec.analytic_id.id
-
-        return vals
-
     @api.multi
     def reactivate_contract(self):
```
ecino: You can also remove completely method `def reactivate_contract` and the inheritance of method `def reactivate_project` in project_compassion.py

ecino (changes requested):

ecino (approved):

PR #1613, opened 2022-02-02
CO-3884 Event planned emails are not sent on the scheduled dates# Ple…

TheRaphael0000:
…ase enter the commit message for your changes. Lines starting

instead of commiting the transaction use savepoints, this allow to leave savepoints when odoo use savepoints

### DIFF
--- a/partner_communication/models/communication_job.py
+++ b/partner_communication/models/communication_job.py
@@ -231,57 +231,62 @@ class CommunicationJob(models.Model):
         """
         # Object ids accept lists, integer or string values. It should contain
         # a comma separated list of integers
-        object_ids = vals.get("object_ids")
-        if isinstance(object_ids, list):
-            vals["object_ids"] = ",".join(map(str, object_ids))
-        elif object_ids:
-            vals["object_ids"] = str(object_ids)
-        else:
-            vals["object_ids"] = str(vals["partner_id"])
+        try:
+            with self.env.cr.savepoint():
+                object_ids = vals.get("object_ids")
+                if isinstance(object_ids, list):
+                    vals["object_ids"] = ",".join(map(str, object_ids))
+                elif object_ids:
+                    vals["object_ids"] = str(object_ids)
+                else:
+                    vals["object_ids"] = str(vals["partner_id"])
 
-        same_job_search = [
-            ("partner_id", "=", vals.get("partner_id")),
-            ("config_id", "=", vals.get("config_id")),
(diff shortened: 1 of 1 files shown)


partner_communication/models/communication_job.py
```diff
@@ -231,57 +231,62 @@ def create(self, vals):
...
-            job.send_mode = send_mode[0]
-        if "auto_send" not in vals and "default_auto_send" not in self.env.context:
-            job.auto_send = send_mode[1]
-
-        if not job.body_html or not strip_tags(job.body_html):
-            job.refresh_text()
-        else:
-            job.set_attachments()
+        try:
+            with self.env.cr.savepoint():
```
ecino: I wouldn't put a savepoint in the create method, only in the send method (where we actually send the communication). You can revert all this code in my opinion.

ecino (changes requested):
I think the savepoint in `send_mail` is enough and the rest can be reverted.

partner_communication/models/communication_job.py
```diff
@@ -167,12 +167,15 @@ def count_pdf_page(self):
...
-                    record.pdf_page_count = pdf.getNumPages()
+                    try:
+                        report = record.report_id.with_context(
+                            lang=record.partner_id.lang, must_skip_send_to_printer=True
+                        )
+                        pdf_str = report.render_qweb_pdf(record.ids)
+                        pdf = PdfFileReader(BytesIO(pdf_str[0]))
+                        record.pdf_page_count = pdf.getNumPages()
+                    except (UserError, PdfReadError, QWebException):
+                        pass
```
ecino:
```suggestion
                        self.env.clear()
                        record.pdf_page_count = 0
```

partner_communication/models/communication_job.py
```diff
@@ -766,7 +769,14 @@ def _send_mail(self):
                 .with_context(lang=partner.lang) \
                 .create_emails(self.email_template_id, [self.id], email_vals)
             self.email_id = email
-            email.send(auto_commit=True)  # Commit emails to avoid sending twice
+
+            try:
+                with self.env.cr.savepoint():
+                    email.send()
+            except Exception:
+                logger.warning("Couldn't send this email : ", self)
```
ecino:
```suggestion
                logger.error("Error while sending communication by email to %s ", partner.email, exc_info=True)
```

ecino (changes requested):

ecino (approved):

PR #1621, opened 2022-02-09
CO-3897 thank you letter from events for amounts big amount should be…

TheRaphael0000:
… routed to Debora in german and Christian in french.

I don't know if it's linked but I had this error.
I forgot that I changed this yesterday.


ecino (approved):
The change looks good but doesn't seem related to the issue.

PR #1628, opened 2022-02-21
CO-3910 Add "needs consent" state for W&P sponsorships

TheRaphael0000:


ecino:
I think all of this should be moved into compassion-switzerland, because "write and pray" sponsorships are a swiss specific thing. Maybe it could be in module `sponsorship_switzerland`

PR #1630, opened 2022-02-24
CO-3741 Preview of generate communication wizards are broken

TheRaphael0000:
- refactor code for preview
- fix the force language feature for the preview
- remove the code that was adding a filter on the partner domain search when change the force language (was this really a desired feature?)

### DIFF
--- a/partner_communication/wizards/generate_communication_wizard.py
+++ b/partner_communication/wizards/generate_communication_wizard.py
@@ -56,7 +56,6 @@ class GenerateCommunicationWizard(models.TransientModel):
         default=lambda s: s._default_report(),
         readonly=False,
     )
-    language_added_in_domain = fields.Boolean()
     preview_email = fields.Html(readonly=True)
     preview_pdf = fields.Binary(readonly=True)
     communication_ids = fields.Many2many(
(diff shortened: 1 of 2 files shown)


partner_communication/wizards/generate_communication_wizard.py
```diff
@@ -129,58 +116,48 @@ def _compute_progress(self):
...
         if self.selection_domain:
             self.partner_ids = self.env["res.partner"].search(
                 safe_eval(self.selection_domain)
             )
-        if not self.force_language:
-            self.language_added_in_domain = False
 
-    @api.onchange("model_id")
-    def onchange_model_id(self):
+    @api.onchange("model_id", "force_language")
```
TheRaphael0000, line 126: When force_language is set to "fr_CH" for example. It does not return a subject nor a body_html in french

partner_communication/wizards/generate_communication_wizard.py
```diff
@@ -129,58 +116,48 @@ def _compute_progress(self):
...
+    def onchange_model_or_language(self):
         if self.model_id:
+            config = self.model_id.email_template_id
+            if self.force_language:
+                config.with_context(lang=self.force_language)
```
ecino: in order to work, you need to reload the variable : `config = config.with_context()`

ecino (approved):

PR #1632, opened 2022-02-25
CS-610 Add attachment icon in interaction resume tree view

TheRaphael0000:
update the monster query to handle has attachment

### DIFF
--- a/crm_compassion/models/interaction_resume.py
+++ b/crm_compassion/models/interaction_resume.py
@@ -27,7 +27,7 @@ class InteractionResume(models.TransientModel):
     communication_date = fields.Datetime()
     subject = fields.Text()
     other_type = fields.Char()
-    has_attachment = fields.Boolean(compute="_compute_has_attachment")
+    has_attachment = fields.Boolean()
     body = fields.Html()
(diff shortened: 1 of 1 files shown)


crm_compassion/models/interaction_resume.py
```diff
@@ -181,8 +190,15 @@ def populate_resume(self, partner_id):
...
-                        mm.id as mass_mailing_id
+                        mm.id as mass_mailing_id,
+                        false as has_attachment
                         FROM "mail_mail_statistics" as mail
+                        FULL OUTER JOIN (
+                            SELECT m.id AS id, Count(ma.message_id) AS nb_attachment
+                            FROM mail_message AS m
+                            RIGHT JOIN message_attachment_rel ma ON ma.message_id = m.id
+                            GROUP BY m.id
+                        ) AS attachment ON mail.mail_mail_id = attachment.id
```
ecino: Here you have "false as has_attachment" but you add a Full outer join, is it necessary?

TheRaphael0000: Forgot to add back before committing when testing

ecino (approved):

PR #1635, opened 2022-03-22
Co 3929 : Make sponsorships end reasons available in no sub reasons

ovesco:
The reasons for no sub were hardcoded, original list is kept and options from end_reasons are added to it


ecino:
Can you rebase your branch in order to keep only the relevant commits for this task? The muskathlon commit should not be here. Also the commit message should include the jira task CO-3929 <commit_message>, otherwise it doesn't sync properly with Jira. Thanks!

ovesco:
Oh yes my bad sorry!

PR #1637, opened 2022-03-23
CO-3916: Only cancel invoices if marked as draft

ovesco:
This avoids having open invoices being automatically cancelled

Please review as I am not sure if this is expected behavior. From my understanding the invoices should be automatically closed when their state is draft only (as stated here: https://jira.compassion.ch/browse/CO-3907).


ecino:
Unfortunately, this doesn't resolve the issue. It looks more complicated and I'm currently investigating about it.

ovesco:
That's what I thought... Sorry about it then I was unsure on how to test it correctly

PR #1655, opened 2022-07-18
Seb translation

sebastiencompassion:
Letters with (translation_status =! "done") have their translation_priority incremented each week.

### DIFF
--- a/sbc_translation/models/correspondence.py
+++ b/sbc_translation/models/correspondence.py
@@ -467,7 +467,7 @@ class Correspondence(models.Model):
         for letter in letters_to_translate:
             old_priority = int(letter.translation_priority)
             if old_priority < 4:
-            letter.translation_priority = str(old_priority + 1)
+                letter.translation_priority = str(old_priority + 1)
 
     ##########################################################################
     #                             PRIVATE METHODS                            #
--- a/sbc_translation/models/correspondence.py
+++ b/sbc_translation/models/correspondence.py
@@ -46,7 +46,7 @@ class Correspondence(models.Model):
         ("2", "High"),
         ("3", "Very high"),
         ("4", "Urgent")
-    ], index=True)
+    ], default="0", index=True)
     translation_priority_name = fields.Char(compute="_compute_translation_priority_name", store=True)
     translation_issue = fields.Selection(
         "get_translation_issue_list", help="Issue about the letter reported by the translator")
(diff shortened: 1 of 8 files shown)


sbc_translation/models/correspondence.py
```diff
@@ -456,6 +456,22 @@ def get_translated_elements(self):
                 })
         return res
 
+    @api.multi
+    def increment_priority_cron(self):
+        """
+        Increment priority of letters to translate, maximum
+        priority is 4.
+        """
+        letters_to_translate = self.search([("translation_status", "!=", "done")])
```
ecino: Also filter the letters that have no translation_status:
```suggestion
        letters_to_translate = self.search([("translation_status", "not in", [False, "done"])])
```

sbc_translation/models/correspondence.py
```diff
@@ -456,6 +456,22 @@ def get_translated_elements(self):
                 })
         return res
 
+    @api.multi
```
ecino:
```suggestion
    @api.model
```

sbc_translation/models/correspondence.py
```diff
@@ -456,6 +456,22 @@ def get_translated_elements(self):
...
+        """
+        letters_to_translate = self.search([("translation_status", "!=", "done")])
+
+        for letter in letters_to_translate:
+            old_priority = int(letter.translation_priority)
+            if old_priority == 4:
+                continue
+            new_priority = old_priority + 1
+
+            letter.write({"translation_priority": str(new_priority)})
```
ecino:
```suggestion
            letter.translation_priority = str(old_priority + 1)
```

sbc_translation/models/correspondence.py
```diff
@@ -456,6 +456,22 @@ def get_translated_elements(self):
...
+        """
+        Increment priority of letters to translate, maximum
+        priority is 4.
+        """
+        letters_to_translate = self.search([("translation_status", "!=", "done")])
+
+        for letter in letters_to_translate:
+            old_priority = int(letter.translation_priority)
+            if old_priority == 4:
+                continue
```
ecino:
```suggestion
            if old_priority < 4:
```

sbc_translation/wizards/reply_to_comments.py
```diff
@@ -0,0 +1,14 @@
+from odoo import models, fields, api
+
+
+class ReplyToComments(models.TransientModel):
+    _name = "reply.to.comments"
```
ecino: I would suggest a name that identifies better what it is. (Note: this is to change everywhere)
```suggestion
    _name = "sbc.reply.to.comments.wizard"
```

sbc_translation/wizards/reply_to_comments.py
```diff
@@ -0,0 +1,14 @@
...
+
+    paragraph_ids = fields.Many2many("correspondence.paragraph", string="Paragraphs", readonly=True)
+    answer = fields.Text(string="Answer")
```
ecino: Html will enable better editing of the reply. The string attribute is not needed if it's the same name as the field.
```suggestion
    answer = fields.Html()
```

sbc_translation/wizards/reply_to_comments.py
```diff
@@ -0,0 +1,14 @@
...
+
+    @api.multi
+    def send_reply(self):
+        corr = self.env['correspondence'].search([('id', '=', self._context['active_id'])])
```
ecino: Browse method is designed for this:
```suggestion
        corr = self.env["correspondence"].browse(self.env.context.get("active_id"))
```

ecino (changes requested):
A few improvements suggested but overall it looks good! Thanks for the changes!

ecino (approved):

PR #1661, opened 2022-10-03
FIX associated child event

NoeBerdoz:
# Related Issue
When scanning qrcode from childpack, the event linked with the child dossier is not reported on the sponsorship

# Description
[1] When a child was associated for an event, the sponsorship created by the mobile sponsorship process did not put the event as origin in the sponsorship.

[2] When clicking on the "Other Child" button on the mobile sponsorship process, the other child given was the same as before.

# Fix
[1] Refactor the function _compute_event to manage the QR case by getting the event_id from the child_id.hold_event

[2] refactor the code that release the current child and choose another by not deleting the last child id, so the function that choose another child can fetch a different child_id from the previous one.

### DIFF
--- a/sms_sponsorship/models/sms_child_request.py
+++ b/sms_sponsorship/models/sms_child_request.py
@@ -122,19 +122,8 @@ class SmsChildRequest(models.Model):
     @api.multi
     @api.depends("date")
     def _compute_event(self):
-        limit_date = datetime.today() - relativedelta(days=7)
-        for request in self.filtered(lambda r: r.source == "SMS" and r.date):
-            event_id = self.env["crm.event.compassion"].search(
-                [
-                    ("accepts_sms_booking", "=", True),
-                    ("start_date", "<=", request.date),
-                    ("start_date", ">=", limit_date),
-                ],
-                order="start_date desc",
-                limit=1,
-            )
-            # event_id is None if start_date of most recent event is>1 week old
-            request.event_id = event_id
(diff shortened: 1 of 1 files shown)


sms_sponsorship/models/sms_child_request.py
```diff
@@ -122,19 +122,24 @@ def _compute_full_url(self):
...
+                        [
+                            ("accepts_sms_booking", "=", True),
+                            ("start_date", "<=", sms_request.date),
+                            ("start_date", ">=", limit_date),
+                        ],
+                        order="start_date desc",
+                        limit=1,
+                    )
+                    # event_id is None if start_date of most recent event is > 1 week old
+                    sms_request.event_id = event_id
```
ecino: No need to loop twice on the records. Last line is for v14 compatibility.
```suggestion
            elif request.source == "SMS" and request.date:
                limit_date = datetime.today() - relativedelta(days=7)
                event_id = self.env["crm.event.compassion"].search(
                    [
                        ("accepts_sms_booking", "=", True),
                        ("start_date", "<=", request.date),
                        ("start_date", ">=", limit_date),
                    ],
                    order="start_date desc",
                    limit=1,
                )
                # event_id is None if start_date of most recent event is > 1 week old
                request.event_id = event_id
            else:
                request.event_id = False
```

ecino (changes requested):

ecino (approved):

PR #1663, opened 2022-10-04
Cp 119 - add 'state church' to is_church variable

duvenhage:
Partners that have a 'church' or a 'state church' tag, should not be included in the new donor onboarding process.


duvenhage:
Wrong branch

PR #1671, opened 2022-10-28
14.0 fix gift move

NoeBerdoz:
Related issue : [CP-172](https://compassion-ch.atlassian.net/browse/CP-172)

# Issue: 
When a gift is sent to API, it is missing the journal id and the field currency is not set correctly

# Fix:
- Set gift currency from the company that manage the gift contract (invoice)
- Set gift journal_id from an added setting that depends on each companies instead of an hardcoded journal code
- Correct the gift mapping to escape special characters in the instruction field and to set correctly the currency and the global partner supporter id.

### DIFF
--- a/gift_compassion/models/sponsorship_gift.py
+++ b/gift_compassion/models/sponsorship_gift.py
@@ -204,7 +204,7 @@ class SponsorshipGift(models.Model):
 
     def _compute_currency(self):
         # Set gift currency depending on its invoice currency
-        self.currency_id = self.invoice_line_ids.move_id.currency_id
+        self.currency_id = self.mapped("invoice_line_ids.move_id.currency_id")[:1]
 
     def _compute_name(self):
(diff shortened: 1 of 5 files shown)


gift_compassion/models/sponsorship_gift.py
```diff
@@ -202,6 +202,10 @@ def _compute_invoice_fields(self):
 
             gift.amount = sum(amounts)
 
+    def _compute_currency(self):
+        # Set gift currency depending on its invoice currency
+        self.currency_id = self.invoice_line_ids.move_id.currency_id
```
ecino: Will fail if there are multiple related invoice lines
```suggestion
        self.currency_id = self.mapped("invoice_line_ids.move_id.currency_id")[:1]
```

ecino (changes requested):

PR #1675, opened 2022-11-15
CP-206 household.json mapping correction + CreateLetter behaviour changed when centralizing for norden and mapping prob

simgonzalez:
The mapping for the natural father alive was missing so the code couldn't interpret the value he received, excepting a string and got a boolean (False) value


simgonzalez:
CreateLetter wasn't able to process images because of a centralization that made behaviour change

ecino (approved):

PR #1677, opened 2022-11-18
CP-210 Send error code to wordpress API not always success code 200

simgonzalez:


rest_json_api/http.py
```diff
@@ -33,6 +33,10 @@ def __init__(self, *args):
                 raise
 
     def _json_response(self, result=None, error=None):
+        if error:
+            if error['code'] == 200:
+                error['code'] = re.search(r"^\d{3}", error['data']['message']).group() # match 3 digits (int)
+                error['message'] += f": {error['data']['message']}"
```
ecino: I think this way we make sure the code works whatever is present in the "error" dictionary.
```suggestion
        if isinstance(error, dict) and error.get('code') == 200:
            error_message = error.get("data", {}).get("message", "")
            if error_message:
                error_code = re.search(r"^\d{3}", error_message).group() # match 3 digits (int)
                if error_code:
                    error['code'] = error_code
                error['message'] = error.get("message", "") + " " + error_message
```

ecino (changes requested):

PR #1678, opened 2022-11-18
CP-225 add the feature of commenting automatically the partner when a mandate or bank account is modified

simgonzalez:


ecino (changes requested):
Changes look good in itself but I wouldn't put them in the `partner_communication` module which has nothing to do with this feature. I would rather move all this inside compassion-nordic/compassion_nordic_accounting (which already has the dependency on module `account_banking_mandate`)

PR #1680, opened 2022-11-30
CP-222 add christmas gift to sponsorship

simgonzalez:


ecino:
The commit list doesn't look right. Can you do an interactive rebase to fix that?
```bash
git fetch <name_of_compassion_remote>
git rebase -i <name_of_compassion_remote>/14.0
<keep only the relevant commits, delete all other lines>
<save file and continue>
git push -f

PR #1681, opened 2022-12-01
Cp 222 add christmas gift

simgonzalez:


sponsorship_compassion/models/account_invoice.py
```diff
@@ -77,3 +77,48 @@ def _compute_invoice_category(self):
...
+        self.ensure_one()
+        return [(1, self.invoice_line_ids.id,
+                 {
+                     "price_unit": amt
+                 }
+                 )]
+
+    def _cancel_invoice(self):
+        self.ensure_one()
+        self.button_cancel()
```
ecino: Those methods are already defined in `recurring_contract` I think, and should be removed here I think.

https://github.com/CompassionCH/compassion-accounting/pull/159/files#diff-dc64132b813702c0c45ab48fa9c91743db9bae443d686468b4aa84034e400e69R185

ecino (changes requested):

ecino (approved):

PR #1682, opened 2022-12-06
CP-261 Wrong object given to the communication

simgonzalez:


partner_communication_compassion/models/lifecycle_events.py
```diff
@@ -39,7 +39,7 @@ def process_commkit(self, commkit_data):
                         {
                             "config_id": communication_type.id,
                             "partner_id": sponsor.id,
-                            "object_ids": lifecycle.child_id.id,
+                            "object_ids": lifecycle.child_id.sponsorship_ids.ids,
```
ecino: The problem is that some communications are still connected to the child:
image
So this change should only happen if the `communication_type` is a PlannedExit or UnplannedExit. We should also only take the first sponsorship.
```suggestion
                            "object_ids": lifecycle.child_id.sponsorship_ids[:1].ids,
```

ecino (changes requested):

ecino (approved):

PR #1693, opened 2023-01-05
CP-314 auto-send of b2s communication

simgonzalez:


ecino:
I don't think this solves the issue. I'm also not sure we should split communications when the send_mode is not the same. If the type is the same I think it still make sense to merge them.

PR #1698, opened 2023-01-09
Cp 59 exhaustiv search

simgonzalez:
Sort by priority to make the search more exhaustiv around the whole countries

### DIFF
--- a/child_compassion/wizards/global_child_search.py
+++ b/child_compassion/wizards/global_child_search.py
@@ -38,8 +38,6 @@ class GlobalChildSearch(models.TransientModel):
         string="Field Offices",
         readonly=False,
     )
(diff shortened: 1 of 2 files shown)


child_compassion/wizards/global_child_search.py
```diff
@@ -38,6 +38,8 @@ class GlobalChildSearch(models.TransientModel):
         string="Field Offices",
         readonly=False,
     )
+    is_exhaustive = fields.Boolean(string="Make exhaustive search",
+                                   help="Get a bit of all children depending on the search criteria")
```
ecino: This was another attempt and can be removed, right? It looks like it's never used.

ecino (changes requested):

ecino (approved):

PR #1708, opened 2023-01-31
cp-106 Edit subject of action resume

simgonzalez:
Deborah would like to be able to edit subject of the interaction logged


ecino (changes requested):
This doesn't solve the issue, because interaction resume is a Transient Model which is recomuputed each time. If the subject or the body is edited from here, it should fetch the source object that generated this interaction resume and edit it there to make the change permanent.

Otherwise they will change in interaction resume, but each time you go out and open again Interaction Resume, your changes won't be saved.

PR #1709, opened 2023-01-31
CP-222 Fixes

ecino:
- Add migration for christmas gift contracts
- Various bug fixes

### DIFF
--- a/sponsorship_compassion/models/res_partner.py
+++ b/sponsorship_compassion/models/res_partner.py
@@ -496,3 +496,21 @@ class ResPartner(models.Model):
                 states = [states]
             sponsorships = sponsorships.filtered(lambda s: s.state in states)
         return sponsorships
+
+    def migrate_christmas_contracts(self, christmas_lines, sponsorships):
+        amount = sum(christmas_lines.mapped("subtotal"))
+        if len(sponsorships) > 1:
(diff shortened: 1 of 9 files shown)


sponsorship_compassion/models/res_partner.py
```diff
@@ -496,3 +496,21 @@ def get_portal_sponsorships(self, states=None):
...
+
+    def migrate_christmas_contracts(self, christmas_lines, sponsorships):
+        amount = sum(christmas_lines.mapped("subtotal"))
+        if len(sponsorships) > 1:
+            amount = amount / len(sponsorships)
+        self.env.cr.execute("""
+            UPDATE recurring_contract
+            SET christmas_invoice = %s
+            WHERE id = ANY(%s)
+        """, [amount, sponsorships.ids])
```
simgonzalez, line 508: Why not using an fstring ?

simgonzalez (approved):

sponsorship_compassion/models/res_partner.py
ecino: sql injection risk, I think we should always use the parameters of the execute function to avoid that.

simgonzalez: I may have created some security issue in the code then :/

PR #1723, opened 2023-03-13
Add pycups in requirement.txt

eicher31:


requirements.txt
```diff
@@ -1,6 +1,7 @@
 jwt
 pyquery
 pytz
+pycups
```
ecino, line 4: It's already there: https://github.com/CompassionCH/compassion-switzerland/blob/12.0/requirements.txt#L21
```suggestion
pycups==1.9.*
```

PR #1729, opened 2023-03-28
FIX: send report object instead of report.report_name

NoeBerdoz:
_Related Issue: T0075 - Print reminder print duplex_

Send the report as a report object instead of report.report_name to ensure the handling of the right report afterwards

Related Pull Request: [1548](https://github.com/CompassionCH/compassion-switzerland/pull/1548), [5](https://github.com/CompassionCH/report-print-send/pull/5)

### DIFF
--- a/partner_communication/models/communication_attachment.py
+++ b/partner_communication/models/communication_attachment.py
@@ -109,10 +109,7 @@ class CommunicationAttachment(models.Model):
             else:
                 to_print = attachment.data
 
-            report = self.env["ir.actions.report"]._get_report_from_name(
-                attachment.report_name
-            ).with_context(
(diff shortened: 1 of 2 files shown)


partner_communication/models/communication_attachment.py
```diff
@@ -109,10 +109,7 @@ def print_attachments(self, output_tray=None):
             else:
                 to_print = attachment.data
 
-            report = self.env["ir.actions.report"]._get_report_from_name(
```
NoeBerdoz, line 112: We don't need to get the report from `_get_report_from_name()` as the report object is already present in self.report_id

ecino (approved):

PR #1731, opened 2023-03-30
T0093 Upgrade crm_request migration

ecino:


partner_search_fuzzy/__init__.py
```diff
@@ -1,11 +1,11 @@
 ##############################################################################
 #
-#    Copyright (C) 2019 Compassion CH (http://www.compassion.ch)
+#    Copyright (C) 2014 Compassion CH (http://www.compassion.ch)
```
NoeBerdoz: Maybe we shouldn't rollback in time, we can set that to 2023, just to say

NoeBerdoz (approved):

PR #1735, opened 2023-04-03
T0090: B2S terminated sponsorships.

simgonzalez:
Don't send letter to terminated sponsorships

### DIFF
--- a/partner_communication_compassion/models/correspondence.py
+++ b/partner_communication_compassion/models/correspondence.py
@@ -171,8 +171,8 @@ class Correspondence(models.Model):
         if self.env.context.get("force_send"):
             eligible_letters = self
         else:
-            eligible_letters = self.filtered(lambda l: l.sponsorship_id.state == 'terminated')
-            (self - eligible_letters).communication_id.filtered(lambda c: c.state == "pending").unlink()
+            eligible_letters = self.filtered(lambda l: l.sponsorship_id.state == 'active')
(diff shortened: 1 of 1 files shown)


ecino (changes requested):
Not a good fix. `self.sponsorship_id` may crash if self is a recordset. Returning False is a bad way. I would rather filter out the letters that have the sponsorship not active and continue the code with those excluded. Also unlink any communication if an old one was already existing.

partner_communication_compassion/models/correspondence.py
```diff
@@ -166,7 +166,15 @@ def send_communication(self):
...
         :return: True
         """
-        partners = self.mapped("partner_id")
+        # We shouldn't send communication to terminated contracts
+        # We should also delete pending communication for those terminated contracts
+        if self.env.context.get("force_send"):
+            eligible_letters = self
+        else:
+            eligible_letters = self.filtered(lambda l: l.sponsorship_id.state == 'terminated')
+            (self - eligible_letters).communication_id.filtered(lambda c: c.state == "pending").unlink()
```
ecino:
```suggestion
            eligible_letters = self.filtered(lambda l: l.sponsorship_id.state == 'active')
            (self - eligible_letters).mapped("communication_id").filtered(lambda c: c.state != "done").unlink()
```

ecino (changes requested):

ecino (approved):

PR #1736, opened 2023-04-04
[MIG][14.0] intervention_compassion

ecino:


intervention_compassion/static/mappings/commitment_mapping.json
```diff
@@ -7,7 +7,8 @@
       "search_relational_record": true
     },
     "Commitment_Amount": {
-      "field": "commitment_amount"
+      "field": "commitment_amount",
+      "to_json_conversion": "str(odoo_value)"
```
NoeBerdoz, line 11: Just a general question, it looks a bit weird to me to have a Python function in a JSON value, what is the goal of this ?

NoeBerdoz (approved):

intervention_compassion/static/mappings/commitment_mapping.json
ecino: Those json files simply load records in Odoo and are a small helper to write mappings easily. In fact we could have gone with standard XML records instead, but the mapping is clearly understandable this way. This python function is then used inside a mapping definition in order to allow more complex data conversion. It adds flexibility.

PR #1741, opened 2023-04-13
T0015 14.0 crm_compassion migration

ecino:


crm_compassion/models/event_compassion.py
```diff
@@ -96,10 +96,10 @@ class EventCompassion(models.Model):
         readonly=False,
     )
     invoice_line_ids = fields.One2many(
```
simgonzalez, line 94: Replacing invoice_line_ids by move_line_ids ?

crm_compassion/views/res_partner_view.xml
```diff
@@ -25,28 +25,28 @@
...
+<!--        <field name="name">res.partner.crm.compassion.remove.tracking</field>-->
+<!--        <field name="model">res.partner</field>-->
+<!--        <field name="inherit_id" ref="mail_tracking.view_partner_form"/>-->
+<!--        <field eval="18" name="priority"/>-->
+<!--        <field name="arch" type="xml">-->
+<!--            <button name="%(mail_tracking.action_view_mail_tracking_email)d" position="attributes">-->
+<!--                <attribute name="invisible">1</attribute>-->
+<!--            </button>-->
+<!--        </field>-->
+<!--    </record>-->
```
simgonzalez: Why wouldn't we delete this code ?

crm_compassion/models/event_compassion.py
```diff
@@ -168,10 +165,7 @@ def _compute_income(self):
     @api.depends("total_income", "total_expense")
     def _compute_balance(self):
         for event in self:
-            if event.total_expense and event.total_income:
-                event.balance = event.total_income / float(event.total_expense)
-            else:
-                event.balance = 0.0
+            event.balance = event.total_income - event.total_expense
```
simgonzalez, line 168: It was a division and it became a substraction ?

simgonzalez (changes requested):

crm_compassion/models/event_compassion.py
ecino: Yes, it didn't make any sense in v12! We discussed this with David. :laughing:

crm_compassion/models/event_compassion.py
ecino: It's just the field name, I don't think we need to rename it. BTW the same field name also exists in account.move table, which proves that it's not necessarily wrong to call the field like that even if it points to move.line object (because conceptually we are still talking about invoices, especially in this case).

crm_compassion/views/res_partner_view.xml
ecino: I didn't see this one, I will remove it :+1:

NoeBerdoz (approved):

PR #1742, opened 2023-04-14
14.0 sponsorship tracking

NoeBerdoz:
Related ticket: T0021

Migration of the module from Odoo v12 to v14 following the related guidelines. 

Removed the hardcoded way of the language notification settings, there is now a generic way that let the user choose a language and a user to link to.

After regarding deeper into this notification functionnality, I saw that actually, it's used nowhere, some SDS team confirmed me that they never received anything related to the sub sponsorship tracking, but that it wasn't necessary to implement it, maybe we can add this in the backlog.

The module has been rename from sponsorship_tracking to sponsorship_sub_management.
There is still the old module name sponsorship tracking in the translations .po files, I don't know how to regenerate them


sponsorship_sub_management/models/sds_sub_followers.py
```diff
@@ -0,0 +1,31 @@
...
+from odoo import api, models, fields
+
+
+class SdsSubFollowers(models.Model):
+    """Configuration for SUB sponsorship Notifications."""
+    _name = 'sds.sub.followers'
+    _description = 'SDS SUB Sponsorship Followers'
+
+    # Users to notify after Child Departure
+    sub_followers = fields.Many2one(
```
ecino: The field name should be singular when it's a Many2one relation, and should be suffixed with _id by convention.
```suggestion
    user_id = fields.Many2one(
```

sponsorship_sub_management/models/sds_sub_followers.py
```diff
@@ -0,0 +1,31 @@
...
+        "res.users",
+        string="Sub sponsorships follower",
+        domain=[("share", "=", False)],
+    )
+
+    # Languages that could have notifications
+    active_languages = fields.Many2one(
```
ecino:
```suggestion
    res_lang_id = fields.Many2one(
```

ecino (changes requested):

ecino:
Can you rebase and squash your commits, for keeping only the signifcant changes in a few commits? If you look at the migration procedure, I advised only a few commits, like 3. We don't need the full history of your developement work.

ecino (approved):

ecino:
I did the rebase and merge myself, the code is now available in branch 14.0. I close this one.

NoeBerdoz:
> I did the rebase and merge myself, the code is now available in branch 14.0. I close this one.

Okay thank you :+1:

PR #1756, opened 2023-05-10
T0168 interaction resume add a full time one

simgonzalez:
This commits add the button to open a full time interaction_resume. Improvement possible: Add a setting preference that would allow the users to define what button they want to see on the interface (both (full/newest) or only one of them they prefer).

DEPENDS ON https://github.com/CompassionCH/compassion-switzerland/pull/1557

### DIFF
--- a/crm_compassion/models/res_partner.py
+++ b/crm_compassion/models/res_partner.py
@@ -72,13 +72,13 @@ class Partner(models.Model):
                 ]
             )
 
-    def open_interaction(self):
+    def open_interaction(self, full=False):
         """
         Populates data for interaction resume and open the view
         :return: action opening the view
         """
         self.ensure_one()
-        self.env["interaction.resume"].populate_resume(self.id)
+        self.env["interaction.resume"].populate_resume(self.id, full)
         partners_with_same_email_ids = (
             self.env["res.partner"]
                 .search([("email", "!=", False), ("email", "=", self.email), ('id', 'not in', self.ids)])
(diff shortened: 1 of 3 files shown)


crm_compassion/models/res_partner.py
```diff
@@ -96,6 +96,31 @@ def open_interaction(self):
...
+            "name": _("Interaction resume"),
+            "type": "ir.actions.act_window",
+            "res_model": "interaction.resume",
+            "view_type": "form",
+            "view_mode": "tree,form",
+            "domain": [("partner_id", "in",
+                        self.ids + partners_with_same_email_ids +
+                        self.other_contact_ids.ids)],
+            "target": "current",
+        }
```
ecino: This code duplicates the open_interaction method. Why not just calling the other?
```suggestion
        return self.populate_resume(self.id, full=True)
```

ecino (changes requested):
I think this needs to be migrated to v14 as well.

crm_compassion/models/res_partner.py
```diff
@@ -72,13 +72,13 @@ def _compute_opportunity_count(self):
                 ]
             )
 
-    def open_interaction(self):
+    def open_interaction(self, full=False):
```
ecino, line 76: In v12 (but not in v14), I think you should add the multi decorator, otherwise the context from the view will be passed as the first optional parameter (and will therefor always call the method with the full interaction).
```suggestion
    @api.multi
    def open_interaction(self, full=False):
```

ecino (changes requested):

ecino (approved):

ecino (approved):

PR #1757, opened 2023-05-10
T0168 interaction resume add a full time one

simgonzalez:
This commits add the button to open a full time interaction_resume. Improvement possible: Add a setting preference that would allow the users to define what button they want to see on the interface (both (full/newest) or only one of them they prefer).


ecino (changes requested):
See my comment in the v12 version.

ecino (approved):

PR #1758, opened 2023-05-10
Add a base automation that automatically switch the communication for…

simgonzalez:
… partner when they go from W&P to normal Sponsorship

### DIFF
--- a/partner_communication_compassion/__manifest__.py
+++ b/partner_communication_compassion/__manifest__.py
@@ -48,6 +48,7 @@
         "data/other_emails.xml",
         "data/depart_communications.xml",
         "data/communication_config.xml",
+        "data/switch_wp_comm_pref.xml",
         "data/utm_data.xml",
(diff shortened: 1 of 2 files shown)


NoeBerdoz (approved):

ecino (changes requested):
The changes should be moved into the module `partner_communication_compassion`

ecino (commented):
After checking, I think this was already done here: https://github.com/CompassionCH/compassion-switzerland/blob/5dcfbe0d9557014c5265c74df49be149781aa9b3/partner_communication_switzerland/models/res_partner.py#L465

I don't like adding an automation based on the `on_write` trigger, because it will be checked each time a sponsorship is written which can greatly impact the performance. In my opinion we don't have anything to do regarding this task, except migrating the code for v14.

ecino:
Not an appropriate change.

PR #1759, opened 2023-05-11
T0208 reconciliation invoice matching rules: use bk date

simgonzalez:
We want to use the bank statement date as the initial date for the limit.

### DIFF
--- a/account_reconcile_compassion/models/account_reconcile_model.py
+++ b/account_reconcile_compassion/models/account_reconcile_model.py
@@ -73,7 +73,7 @@ class AccountReconcileModel(models.Model):
         :return:                (query, params)
         '''
         query, params = super()._get_invoice_matching_query(st_lines_with_partner, excluded_ids)
-        bank_statement_date = self.env.context.get("bank_statement_date", False)
+        bank_statement_date = self.env.context.get("bank_statement_date")
         if params["aml_date_limit"] and bank_statement_date:
             date_limit = bank_statement_date - relativedelta(months=self.past_months_limit)
             params['aml_date_limit'] = date_limit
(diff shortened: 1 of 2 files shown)


account_reconcile_compassion/models/account_reconcile_model.py
```diff
@@ -59,3 +61,19 @@ def product_changed(self, product_id, statement_id):
...
+
+        :param st_lines_with_partner: A list of tuples (statement_line, partner),
+                                      associating each statement line to treate with
+                                      the corresponding partner, given by the partner map
+        :param excluded_ids:    Account.move.lines to exclude.
+        :return:                (query, params)
+        '''
+        query, params = super()._get_invoice_matching_query(st_lines_with_partner, excluded_ids)
+        bank_statement_date = self.env.context.get("bank_statement_date", False)
+        if params["aml_date_limit"] and bank_statement_date:
```
ecino: Are there cases where the value might not exist in the dictionary? In that case you should replace the test to avoid raising errors in those cases:
```suggestion
        if params.get("aml_date_limit") and bank_statement_date:
```

account_reconcile_compassion/models/account_reconcile_model.py
```diff
@@ -59,3 +61,19 @@ def product_changed(self, product_id, statement_id):
...
+        model to the provided statement lines.
+
+        :param st_lines_with_partner: A list of tuples (statement_line, partner),
+                                      associating each statement line to treate with
+                                      the corresponding partner, given by the partner map
+        :param excluded_ids:    Account.move.lines to exclude.
+        :return:                (query, params)
+        '''
+        query, params = super()._get_invoice_matching_query(st_lines_with_partner, excluded_ids)
+        bank_statement_date = self.env.context.get("bank_statement_date", False)
```
ecino: This is the default value of the `get` method, no need to specify it.

```suggestion
        bank_statement_date = self.env.context.get("bank_statement_date")
```

ecino (changes requested):

ecino (approved):

PR #1770, opened 2023-05-25
T0095 add survival sponsorship mecanism

simgonzalez:
* Add a sponsorship type
* Add some specific fields on the product that are related to this new sponsorship type
* Add a base automation that warn admin users in case we reach the maximum number of slots available in a project.


survival_sponsorship_compassion/views/sponsorship_contract_view.xml
```diff
@@ -0,0 +1,17 @@
+<odoo>
+    <record id="survival_sponsorship_product_filter_tree_view" model="ir.ui.view">
+        <field name="name">survival.sponsorship.product.filter.tree</field>
+        <field name="model">recurring.contract.line</field>
+        <field name="inherit_id" ref="sponsorship_compassion.view_compassion_sponsorship_line_tree"/>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='product_id']" position="attributes">
+                <attribute name="domain">
+                    [('survival_sponsorship_sale', '=', active_ids[0].type == 'CSP')]
```
simgonzalez: @ecino  do you have any idea why my domain doesn't work ? I tried many things but nothing is working right.
The goal is to only show the Survival product in case the type of the sponsorship is survival.

survival_sponsorship_compassion/models/recurring_contract.py
```diff
@@ -0,0 +1,23 @@
...
+
+class SponsorshipContract(models.Model):
+    _inherit = ["recurring.contract"]
+    _name = "recurring.contract"
+
+    type = fields.Selection(selection_add=[('CSP', 'Survival Sponsorship')])
+
+    @api.multi
+    def invoice_paid(self, invoice):
+        self.filtered(lambda c: c.type == 'CSP').contract_active()
```
ecino, line 22: This will completely bypass the original method for all contracts. You should call the super method on the other types of contracts.

survival_sponsorship_compassion/models/recurring_contract.py
```diff
@@ -0,0 +1,23 @@
...
+#
+##############################################################################
+
+from odoo import fields, models, api
+from odoo.exceptions import UserError
+
+
+class SponsorshipContract(models.Model):
+    _inherit = ["recurring.contract"]
+    _name = "recurring.contract"
```
ecino: This is not necessary when you do simple inheritance (name of the class is the same). You can change this everywhere.
```suggestion
    _inherit = "recurring.contract"
```

survival_sponsorship_compassion/models/recurring_contract_group.py
```diff
@@ -0,0 +1,30 @@
+##############################################################################
+#
+#    Copyright (C) 2014-2015 Compassion CH (http://www.compassion.ch)
```
ecino: Either do it correctly, or remove the file headers (everywhere in your new code).

survival_sponsorship_compassion/models/recurring_contract_group.py
```diff
@@ -0,0 +1,30 @@
...
+#    Releasing children from poverty in Jesus' name
+#    @author: Cyril Sester, Emanuel Cino
```
ecino: Either do it correctly, or remove the file headers (everywhere in your new code).

survival_sponsorship_compassion/models/recurring_contract_group.py
```diff
@@ -0,0 +1,30 @@
...
+#
+#    The licence is in the file __manifest__.py
+#
+##############################################################################
+
+import logging
+
+from odoo import api, fields, models
+
+logger = logging.getLogger(__name__)
```
ecino: This is not used and can be removed (check all imports, you may have others).

survival_sponsorship_compassion/models/recurring_contract_line.py
```diff
@@ -0,0 +1,36 @@
...
+#
+#    The licence is in the file __manifest__.py
+#
+##############################################################################
+
+from odoo import fields, models, api
+from odoo.exceptions import UserError
+
+
+class SponsorshipContract(models.Model):
```
ecino:
```suggestion
class ContractLine(models.Model):
```

survival_sponsorship_compassion/models/recurring_contract_line.py
```diff
@@ -0,0 +1,36 @@
...
+    _inherit = ["recurring.contract.line"]
+    _name = "recurring.contract.line"
+
+    contract_type = fields.Selection(
+        selection_add=[('CSP', 'Survival Sponsorship')],
+        related="contract_id.type",
+        readonly=True,
+    )
```
ecino: I don't think you need to  inherit the related field for adding the selection value. Since the original field is inherited, this one should automatically follow in my opinion.

ecino (changes requested):

ecino (approved):

PR #1772, opened 2023-05-30
T0147 migrate label module

simgonzalez:


label/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module adds some report template and report actions associated with them.
```
ecino: Please be more specific and detail the unique report the module is adding. Include the specifications of the label sheet it produces.

label/reports/one_label.xml
```diff
@@ -0,0 +1,37 @@
+<odoo>
+    <template id="one_label">
+        <!--        <link rel="stylesheet" href="/label/static/src/css/label.css"/>-->
```
ecino: Remove the line

label/reports/one_label.xml
```diff
@@ -0,0 +1,37 @@
...
+        <!-- In order to keep the layout whatever the datas are, everything is inside a div that hide the overflow and occupy all the place -->
+        <div class="o_label_table_cell">
+            <table class="o_label_cell_table">
+                <tr>
+                    <td style="vertical-align: middle; height: 100%;">
+                        <div class="o_label_div_cell">
+                            <table class="o_label_text_table">
+                                <t t-foreach="11" t-as="val" style="margin-bottom:20px;">
```
ecino, line 10: Do we really want margins and paddings defined in pixels ? I fear that it can change the result depending on other factors that could change from one server to another for instance.

label/README.rst
```diff
@@ -0,0 +1,75 @@
+====================
```
ecino, line 1: Please update the readme file using the oca script (that will merge the files present in the readme folder)

label/__manifest__.py
```diff
@@ -0,0 +1,46 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+# pylint: disable=C8101
+{
+    "name": "Mass Label Reporting",
+    "version": "14.0.1.0.0",
+    "author": "Serpent Consulting Services Pvt. Ltd.",
```
ecino: At this point I think we can replace with Us

label/__manifest__.py
```diff
@@ -0,0 +1,46 @@
...
+    "category": "Tools",
+    "website": "http://www.serpentcs.com",
```
ecino: Change this also

ecino (changes requested):

label/reports/one_label.xml
simgonzalez: Normally pixels margin/padding are always working the same if the base layout doesn't change. Which means that if we don't modify the paper format the outcome should be the same in my opinion.

ecino:
OK Like that, but I think we will merge this module with `report_compassion` when migrating that module. I will enter the task accordingly. Having a module just for one report is overkill.

PR #1774, opened 2023-06-01
571 change correspondent payer synv

simgonzalez:


sponsorship_compassion/data/gmc_action.xml
```diff
@@ -36,6 +36,20 @@
             <field name="batch_send">1</field>
             <field name="priority">50</field>
         </record>
+        <!-- Sponsorship Message Creation -->
+        <record id="upsert_correspondent_commitment" model="gmc.action">
+            <field name="direction">out</field>
+            <field name="name">UpsertCorrespondentCommitment</field>
+            <field name="mapping_id" ref="mapping_upsert_correspondence_commitment"/>
+            <field name="model">sponsorship.compassion</field>
```
ecino: This model doesn't exist. Didn't you test the message?
```suggestion
            <field name="model">recurring.contract</field>
```

sponsorship_compassion/models/contracts.py
```diff
@@ -66,8 +66,8 @@ class SponsorshipContract(models.Model):
     transfer_partner_id = fields.Many2one(
         "compassion.global.partner", "Transferred to", readonly=False
     )
-    global_id = fields.Char(
-        help="Connect global ID", readonly=True, copy=False, tracking=True
+    gmc_commitment_id = fields.Char(
+        help="Connect global ID", readonly=True, copy=False, tracking=True, oldname="global_id"
```
ecino: `oldname` is no longer supported in v13. You should replace by a script (using openupgrade library) in pre-migration script.
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0

ecino (changes requested):

sponsorship_compassion/data/gmc_action.xml
simgonzalez: i've tested but i guess i changed on the interface and not in the file (apparently)

sponsorship_compassion/models/contracts.py
simgonzalez: In my local tests that did work but i'll look in the new method to do that.

ecino (approved):

PR #1776, opened 2023-06-12
T0021 [IMP] A few corrections for the sub_management module

ecino:


sponsorship_sub_management/__manifest__.py
```diff
@@ -36,9 +36,7 @@
     "license": "AGPL-3",
     "website": "https://www.compassion.ch",
     "depends": [
-        "sponsorship_compassion",
-        "base_automation",
-        "crm_compassion"
```
NoeBerdoz, line 41: Removing crm_compassion from dependencies will cause issue when installing and using the module on an instance that doesn't have it installed as it is used in the contract model on this function [def onchange_child_id(self)](https://github.com/CompassionCH/compassion-modules/blob/165713ac38fd81ba999fd9ab138bf2a7d42f074b/sponsorship_sub_management/models/contracts.py#LL116C18-L116C18) on super().onchange_child_id()

sponsorship_sub_management/__manifest__.py
```diff
@@ -36,9 +36,7 @@
     "license": "AGPL-3",
     "website": "https://www.compassion.ch",
     "depends": [
-        "sponsorship_compassion",
-        "base_automation",
```
NoeBerdoz, line 40: The module is used with 3 automated actions
- Sponsorship SUB Proposal
- Sponsorship SUB Proposal validation
- Check SUB Accept/Reject after 50 days

I think we need these for the module to work correctly, and the automated actions are depending on base_automation right?

NoeBerdoz (commented):
2 comments about the dependencies, otherwise all good thank you

sponsorship_sub_management/__manifest__.py
ecino: Module sponsorship_compassion is already depending on base_automation through recurring_contract, so we don't need to specify again the dependency.

sponsorship_sub_management/__manifest__.py
ecino: Good catch. However the dependency was not necessary so I adapted the onchange method so that it's not the same as the one in crm_compassion module. If we can remove dependencies it's always good.

NoeBerdoz (approved):

PR #1791, opened 2023-07-13
T0233 - Cropping widget

simgonzalez:
Add a widget for the user to be able to crop his picture. It will compress it at the same time


ecino:
Please move this module on our [OCA/web fork](https://github.com/CompassionCH/web).

PR #1803, opened 2023-07-21
T0412 Supervisor change

simgonzalez:
Supervisor default should be SDS team


ecino (changes requested):
The original version provided a method that can be inherited on purpose to change the supervisor selection. Please restore the original code and inherit in module `sbc_switzerland` in order to assign sds@compassion.ch

This way, the module can be also installed for other Compassion Offices if needed.

PR #1807, opened 2023-07-26
T0394 can't reply to letter comment due to tree view in the wizard

simgonzalez:
When the tree view is available in the wizard the button doesn't work if we take out the tree view direclty of the view it works all good. The translator supervisor has the comment and the letter paragraph behind the wizard in gray.

### DIFF
--- a/sbc_translation/data/mail_template.xml
+++ b/sbc_translation/data/mail_template.xml
@@ -1,4 +1,4 @@
-<odoo>
+<odoo noupdate="1">
     <record id="translation_issue_notification" model="mail.template">
         <field name="name">Translation issue notification</field>
         <field name="subject">A translation issue was raised</field>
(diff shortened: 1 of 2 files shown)


sbc_translation/data/mail_template.xml
```diff
@@ -36,5 +36,11 @@
         </table>
         <hr/>
         <p><a t-att-href="object.translation_url">Open the letter in the translation platform</a></p>
+        <hr/>
+        <br/><p>Thank you for your help,
+        % if user and user.signature:
+        ${user.signature | safe}
+        % endif
```
simgonzalez, line 43:
```suggestion
        
```
@ecino I tried to add the user signature but when testing it doesn't work. Can you correct it for me ? Thanks in advance

sbc_translation/wizards/reply_to_comments_view.xml
```diff
@@ -5,17 +5,9 @@
...
             <form string="Reply to comments">
-                <group>
-                    <field name="paragraph_ids">
-                        <tree default_order="page_id asc, sequence asc">
-                            <field name="original_text"/>
-                            <field name="translated_text"/>
-                            <field name="comments"/>
-                        </tree>
-                    </field>
-                </group>
```
simgonzalez, line 16: I saw with Roma, it's not the most convenient to not have it shown but at least the reply works for the moment.

ecino:
The proper fix is here: https://github.com/CompassionCH/compassion-modules/commit/d826897d738370bc9613921c536b2ae52a7914e1

PR #1808, opened 2023-07-27
T0404 mig reports

simgonzalez:


sponsorship_reporting/__manifest__.py
```diff
@@ -0,0 +1,53 @@
...
+    "summary": "Add the reports for the sponsorships",
+    "version": "14.0.1.0.0",
+    "author": "Compassion CH",
+    "category": "Finance",
+    "website": "http://www.compassion.ch",
+    "depends": [
+        "account_fiscal_year",  # OCA/account-financial-reporting
+        "account_fiscal_month",  # OCA/account-financial-reporting
+        "sponsorship_sub_management",  # CompassionCH/compassion-modules
+        "crm_phone",
```
ecino: Is this really needed? Can we remove it?

simgonzalez: Yes they are needed, I already reduced the list of dependencies

ecino: crm_phone? can you let me know why?

simgonzalez: because of schedule_call method if my memory is good:D

ecino: I don't see any schedule_call in this code. Please point the reference or remove this dependency and check if the module is working without it.

simgonzalez: I tried without it and it seems to work as good.

PR #1809, opened 2023-07-31
T0349 adapt the field to make it look better

simgonzalez:


ecino (changes requested):
The only problem is that we won't be able to search using this field if you don't provide a search method or you store it in database.

ecino (approved):

PR #1810, opened 2023-07-31
T0395 competence filter not working

simgonzalez:


sbc_translation/models/translation_user.py
```diff
@@ -31,8 +31,9 @@ class TranslationUser(models.Model):
         "res.lang.compassion", domain=[("translatable", "=", True)],
         help="Utility field only used for the search view"
     )
-    search_competence_id = fields.Many2one(
-        "translation.competence", help="Utility field only used for the search view"
+    search_competence_ids = fields.Many2one(
```
ecino: I don't see the point of renaming the field, especially if you put a suffix that don't reflect the correct field type.

sbc_translation/models/translation_competence.py
```diff
@@ -16,7 +16,7 @@ class TranslationCompetence(models.Model):
         "translation.competence", "Fallback competence",
         help="Letters will move to this pool if they sit for too long waiting to be translated."
     )
-    name = fields.Char(compute="_compute_name")
+    name = fields.Char(compute="_compute_name", store=True)  # We need to store it to filter on it
```
ecino, line 19: You should add @api.depends decorator on the compute method to retrigger the compute if source or dest language is changed. That should not happen often, but we prefer making sure the name is always correct.

ecino (changes requested):

ecino (approved):

ecino:
Please rebase onto devel branch. There are some conflicts in the code.

sbc_translation/models/translation_user.py
```diff
@@ -31,7 +31,7 @@ class TranslationUser(models.Model):
         "res.lang.compassion", domain=[("translatable", "=", True)],
         help="Utility field only used for the search view"
     )
-    search_competence_ids = fields.Many2many(
+    search_competence_id = fields.Many2many(
         "translation.competence",
```
ecino, line 35: I think you didn't resolve the conflicts correctly. The bad changes are back.

simgonzalez: you're right (:

ecino: Still not resolved.

simgonzalez: you're right (:

sbc_translation/models/translation_user.py
```diff
@@ -31,11 +31,10 @@ class TranslationUser(models.Model):
         "res.lang.compassion", domain=[("translatable", "=", True)],
         help="Utility field only used for the search view"
     )
-    search_competence_ids = fields.Many2many(
+    search_competence_id = fields.Many2one(
         "translation.competence",
-        help="Utility field only used for the search view",
-        compute="_compute_competences",
```
ecino, line 37: I don't see the point here. Why do you remove the compute and change to Many2one? The translator can have multiple competences, not only one. If you remove the compute, this field will always be empty.

ecino (changes requested):

simgonzalez:
i've tested the solution i'm proposing and it actually works. I don't know exactly why but it does :D

ecino (approved):

PR #1811, opened 2023-08-02
T0376 search functionnality not working as expected

simgonzalez:
It's not possible to search on a computed field

### DIFF
--- a/sbc_translation/models/correspondence.py
+++ b/sbc_translation/models/correspondence.py
@@ -23,6 +23,7 @@ class Correspondence(models.Model):
         """
     _inherit = "correspondence"
 
+    name = fields.Char(store=True)
     new_translator_id = fields.Many2one("translation.user", "Local translator")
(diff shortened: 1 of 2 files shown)


sbc_translation/models/correspondence.py
```diff
@@ -23,6 +23,7 @@ class Correspondence(models.Model):
         """
     _inherit = "correspondence"
 
+    name = fields.Char(store=True)
```
ecino: this change can be done directly in the super module (sbc_compassion), and you should make sure that the compute method is correctly decorated with api.depends

ecino (changes requested):

sbc_translation/models/correspondence.py
simgonzalez: Making the modification in the parent module makes the field uselessly stored when sbc_translation isn't installed

ecino (approved):

PR #1829, opened 2023-11-01
fix but need to check edge-cases

jacrodile:


sponsorship_compassion/models/contracts.py
```diff
@@ -998,6 +998,22 @@ def _generate_gifts(self, invoicer, gift_type):
...
                 contracts -= contract
 
+            # checks if there is already a gift for this year which has been cancelled
+            current_year = fields.Datetime.now().year
+            start_of_year = fields.Datetime.from_string(f"{current_year}-01-01 00:00:00")
+            end_of_year = fields.Datetime.from_string(f"{current_year}-12-31 23:59:59")
+            gift_this_year = self.env["account.move.line"].search(
+                [
+                    ("partner_id", "=", contract.partner_id.id),
+                    ("name", "ilike", f"{gift_type}"),
```
ecino: Filtering on the name is a bit risky. I would rather filter on the product_id field. I also miss a filter for the contract to make sure we are considering the same contract gift and not any other one.

ecino (changes requested):

ecino:
This is already included in #1832

PR #1830, opened 2023-11-01
T0641 FEAT: Change invoice matching

nfluckiger:
Modify the invoice matching query to be able to take only the current month.  In the same time we modify the date used in the query to use the line date and not the invoice date

### DIFF
--- a/account_reconcile_compassion/models/account_reconcile_model.py
+++ b/account_reconcile_compassion/models/account_reconcile_model.py
@@ -16,6 +16,15 @@ class AccountReconcileModel(models.Model):
         default=True,
         help="Check to disable thank you letter for donation"
     )
+    only_this_month = fields.Boolean(
+        default=False,
+        help="Check to search only from the start of the month"
+    )
+
+    @api.onchange("past_months_limit")
+    def _uncheck_only_this_month(self):
+        if self.past_months_limit and self.only_this_month:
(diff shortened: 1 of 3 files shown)


account_reconcile_compassion/models/account_reconcile_model.py
```diff
@@ -69,8 +78,33 @@ def _get_invoice_matching_query(self, st_lines_with_partner, excluded_ids):
...
         '''
         query, params = super()._get_invoice_matching_query(st_lines_with_partner, excluded_ids)
-        bank_statement_date = self.env.context.get("bank_statement_date")
-        if params.get("aml_date_limit") and bank_statement_date:
-            date_limit = bank_statement_date - relativedelta(months=self.past_months_limit)
-            params['aml_date_limit'] = date_limit
-        return query, params
\ No newline at end of file
+        for line in st_lines_with_partner:
+            line = line[0]
```
ecino, line 82: It's a bad practice to redefine the variable that is used in a loop. I would rather use another variable for that:
```suggestion
        for line in st_lines_with_partner:
            bank_st_line = line[0]
```

ecino:
@nfluckiger can you check my review, rebase and fix the conflicts please?

ecino (changes requested):

PR #1831, opened 2023-11-03
 [T0648] FIX rename type beneficiary to participant

NoeBerdoz:
Related ticket: [T0648](https://erp.compassion.ch/web#id=1224&action=521&active_id=895&model=project.task&view_type=form&menu_id=924)

Beneficiary has been renamed to Participant

### DIFF
--- a/child_compassion/models/child_lifecycle_event.py
+++ b/child_compassion/models/child_lifecycle_event.py
@@ -26,6 +26,7 @@ class ChildLifecycleEvent(models.Model):
     type = fields.Selection(
         [
             ("Beneficiary Update", "Beneficiary Update"),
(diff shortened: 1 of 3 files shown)


ecino (changes requested):
I would prefer that you add the new value instead of replacing the existing one to make sure we have backward compatibility and don't lose the value of the old records.

NoeBerdoz:
> I would prefer that you add the new value instead of replacing the existing one to make sure we have backward compatibility and don't lose the value of the old records.

Done :+1:

NoeBerdoz:
@ecino 
I added a few changes, there was a related issue in household

ecino (approved):

PR #1872, opened 2024-03-25
displaying success or failure message when the letters is generated

minedcp:
Displaying a success notification when letters has successfully generated and a danger notification when letters hasn't been generated.

### DIFF
--- a/sbc_compassion/models/correspondence_s2b_generator.py
+++ b/sbc_compassion/models/correspondence_s2b_generator.py
@@ -228,13 +228,14 @@ class CorrespondenceS2bGenerator(models.Model):
             # If the operation succeeds, notify the user
             message = "Letters have been successfully generated."
             self.env.user.notify_success(message=message)
+            return self.write({"state": "done", "date": fields.Datetime.now()})
 
         except Exception as error:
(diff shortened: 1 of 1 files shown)


sbc_compassion/models/correspondence_s2b_generator.py
```diff
@@ -189,42 +189,54 @@ def generate_letters_job(self, utms=None):
...
+                        for atchmt in self.image_ids
+                    ]
+                letters += letters.create(vals)
+
+            self.letter_ids = letters
+
+            # If the operation succeeds, notify the user
+            message = "Letters have been successfully generated."
+            self.env.user.notify_success(message=message)
+
```
ecino, line 232: The `write` at line 237 should be moved at the end of the try block, otherwise we will write the state of the generation to done even when an error occured.

ecino (changes requested):

sbc_compassion/models/correspondence_s2b_generator.py
minedcp: The `write` at line 237 has been moved at the end of the try block to prevent the write state of the generation to done when an error occurs.

PR #1875, opened 2024-04-03
T1043: ensure proper execution of contract_waiting method

lucadela96:


NoeBerdoz (approved):
@ecino  by setting the parameter in super, we ensures proper execution of parent class logic only on filtered contracts, this avoid issue with contracts that has no waiting payment states.

PR #1879, opened 2024-04-10
T1053: Restored the old logic, which was removed by odoo maintainers.

lucadela96:


ecino:
Merged manually and partially.

PR #1881, opened 2024-04-17
Fix t1061 sds 21 migration event type group visit

minedcp:
added new type group_visit in def get_event_types for migration event type group visit fix in crm_compassion/models/event_compassion.py


ecino:
We will use sponsor_tour instead which seems to be used for the same purpose. I already fixed the migration to convert existing events.

PR #1882, opened 2024-04-18
Update contracts.py

davidwul:
without the return, it will check again the state in the super and throw an error.

this change is already in production and seems to be ok


ecino (changes requested):
The proper way would be to exclude the "contract" from the super call and still use the super call at the end. It's risky to return inside a loop, we might miss other cases.

PR #1886, opened 2024-04-25
[T0717] - Move In: Sms communication mode into compassion module

langjer:


partner_communication/models/communication_job.py
```diff
@@ -159,6 +166,14 @@ class CommunicationJob(models.Model):
...
     #                                          "Printer Output Bin")
 
+    sms_cost = fields.Float()
+    # sms_provider_id = fields.Many2one(
+    #     "sms.provider",
+    #     "SMS Provider",
+    #     default=lambda self: self.env.ref("sms_939.large_account_id", False),
+    #     readonly=False,
+    # )
+
```
ecino: Please remove commented code. I think providers may not be the way to work in v14, please check with David how it's supposed to work, because he migrated the module `sms_939` and knows how the SMS is supposed to be sent. Make sure that the code in this file is not dependent on a Swiss specific provider but that it can be set by the user or by a setting.

partner_communication/models/communication_job.py
```diff
@@ -491,6 +512,78 @@ def reset(self):
...
+        :return: list of sms_texts
+        """
+        link_pattern = re.compile(r'<a href="([^<>]*)">([^<]*)</a>')
+        sms_medium_id = self.env.ref("sms_sponsorship.utm_medium_sms").id
+        sms_texts = []
+        for job in self.filtered(lambda j: j.state == "pending" and j.partner_mobile):
+            sms_text = job.convert_html_for_sms(link_pattern, sms_medium_id)
+            sms_texts.append(sms_text)
+            job.partner_id.with_context(
+                sms_provider=job.sms_provider_id
```
ecino: This context won't work here. Either we can chose a provider in another way in v14, or we simply delegate the provider choice outside (meaning it cannot be set on each communication job, but in the general configuration).

partner_communication/models/communication_job.py
```diff
@@ -491,6 +512,78 @@ def reset(self):
...
+            ).message_post_send_sms(sms_text, note_msg=job.subject)
+            job.write(
+                {
+                    "state": "done",
+                    "sent_date": fields.Datetime.now(),
+                    # "sms_cost": ceil(float(len(sms_text))
+                    # // SMS_CHAR_LIMIT) * SMS_COST,
```
ecino: Remove this. We can lose the feature of calculating sms cost if it's too complicated to implement (it's not a very used functionality).

ecino (changes requested):

ecino:
Manually merged.

PR #1888, opened 2024-04-29
reload of view for missing models when sending warning

minedcp:
Fixed the broken code by assuring that the view for sending warning has been correctly loaded using a try catch, the view child_compassion/data/validity_checks_cron.xml has a noupdate which needs to be loaded if non existent for sending the warning. added the reload in partner_communication_reminder/models/compassion_hold.py.


ecino (changes requested):
I don't think this resolves the issue. I don't see how loading some crons in the database would help.

PR #1891, opened 2024-05-06
T1060

lucadela96:
Added business logic to reply to issue about letters on the translation platform (and some changes due to pre-commit).


ecino:
Can you rebase and keep only the changes in `sbc_translation`? If pre-commit fails in other modules, then it's best to do a separate PR (to avoid conflicts when others also run pre-commit on their side).

NoeBerdoz (approved):

NoeBerdoz:
pre-commits errors are not related to the changes, therefore I'm merging the PR

PR #1900, opened 2024-05-23
Fix t1058 sds 18

minedcp:
added text misiing when downloading pdf


sbc_compassion/controllers/download_correspondence.py
```diff
@@ -25,7 +25,7 @@ def generate_correspondence_pdf(self, object_id=None, **parameters):
             raise BadRequest()
 
         correspondence = request.env["correspondence"].browse(int(object_id))
-        binary = correspondence and correspondence.get_image()
+        binary = correspondence and correspondence.get_image() or correspondence.create_text_boxes()
```
ecino: This cannot work, `create_text_boxes` returns a recordset, not a binary file.

minedcp:
removed useless code from generate_correspondance_pdf and updated the create text box function that caused the erasure of original text.

ecino (approved):

PR #1905, opened 2024-05-28
[T1071] - Fix: Search method of sponsorship contract

langjer:
Search method of sponsorship contract is now base only on the commitment number rather than child code and the commitment.

### DIFF
--- a/account_reconcile_compassion/static/src/js/account_move_reconciliation.js
+++ b/account_reconcile_compassion/static/src/js/account_move_reconciliation.js
@@ -477,20 +477,8 @@ odoo.define("account_reconcile_compassion.reconciliation", function (require) {
             }
           });
 
-        var additional_ref = line.st_line.additional_ref;
-        if (additional_ref) {
-          var child_code = additional_ref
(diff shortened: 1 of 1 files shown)


account_reconcile_compassion/static/src/js/account_move_reconciliation.js
```diff
@@ -477,20 +477,8 @@ odoo.define("account_reconcile_compassion.reconciliation", function (require) {
...
-            .match(/\w+/)[0];
-        } else {
-          var child_code = false;
-        }
-
-        var gift_commitment_number_ref_key = line.st_line.payment_ref.substring(
-          19,
-          21
+              var gift_commitment_number_ref_key = line
+                  .st_line.payment_ref.substring(19, 21
```
ecino, line 481: Is this change wanted?

langjer: Yes, it's the removing of lines 480 to 489 (child_code  search part). 
From the rpc query too: `"|",
                ["child_code", "like", child_code],`

PR #1910, opened 2024-06-11
[T1415] - Add: Button "reconcile items" & "1050 items" in  partner view

langjer:


ecino (changes requested):
This change should be moved in compassion-switzerland repo, maybe under accounting_swizterland or partner_compassion module.

PR #1918, opened 2024-06-18
T1389 interaction resume refactor

ecino:
Completely rework interaction.resume feature :

- Use a web page instead of backend view
- FIX sms and phonecalls links with communication jobs


NoeBerdoz (approved):
Massive work in a very short time, well done :sunglasses:

crm_compassion/migrations/14.0.1.2.0/pre-migration.py
```diff
@@ -0,0 +1,9 @@
+def migrate(cr, version):
```
Prazn, line 1: doesn't really change much... but shouldn't this be in a `post-migration` since we are not changing the structure of the database here

interaction_resume/models/abstract_interaction_source.py
```diff
@@ -0,0 +1,95 @@
...
+            ("partner_id", "=", partner.id),
+            ("partner_id.email", "=", partner.email),
+            ("partner_id", "in", partner.other_contact_ids.ids),
+        ]
+
+    def _get_interaction_total(self, search_domain):
+        return self.search_count(search_domain)
+
+    def _get_interaction_direction_domain(self, partner, direction):
+        # Should be implemented by the concrete model
```
Prazn: don't know if it's a good idea, but if we really want the users to implement this method (so that they don't have weird behaviors) should we not raise a `NotImplementedException` ?

interaction_resume/models/mailing_trace.py
```diff
@@ -0,0 +1,78 @@
...
+        self.ensure_one()
+        res = ""
+        if self.mass_mailing_id.body_html:
+            re_pattern = re.compile(r"(\n)+")
+            res = html2plaintext(self.mass_mailing_id.body_html)
+            res = re_pattern.sub("\n", res)[:200] + "..."
+        return res
+
+    def _get_interaction_data(self):
+        """
```
Prazn: personally i would remove this comment since the method name is clear enough ^^

interaction_resume/models/partner_communication.py
```diff
@@ -0,0 +1,58 @@
...
+    _name = "partner.communication.job"
+
+    COMMUNICATION_TYPE_MAPPING = {
+        "digital": "Email",
+        "physical": "Paper",
+        "sms": "SMS",
+    }
+
+    def _get_interaction_data(self):
+        """
```
Prazn: same here, I would remove this comment

interaction_resume/models/sms.py
```diff
@@ -0,0 +1,59 @@
...
+    _inherit = ["sms.sms", "interaction.source"]
+    _name = "sms.sms"
+
+    date = fields.Datetime(related="create_date", search="_search_date")
+
+    def _search_date(self, operator, value):
+        return [("create_date", operator, value)]
+
+    def _get_interaction_data(self):
+        """
```
Prazn: and here ^^ 

But if you think this comment is usefull you can keep it

PR #1922, opened 2024-06-25
computing of attachment correctly updated

minedcp:
added the computing function when new attachments are being set.

### DIFF
--- a/partner_communication/models/communication_job.py
+++ b/partner_communication/models/communication_job.py
@@ -244,7 +244,7 @@ class CommunicationJob(models.Model):
                     }
                 )
             job.ir_attachment_ids = [
-                (6, attachment.id) for attachment in job.ir_attachment_tmp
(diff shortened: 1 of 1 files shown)


partner_communication/models/communication_job.py
```diff
@@ -244,7 +244,7 @@ def _inverse_ir_attachment_tmp(self):
                     }
                 )
             job.ir_attachment_ids = [
-                (4, attachment.id) for attachment in job.ir_attachment_tmp
+                (6, attachment.id) for attachment in job.ir_attachment_tmp
```
ecino: This cannot be correct, I think the old version was correct.

ecino (changes requested):

partner_communication/models/communication_job.py
minedcp: yes you are right i put it back as before.

PR #1924, opened 2024-07-11
Fix t1306 crm teams empty opportunities not shown

clementcharmillot:
### Description

When a CRM stage (column) has 0 opportunities (cards) in it, it is not shown to its belonging sales team members (if a sales team is set in the stage parameters) in the kanban view.

We need to show the stage to the team members even when there is 0 opportunities in it.

It could be related to a wrong XML parameter somewhere on the compassion CRM module.

### How to reproduce

- In CRM Module, create your Sale Team and add your user to it

- On the Pipeline view, create a new column, edit it with the wheel icon, set the Sale Team to yours

- Hard Refresh the page

### DIFF
--- a/crm_compassion/models/crm_lead.py
+++ b/crm_compassion/models/crm_lead.py
@@ -97,9 +97,11 @@ class CrmLead(models.Model):
             search_domain = [
                 *(["|"] * (len(team_id_domain) - 1) + team_id_domain),
             ]
-            res += stages.browse(stages._search(
-                search_domain, order=order, access_rights_uid=SUPERUSER_ID
-            ))
+            res += stages.browse(
+                stages._search(
+                    search_domain, order=order, access_rights_uid=SUPERUSER_ID
+                )
+            )
 
             # Order is lost by the merge and empty stages are pushed back, so we need
             # to sort them.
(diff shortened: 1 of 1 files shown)


crm_compassion/models/crm_lead.py
```diff
@@ -81,3 +81,39 @@ def _compute_name(self):
...
+                *search_domain,
+                *(["|"] * (len(team_id_domain) - 1) + team_id_domain),
+            ]
+
+        stage_ids = stages._search(
+            search_domain, order=order, access_rights_uid=SUPERUSER_ID
+        )
+        return stages.browse(stage_ids) + super(CrmLead, self)._read_group_stage_ids(
+            stages, domain, order
+        )
```
ecino: As the super() method already returns records, can't we simplify the inheritance here? I'm not sure it is equivalent, but something like this?
```suggestion
        res = super()._read_group_stage_ids(stages, domain, order)
        # if the domain contains team_id filters, add them to the search domain
        team_id_domain = [
            cond
            for cond in domain
            if hasattr(cond, "__getitem__") and cond[0] == "team_id"
        ]
        if len(team_id_domain) > 0:
            search_domain = [
                *(["|"] * (len(team_id_domain) - 1) + team_id_domain),
            ]
            res += self.browse(self._search(
                search_domain, order=order, access_rights_uid=SUPERUSER_ID
            ))
        return res
```

PR #1930, opened 2024-07-19
T1630 - Change supporter 'Unreconciled items' filter (1/2)

Prazn:
# Description
In the supporters page, there were two buttons "1050 Items" and "Unreconciled Items" that basically applied the same filters. So we remove the "1050 Items" button and only keep the other one.

# Technical Aspects
* Remove unused code
* Add `parent_state = posted` filter to the button
* Rename the page so that it corresponds better to the button

### DIFF
--- a/sponsorship_compassion/models/res_partner.py
+++ b/sponsorship_compassion/models/res_partner.py
@@ -150,7 +150,7 @@ class ResPartner(models.Model):
                     ("reconciled", "=", False),
                     ("account_id.reconcile", "=", True),
                     ("account_id.code", "=", "1050"),
-                    ("parent_state", "=", "posted")
+                    ("parent_state", "=", "posted"),
                 ]
             )
             partner.receivable_items = move_line_obj.search_count(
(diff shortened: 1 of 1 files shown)


sponsorship_compassion/models/res_partner.py
```diff
@@ -552,3 +549,16 @@ def _updt_invoices_rp(self, vals):
...
+        args,
+        offset=0,
+        limit=None,
+        order=None,
+        count=False,
+        access_rights_uid=None,
+    ):
+        return super(ResPartner, self)._search(
+            args, offset, limit, order, count, access_rights_uid
+        )
```
ecino: I think this can be removed.
```suggestion
```

Prazn: Oops, forgot to remove that

Prazn:
@ecino pre-commit fails on file that I cleaned up in https://github.com/CompassionCH/compassion-modules/pull/1933/commits/9b055a0599bd23ae4dfa2de32edf4b1c4d4f63d7.

I suggest that we ignore the pre-commit here since it will be resolved by the other PR.

PR #1932, opened 2024-07-29
FIX-T1550-bug-unable-to-download-child-picture-in-PDF

Gordack:


ecino (commented):
Shouldn't all the changes be moved into the `partner_communication_compassion` module instead of `child_compassion`? Because the print option would never appear only with `child_compassion` module installed.

Gordack:
> Shouldn't all the changes be moved into the `partner_communication_compassion` module instead of `child_compassion`? Because the print option would never appear only with `child_compassion` module installed.

I agree, but the only component we use from `partner_communication_compassion`in `child_compassion` is the report `child_picture.xml`. So I don't know if it make sense to move all the changes in `partner_communication_compassion` or if it make more sense to move the report `child_picture.xml` from `partner_communication_compassion` to `child_compassion`, as it is made for the childpack ?

ecino:
You are right, moving it to `child_compassion` is also a valid option. But then you need to make sure that all references to the report in the code are renamed `child_compassion.xml_id` instead of `partner_communication_compassion.xml_id`. Either way is fine for me and better than splitting the components between the two modules.

Gordack:
For this to build I had to modify the "report_paperformat_parameter_paperformat_id_fkey" constraint from on delete "Restrict" to on delete "Cascade". Also, I had to modify compassion-switzerland : https://github.com/CompassionCH/compassion-switzerland/pull/1641

PR #1933, opened 2024-07-29
T1629 - New Biennial communication sends old picture of child

Prazn:
# Description
When we recieve a new picture of a child through the `BeneficiaryKit` GMC action, we update the picture of the child and then send a `NewBiennial` communication to the sponsor. However this communication doesn't contain the new picture but instead contains the previous picture.

# Technical Aspects
It seems that even after the child model is updated, the `fullshot` and `portrait` fields still pointed to the previous picture. These fields are `related` to `picture_ids` and for some readon the `picture_ids` list is not up to date when the fields are called to fetch the picture.

I changed these fields to be `computed` instead, when called we sort the `picture_ids` and fetch the latest picture.

### DIFF
--- a/child_compassion/models/child_compassion.py
+++ b/child_compassion/models/child_compassion.py
@@ -352,12 +352,14 @@ class CompassionChild(models.Model):
     @api.depends("pictures_ids")
     def _compute_portrait(self):
         for child in self:
-            child.portrait = child.pictures_ids.sorted()[0].headshot
+            if child.pictures_ids:
+                child.portrait = child.pictures_ids.sorted()[0].headshot
 
     @api.depends("pictures_ids")
     def _compute_fullshot(self):
         for child in self:
-            child.fullshot = child.pictures_ids.sorted()[0].fullshot
+            if child.pictures_ids:
+                child.fullshot = child.pictures_ids.sorted()[0].fullshot
 
     @api.model
     def _available_states(self):
--- a/child_compassion/models/child_compassion.py
+++ b/child_compassion/models/child_compassion.py
@@ -285,8 +285,8 @@ class CompassionChild(models.Model):
         readonly=False,
     )
     household_id = fields.Many2one("compassion.household", "Household", readonly=True)
-    portrait = fields.Image(related="pictures_ids.headshot")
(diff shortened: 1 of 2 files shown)


child_compassion/models/child_compassion.py
```diff
@@ -349,6 +349,18 @@ def _compute_available(self):
         for child in self:
             child.is_available = child.state in self._available_states()
 
+    @api.depends("pictures_ids")
+    def _compute_portrait(self):
+        for child in self:
+            if child.pictures_ids:
+                child.portrait = child.pictures_ids.sorted()[0].headshot
```
ecino: Like this it doesn't fail when no pictures exist.
```suggestion
                child.portrait = child.pictures_ids.sorted()[:1].headshot
```

child_compassion/models/child_compassion.py
```diff
@@ -349,6 +349,18 @@ def _compute_available(self):
...
+
+    @api.depends("pictures_ids")
+    def _compute_fullshot(self):
+        for child in self:
+            if child.pictures_ids:
+                child.fullshot = child.pictures_ids.sorted()[0].fullshot
```
ecino:
```suggestion
                child.fullshot = child.pictures_ids.sorted()[:1].fullshot
```

ecino (changes requested):

child_compassion/models/child_compassion.py
Prazn: From what I've seen, the check if `child.picture_ids` should already test that the list is not empty or `None` (I had to [look it up](https://stackoverflow.com/questions/53513/how-do-i-check-if-a-list-is-empty) because I'm still not used to python), so it should not fail. 

Do you still think that it's better to index with `[:1]` ? I personally feel that it's clearer with `[0]` but it might just be that I'm not used to it.

ecino: `child.pictures_ids` can be an empty recordset which will evaluate to False. But accessing an element in an empty set with an index is not allowed and will raise and IndexError. However, slicing the set will work and return a subset (empty in this case), and then `.headshot` will return False without raising an error.

PR #1934, opened 2024-07-30
T0003 - Refactor 'Lost' leads in CRM view

Prazn:
# Description
In the CRM module, when leads would become 'Lost' they would be archived and have the probability set to 0. The issue with this is that since they were archived, by default they would not appear in the kanban, except when using specific filters which was confusing for the church reps.

We decided to change this behavior, now columns can be defined as 'Lost' columns, any lead inside that column will be 'Lost'. We then also make it so that lost items appear in the kanban. With these two changes the lost leads will remain visible, and how they are handled is more intuitive for the church reps.

# Technical Aspects
* inherited `crm.stage` to add a new `is_lost` field 
* inherited `crm.lead.lost` to add a new `lost_stage_id` field 
* adapted view of `crm.stage` to use new field
* adapted model/view of `crm.lead` to change behavior of `is_lost` by using the new field
* added migration to adapt existing data to use the new field


crm_compassion/migrations/14.0.1.3.0/post-migration.py
Prazn, line 1: @ecino This migration was to make the existing leads more coherent with the new way we handle lost leads, but  I'm not 100% sure we want this. So basically to : 
* Set leads already in a 'lost' stage as 'lost' (so `active = false` and `probability = 0`)
* Move leads that are defined as 'lost' but not in a 'lost' stage to the default 'Lost & Over (unassigned)' stage 

It might be the case that the church reps don't want to change the existing data.

ecino: I think we need and want this migration :+1:

crm_compassion/models/crm_lead.py
```diff
@@ -50,6 +50,24 @@ def create_event(self):
...
+                vals.update({"active": False, "probability": 0})
+            elif is_lost and not stage.is_lost:
+                vals.update({"active": True, "lost_reason": False})
+                lead_probabilities = self._pls_get_naive_bayes_probabilities()
+                if self.id in lead_probabilities:
+                    vals.update({"automated_probability": lead_probabilities[self.id]})
+                    if self.is_automated_probability:
+                        vals.update({"probability": lead_probabilities[self.id]})
+
+        super().write(vals)
```
ecino:
```suggestion
        return super().write(vals)
```

ecino (approved):
The `write` method is supposed to return `True`

crm_compassion/migrations/14.0.1.3.0/pre-migration.py
```diff
@@ -0,0 +1,13 @@
...
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE crm_stage
+        SET is_lost = TRUE
+        WHERE LOWER(name) LIKE '%lost%';
+        """,
+    )
```
ecino: This should also be in post-migrate file, as the first step. Because the column is_lost doesn not yet exist in pre-migration.

ecino (changes requested):

crm_compassion/migrations/14.0.1.3.0/pre-migration.py
Prazn: Oh yeah, you are right. Moved to post-migration

crm_compassion/migrations/14.0.1.3.0/post-migration.py
Prazn: Oki doki

ecino (approved):

PR #1942, opened 2024-08-09
T1669 - attaching a background check and child protection agreement to "personal information"

clementcharmillot:
Adds two new binary fields in `ResPartner` displayed in `Personal Information` under `Child Protection` named `child_protection_agreement`, and `background_check`:
image
## Note
The request hints that existing fields were present in `12.0` for that. I checked all `ResPartners` models in `12.0` and didn't find any.


ecino:
Fields already exist, it's criminal record and child protection. It was a matter of access rights (adding the people in the group). I still have to clarify wheter the child protection date should be writable.

PR #1944, opened 2024-08-12
T1687 - Cannot log calls in interaction resume

clementcharmillot:
## Details
Fix an exception when logging a call from the wizard shown by pressing `Log Call` and submitting a log. 
image
## Solution
- Added back the `communication_id` field on the model `crm_phonecall` as it is the most conservative approach. Field removed in commit fcf0b20.
- If we don't want this field anymore, `call_log` should be edited to not pass a `communication_id` in file `compassion-modules/partner_communication_crm_phone/wizards/call_wizard.py`.

https://github.com/CompassionCH/compassion-modules/blob/8bdb323865ae6e917882d7541a76f675ed1c8d42/partner_communication_crm_phone/wizards/call_wizard.py#L46-L52

### DIFF
--- a/interaction_resume/models/crm_phonecall.py
+++ b/interaction_resume/models/crm_phonecall.py
@@ -1,4 +1,4 @@
-from odoo import fields, models
+from odoo import models
 
 TRACKING_STATUS_MAPPING = {
     "open": "sent",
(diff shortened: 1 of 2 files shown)


ecino:
In fact I would prefer the less conservative approach and remove the field value from the log wizard.

ecino (changes requested):
In fact I would prefer the less conservative approach and remove the field value from the log wizard.

ecino (approved):

PR #1945, opened 2024-08-12
T1690 interaction resume buttons

alexis-allemann:
In the "Interaction resume" view, you need to have selected at least one line in order to see the "Refresh", "Fetch more..." and "Log interaction" action buttons. This represents a UX problem, since these actions are not linked to lines but to the global "res_partner" element.

### DIFF
--- a/child_compassion/models/child_compassion.py
+++ b/child_compassion/models/child_compassion.py
@@ -9,7 +9,7 @@
 ##############################################################################
 import logging
 import traceback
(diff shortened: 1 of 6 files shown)


alexis-allemann:
Proposed solution: 

Action buttons are added in javascript and make RPC or "do_action" calls to trigger the actions defined in the template.

Screenshot from 2024-08-12 15-56-05

ecino (approved):

PR #1946, opened 2024-08-13
T1705 - Regular partners should not be transmitted to GMC

clementcharmillot:
## Change
Only upsert to GMC if the partner has a `global_id`.

## Findings
Partners are upserted to GMC if they are edited. Keep this behavior only if it was upserted before.

## Note
Partners with a `global_id` will continue to send updates to GMC.

### DIFF
--- a/sponsorship_compassion/models/res_partner.py
+++ b/sponsorship_compassion/models/res_partner.py
@@ -240,7 +240,8 @@ class ResPartner(models.Model):
         notify = functools.reduce(
             lambda prev, val: prev or val in vals, notify_vals, False
         )
-        if notify and not self.env.context.get("no_upsert"):
+
+        if notify and self.global_id and not self.env.context.get("no_upsert"):
             self.upsert_constituent()
 
         self._updt_invoices_rp(vals)


ecino (changes requested):
It's not a good change since new sponsors (that have their first sponsorship) have no `global_id` yet and should be transmitted to GMC in order to receive a number. The check should be "has this partner any active sponsorship?".

clementcharmillot:
I thought `upsert_constituent()` is called from elsewhere when starting a sponsorship on the target sponsor, and this code is only used to sync the sponsor changes with GMC after, thus we can check if it was sent before. Is it not the case? If we check for sponsorship and there is a pause between sponsorships, any change during this period won't be sync with GMC. In this case shouldn't it check for both: it has a sponsorship or was sync with GMC (has `global_id`)?

When a contract is activated (or edited), it calls `partner.upset_constituent()`.

https://github.com/CompassionCH/compassion-modules/blob/ac37269341a02d4a5c5da64bb70a5d755204d85a/sponsorship_compassion/models/contracts.py#L820C8-L827C20

ecino (approved):
You are right, I looked too quickly after your changes. In that case I think it's ok the way you did it.

PR #1947, opened 2024-08-15
T1657 - Fix invoice generation for Sub Proposal (3/3)

Prazn:
Follow-up of https://github.com/CompassionCH/compassion-accounting/pull/239

**DO NOT merge this PR until https://github.com/CompassionCH/compassion-accounting/pull/239 is validated.**

### DIFF
--- a/sponsorship_compassion/models/contract_group.py
+++ b/sponsorship_compassion/models/contract_group.py
@@ -70,3 +70,38 @@ class ContractGroup(models.Model):
             if not contract.send_gifts_to
             else contract[contract.send_gifts_to]
         )
+
+    def _should_skip_invoice_generation(self, invoicing_date, contract=None):
+        self.ensure_one()
+
+        if contract is None:
+            return super()._should_skip_invoice_generation(invoicing_date)
+
+        search_filter = [
+            ("state", "!=", "cancel"),
+            ("invoice_date_due", "=", invoicing_date),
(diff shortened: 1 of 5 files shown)


ecino (approved):

sponsorship_compassion/models/contract_group.py
```diff
@@ -70,3 +70,38 @@ def _get_partner_for_contract(self, contract):
...
+            (
+                "line_ids.product_id",
+                "in",
+                contract.product_ids.ids,
+            ),
+        ]
+
+        existing_invoices = self.env["account.move"].search_count(search_filter)
+
+        is_sub_proposal = contract.parent_id.child_id
```
ecino: I can see a problem now. This test will be True for the whole time that the SUB proposal contract is active. However, this check is only useful at the first invoice generation of the contract. Any subsequent generation should behave normally as all other contracts. Could we therefore check that the contract has no invoice yet?
```suggestion
        is_sub_proposal = contract.parent_id.child_id and not contract.invoice_line_ids
```

ecino (changes requested):

sponsorship_compassion/models/contract_group.py
Prazn: Okay, done.

ecino (approved):

PR #1950, opened 2024-08-20
T1706 - Support request partner assignation

alexis-allemann:
In Odoo, there are main contacts and "Linked partners" that have a different email address but are associated with the main contact. These linked partners are usually archived, meaning they don't appear in the lists.

When a sponsor sends an email using a secondary email address, the partner is matched and the related partner is identified. If the found partner is a secondary contact, it's preferable to link the support request to the main partner instead (because all history is maintained there). However, the secondary email address should still be used when replying to the sponsor in the support.

### DIFF
--- a/crm_request/models/request.py
+++ b/crm_request/models/request.py
@@ -166,8 +166,8 @@ class CrmClaim(models.Model):
 
         if "partner_id" not in custom_values:
             match_obj = self.env["res.partner.match"]
-            partner = match_obj.match_values_to_partner(
-                {"email": email_normalize(defaults["email_from"])}, match_create=False
+            partner = match_obj._match_email(
+                {"email": email_normalize(defaults["email_from"])}
             )
             if partner:
                 defaults["partner_id"] = partner.id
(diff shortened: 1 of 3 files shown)


ecino:
I can see the matching changes but not the "reply". Will it work and reply to the secondary email?

alexis-allemann:
@ecino 

> I can see the matching changes but not the "reply". Will it work and reply to the secondary email?

Yes, the responses will indeed be sent to the email specified in the request. This is because the `create` method in [crm_request/models/mail_mail.py](https://github.com/CompassionCH/compassion-modules/blob/14.0/crm_request/models/mail_mail.py) overrides the email with the one provided in the context (specifically, the `use_email_alias` attribute). During the `reply` action in [crm_request/models/request.py](https://github.com/CompassionCH/compassion-modules/blob/2ff83f216539ceacb2e4c31b4fc2257b7f480721/crm_request/models/request.py#L76), you can see that this attribute is set using the email from the `email_from` attribute.

PR #1953, opened 2024-08-21
T1732 - Fix incorrect loading of interactions for partners without email addresses

alexis-allemann:
# Problem
When loading the interaction resume for a partner without a defined email address, the system mistakenly loads interactions for all partners who lack an email address. This behavior results in incorrect data being displayed.

# Solution
To address this issue, a check is introduced to verify if the partner has an email address before attempting to load their interactions. If the partner does not have an email address, the search scope is restricted to prevent loading interactions from other partners without email addresses.

### DIFF
--- a/interaction_resume/models/abstract_interaction_source.py
+++ b/interaction_resume/models/abstract_interaction_source.py
@@ -22,6 +22,8 @@ class InteractionSource(models.AbstractModel):
         @param until:
         """
         search_domain = self._get_interaction_partner_domain(partner)
+        if not search_domain:
+            return True
         records = self.search(
             [
                 *search_domain,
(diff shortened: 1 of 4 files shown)


alexis-allemann:
# ⚠️ Warning

It is necessary to execute the following SQL query on the production database after this PR has been deployed. This query will remove the excess rows in the interaction_resume table and reset the page and date indexes that have already been fetched.

```sql
begin transaction;

update res_partner rp set last_interaction_fetch_date = null, last_interaction_fetch_page = 0 where rp.id in (
	select distinct rp.id from interaction_resume ir join res_partner rp on ir.partner_id = rp.id where rp.email is null
);

delete from interaction_resume ir where ir.partner_id in (
	select distinct rp.id from interaction_resume ir join res_partner rp on ir.partner_id = rp.id where rp.email is null
);

end transaction;
```

ecino (approved):

PR #1954, opened 2024-08-26
T1737 - Constraint Error When Creating/Editing CRM Opportunities Without Customer Email

alexis-allemann:
# Context

When the church reps wants to create or edit an opportunity (crm.lead) and make it with a customer but without an email, there is a constraint error.

Sometimes, they are working with customers that has no emails, they should be able to pass against this email constraint.

# Analysis

The error occurs because the contact is part of an email distribution list. As a result, removing the contact’s email address leads to an error. This is a standard error designed to prevent cascading failures when sending emails through the distribution list.

# Proposed solutions

There are two possible solutions:

1. Keep this protection in place to ensure no errors occur when using distribution lists (current solution).
2. Remove the contact from the distribution list.

In my opinion, we should implement the second solution. If an email address is removed from a contact, it is because it is no longer in use or relevant. Therefore, not sending emails to this address for marketing campaigns should not cause any issues.

### DIFF
--- a/crm_compassion/models/res_partner.py
+++ b/crm_compassion/models/res_partner.py
@@ -8,7 +8,7 @@
 #
 ##############################################################################
 
-from odoo import _, fields, models
+from odoo import api, fields, models
 
 from odoo.addons.partner_communication.models.communication_config import (
     CommunicationConfig,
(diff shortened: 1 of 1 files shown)


ecino (changes requested):
While I'm ok with the proposed fix, I think this has to be moved into a separate module in `compassion-switzerland` repository. I propose to create a new module, call it `mass_mailing_switzerland` for instance and put the code inside it. We can then use this module to add all our custom behaviors related to mass_mailing features.

Furthermore, `@api.constrains` is normally used for checking data and raising issue if needed, not to perform further operations. In that case I would prefer that you inherit the `write` method, check if email is removed and apply the removal of the mailing contacts from there.

alexis-allemann:
Following Emma's comment, issue will be fixed in https://github.com/CompassionCH/compassion-switzerland/pull/1653.

PR #1955, opened 2024-08-27
T1767 - Funds partner matching

clementcharmillot:
- Removes any spaces or dashes around the name when trying to match a partner to an existing one:
  - ` - Jean-Édouard von Burg-- ` -> `Jean-Édouard von Burg`

### DIFF
--- a/partner_auto_match/models/res_partner_match.py
+++ b/partner_auto_match/models/res_partner_match.py
@@ -55,7 +55,7 @@ class ResPartnerMatch(models.AbstractModel):
                 continue
 
         # Postprocess partner (either update or create it depending on context options)
-        if len(partner) == 1 and match_update:
+        if partner and len(partner) == 1 and match_update:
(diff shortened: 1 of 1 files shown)


partner_auto_match/models/res_partner_match.py
```diff
@@ -107,7 +107,7 @@ def update_partner(self, partner, vals, async_mode=True, delay=1):
     @api.model
     def _preprocess_vals(self, vals):
         """Transform, if needed and before matching, the infos received"""
-        pass
+        vals["name"] = vals["name"].strip(" -")
```
ecino: Maybe check that "name" is present, in case other rules don't receive "name" in the vals.
```suggestion
        if "name" in vals:
            vals["name"] = vals["name"].strip(" -")
```

ecino (changes requested):

ecino (approved):

PR #1958, opened 2024-09-02
T1740 - "Correct active text": Body_html cannot be edited.

clementcharmillot:
`raw_template_edit_mode` is always `False` and this is the only place it is used. Couldn't it be removed altogether?


ecino:
@clementcharmillot Yes, please remove it completely. It's an old setting not used anymore.

PR #1959, opened 2024-09-02
T1574 - Fix contact parent (2/2)

Prazn:
follow-up of [PR](https://github.com/CompassionCH/compassion-switzerland/pull/1652)

### DIFF
--- a/partner_communication/views/res_partner_view.xml
+++ b/partner_communication/views/res_partner_view.xml
@@ -35,9 +35,9 @@
         position="before"
       >
                 <label
-            for="parent_id"
-            class="oe_edit_only"
-          >Related Company</label>
+          for="parent_id"
+          class="oe_edit_only"
+        >Related Company</label>
             </xpath>
         </field>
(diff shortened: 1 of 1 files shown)


ecino (commented):
Why not having this change also in `partner_compassion` module? I don't see the point of having this in `partner_communication`, it is totally unrelated in my opinion.

Prazn:
> Why not having this change also in `partner_compassion` module? I don't see the point of having this in `partner_communication`, it is totally unrelated in my opinion.

I agree that it makes sense to do it in `partner_compassion`. What I'm still not sure is should I still do it here since we are fixing something that comes from the base `contacts` module ? Because if let's say we decide not to use the module `partner_compassion`, we would not have this fix anymore.

ecino:
Yes please.

Prazn:
> Yes please.

Alright, published new PR - https://github.com/CompassionCH/compassion-switzerland/pull/1660

PR #1960, opened 2024-09-04
Fix t1710 display correspondent field in survival sponsorship

Gordack:
Add survival sponsorship to sponsorship types


ecino (changes requested):
This has too many side effects because this list is used at different places. It won't behave properly for some cases if survival sponsorships are included in the list.

The appropriate change is to inherit the `res.partner` method `def _compute_related_contracts(self):` in module `survival_sponsorship_compassion` in order to add the survival sponsorships to the relations.

PR #1963, opened 2024-09-04
T1364 - Fix Letter Translation Logic: Hide English Transcription for Sponsors Speaking Source Language

alexis-allemann:
When a B2S letter is received via a GMC message, it sometimes happens that no transcription in the original language is provided, but an English translation is available. Currently, when sending the letter to the sponsor, the system includes the English translation if the sponsor speaks either the original language or English.

However, if the sponsor speaks the original language of the letter, there is no need to display the English translation. If the sponsor speaks neither the original language nor English, then the standard translation process for the letter should be maintained.

This update ensures that English translations are only included when necessary, improving the clarity and relevance of the communication sent to sponsors.

### DIFF
--- a/sbc_translation/models/correspondence.py
+++ b/sbc_translation/models/correspondence.py
@@ -215,7 +215,7 @@ class Correspondence(models.Model):
         needed and upload to translation platform."""
         for letter in self:
             if (
-                (letter.beneficiary_language_ids & letter.supporter_languages_ids)
+                (letter.original_language_id in letter.supporter_languages_ids)
                 or letter.has_valid_language
                 or self.env.context.get("force_publish")
             ):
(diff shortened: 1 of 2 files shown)


sbc_translation/models/correspondence.py
```diff
@@ -215,7 +215,7 @@ def process_letter(self):
         needed and upload to translation platform."""
         for letter in self:
             if (
-                (letter.beneficiary_language_ids & letter.supporter_languages_ids)
+                (letter.original_language_id in letter.supporter_languages_ids)
```
ecino:
```suggestion
                (letter.beneficiary_language_ids & letter.supporter_languages_ids)
```

PR #1964, opened 2024-09-04
T1779 - Remove dupplicate child pictures

Prazn:
# Description
In the `Sponsorship - Certificate` correspondence, we attach the first and last picture of the sponsored child, however there were a couple of case where the pictures were the same.

# Technical Aspects
It seems that in the database there are multiple children that have the same `image_url`, resulting in them having dupplicate images.

A couple of fixed have been made : 
* added `_sql_contraint` and a check when creating a new image to ensure that there is no other image for that child with the same `image_url`
* added check before create that we aren't adding an already existing image (this can be caused for example when a sponsorship is terminated, we update the child information with the current information from GMC, causing dupplicate images)
* added a migration that will clear the dupplicates from the `compassion_child_pictures` table, keeping only the oldest ones.

### DIFF
--- a/child_compassion/models/child_pictures.py
+++ b/child_compassion/models/child_pictures.py
@@ -84,8 +84,9 @@ class ChildPictures(models.Model):
         a new Pictures object. Check if picture is the same as the previous
         and attach the pictures to the last case study.
         """
-        if self._child_picture_already_exists(vals):
-            return False
+        existing_picture = self._child_picture_already_exists(vals)
+        if existing_picture:
+            return existing_picture
 
         pictures = super().create(vals)
         # Retrieve Headshot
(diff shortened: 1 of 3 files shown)


child_compassion/models/child_pictures.py
```diff
@@ -121,6 +131,16 @@ def _find_same_picture(self):
         )
         return same_pics
 
+    def _child_picture_already_exists(self, vals):
+        already_exists = self.search_count(
```
ecino:
```suggestion
        already_exists = self.search
```

child_compassion/models/child_pictures.py
```diff
@@ -121,6 +131,16 @@ def _find_same_picture(self):
...
+            [
+                ("child_id", "=", vals.get("child_id")),
+                ("image_url", "=", vals.get("image_url")),
+            ]
+        )
+
+        return bool(already_exists)
```
ecino:
```suggestion
            ],
            limit=1
        )
        return already_exists
```

child_compassion/models/child_pictures.py
```diff
@@ -76,6 +84,8 @@ def create(self, vals):
         a new Pictures object. Check if picture is the same as the previous
         and attach the pictures to the last case study.
         """
+        if self._child_picture_already_exists(vals):
+            return False
```
ecino:
```suggestion
        existing_picture = self._child_picture_already_exists(vals)
        if existing_picture:
            return existing_picture
```

ecino (changes requested):

ecino (approved):

PR #1965, opened 2024-09-05
T1796 letter comments flooding chatter

clementcharmillot:
Merge chatter's letter comment update messages based on the following rules:
- No message of any kind was posted since the last update message.
- The last update was made by the same user as the new update.
- Forced merge if the last update was less than 30 seconds ago. (To be tweaked).
- The edit is adding text at the end or beginning of a message
image
- The old and new messages have a similarity of at least 0.8 (To be tweaked on feedback) using builtin `SequenceMatcher()`. It catches small typo fixes and small deletions. Potential problem: a comment can be erased silently if the comment is deleted slowly character by character.
### Misc
- Move `sbc_translation` templates in the `data` folder to `templates`.


ecino (approved):

ecino:
@NoeBerdoz I did a quick review, please also check.

NoeBerdoz (commented):
In my opinion, the proposed solution adds unnecessary complexity and technical debt to the already too heavy correspondence model. 

Instead of merging comments in Odoo, we should adjust the platform's autosave behavior to reduce the chatter spam (and as well to reduce unnecessary calls to the backend).

NoeBerdoz:
I will close this PR for the explanation given in my previous comment. 
Something that was good and needs to be redone concerning this PR  is the separation of the template view in a new file `chatter.xml`.

PR #1970, opened 2024-09-23
T1688 - Add Ambassador field in reconciliation

Prazn:
# Description
During the reconciliation process, the accounting team used to be able to define the Ambassador of an `account.move` in odoo 12, but as of now this is not possible anymore.

# Technical Aspects
 Most of the changes is reverting what was done in [this commit](https://github.com/CompassionCH/compassion-modules/commit/f47c0080e98a6db85dc49d4492e5bbbcc61fda40), I did not figure out why it was removed but it seems to work.

### DIFF
--- a/account_reconcile_compassion/models/account_reconcile_model.py
+++ b/account_reconcile_compassion/models/account_reconcile_model.py
@@ -13,6 +13,7 @@ class AccountReconcileModel(models.Model):
     sponsorship_id = fields.Many2one(
         "recurring.contract", "Sponsorship", readonly=False
     )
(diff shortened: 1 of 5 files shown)


account_reconcile_compassion/models/account_reconcile_model.py
```diff
@@ -123,6 +124,8 @@ class AccountReconcileModelLine(models.Model):
         readonly=True,
     )
 
+    # ADD STUFF HERE ?????
+
```
ecino:
```suggestion
```

ecino (approved):

PR #1976, opened 2024-10-11
T1696-language errors in children description

EliasGajo:
I added a controller which allow the user to update the description fields of a child (all the descriptions will be recomputed with the actual translation keys).


ecino (changes requested):
Website routes are not the standard way for letting employees perform some operations, especially when they have no authentication required. It means any person or crawler could call the page and trigger updates, which could rapidly eat a lot of resources. I would prefer a simple server action that can be added somwhere in the menus

child_compassion/models/child_compassion.py
```diff
@@ -537,6 +537,32 @@ def fetch_translations(self):
         self._fetch_translations(self.env.ref("child_compassion.beneficiaries_details"))
         return self.edit_translations()
 
+    @api.model
+    def update_all_children_description(self):
+        children = self.env["compassion.child"].sudo().search([])
```
ecino: We can avoid updating children that are no longer active, it will be a bit quicker.
```suggestion
        children = self.env["compassion.child"].sudo().search([
            ("state", "not in", ["F", "R"])
        ])
```

child_compassion/models/child_compassion.py
```diff
@@ -537,6 +537,32 @@ def fetch_translations(self):
...
+        infos = ""
+        for child in children:
+            error = child.update_child_descriptions()
+            if len(error) > 0:
+                infos += "\n%s" % error
+        if len(infos) > 0:
+            raise UserError(infos)
```
ecino: Maybe launching a background job for each child will avoid a very long process being stuck. The error is returned and will therefore be visible in the job result. Priority 50 is for allowing other jobs to be executed before those.
```suggestion
            child.with_delay(priority=50).update_child_descriptions()
```

ecino (changes requested):

child_compassion/models/child_compassion.py
```diff
@@ -537,6 +537,30 @@ def fetch_translations(self):
         self._fetch_translations(self.env.ref("child_compassion.beneficiaries_details"))
         return self.edit_translations()
 
+    @api.model
+    def update_all_children_description(self):
+        children = self.env["compassion.child"].sudo().search([
+            ("state", "not in", ["F", "R"])
+        ])
+        infos = ""
```
ecino: This is not used.
```suggestion
```

child_compassion/models/child_compassion.py
```diff
@@ -537,6 +537,30 @@ def fetch_translations(self):
...
+            .search([("child_id", "=", self.id)], order="write_date desc", limit=1)
+        )
+        if len(descriptions) > 0:
+            description = descriptions[0]
+            description._generate_all_translations()
+            return ""
+        return "Error with %s (%s) : description not found" % (
+            self.name if self.name else "",
+            self.id,
+        )
```
ecino: compassion.child.description is a wizard. The table is therefore most of the time empty. The process for generating a description is simply to create one record in this table. I know it's a bit an akward design but this is a legacy functionality we may once rethink better.
```suggestion
        self.ensure_one()
        return self.env["compassion.child.description"].create({
            "child_id": self.id,
        })
```

ecino (changes requested):

ecino (approved):

PR #1977, opened 2024-10-15
T1217 : pre migration script

EliasGajo:
The translator field is deleted everywhere and in the database.
The migration script try to resolve all the missing links between res.partner and translation.user. If and only if all the missing links are resolved, the translator_id field is deleted from the database.

### DIFF
--- a/sbc_compassion/migrations/14.0.1.0.3/pre-migration.py
+++ b/sbc_compassion/migrations/14.0.1.0.3/pre-migration.py
@@ -1,0 +1,67 @@
+import logging
+from openupgradelib import openupgrade
+
+_logger = logging.getLogger(__name__)
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    cr = env.cr
+    cr.execute(
+        """
+        SELECT DISTINCT(id) FROM res_partner 
+        WHERE translation_user_id IS NULL 
+        AND id IN (
+            SELECT DISTINCT(translator_id) FROM correspondence c
+            WHERE c.translator_id IS NOT NULL
+            AND c.new_translator_id IS NOT NULL
+        );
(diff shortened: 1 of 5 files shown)


sbc_compassion/migrations/14.0.1.0.3/pre-migration.py
```diff
@@ -0,0 +1,26 @@
...
+        """
+    )
+
+    # Remove the column translator which is unused from 2016
+    cr.execute(
+        """
+        ALTER TABLE correspondence
+            DROP COLUMN IF EXISTS translator;
+        """
+    )
```
ecino, line 26:
```suggestion
```

sbc_compassion/models/correspondence.py
```diff
@@ -221,7 +221,6 @@ class Correspondence(models.Model):
     original_letter_url = fields.Char()
     final_letter_url = fields.Char()
     import_id = fields.Many2one("import.letters.history", readonly=False)
-    translator = fields.Char()
```
ecino, line 224: We need it

sbc_compassion/views/correspondence_view.xml
```diff
@@ -166,10 +166,6 @@
                         </group>
                         <group>
                             <field
-                name="translator"
-                attrs="{'invisible': [('translator', '=', False)]}"
-              />
-                            <field
```
ecino, line 172: We need it

ecino (changes requested):
We need the translator field. It's used to store the translator name of the Field Office worker. OK, this information is probably useless for us. In that case, we want to remove it from the GMC mapping.

image

ecino:
Not needed.

PR #1979, opened 2024-10-15
T0024 - Allow passing arbitrary data to email template through context

nlachat-compassion:
This PR adds functionality to `partner_communication/models/communication_job.py`:`CommunicationJob.refresh_text` to pass arbitrary data to email templates from the context.
This is required to implement https://github.com/CompassionCH/compassion-switzerland/pull/1663 without computing the list of partners to archive twice (once in the cron job to determine if an email should be sent, and once in the email template to display the list of partners).

As I am new to odoo, maybe there is a simpler/cleaner way to do what I want to do. If so, this PR is maybe redundant.

If accepted, this PR must be merged **BEFORE** https://github.com/CompassionCH/compassion-switzerland/pull/1663

### DIFF
--- a/partner_communication/models/communication_job.py
+++ b/partner_communication/models/communication_job.py
@@ -637,7 +637,9 @@ class CommunicationJob(models.Model):
             if template_id and job.object_ids:
                 try:
                     fields = template_id.with_context(
-                        template_preview_lang=lang
+                        template_preview_lang=lang,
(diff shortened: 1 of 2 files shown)


nlachat-compassion:
See https://github.com/CompassionCH/compassion-switzerland/pull/1663#issuecomment-2482712915 for why this was closed

PR #1983, opened 2024-10-17
[MIG] Rr/17.0 gift compassion

robinrolle:
CODE NOT TESTED LOCALLY due to lack of data.

Migration changes for gift_compassion module.

Changes required by OCA as been applied

Refactored data/base_automation.xml. Fields state and code from base.automation are not supported anymore.

Model account.analytic.tag is not supported as well and has been replaced by account.analytic.distribution.


ecino:
In the file `sponsoship_gift.py` the analytic tags are used but the code is left the same as in v14. I think it needs to be adapted to the changes in v17 to the distribution. We may also need a migration script in order to FIX  the ids of the settings no, or at least unset them so that we can reset them properly after migration?

PR #1985, opened 2024-10-21
Christmas gift can be generated even without being active

davidwul:
Christmas gift being a fund does not need to wait for being activate before generating invoices. 
This change is already on the nordic prod

### DIFF
--- a/sponsorship_compassion/models/contract_group.py
+++ b/sponsorship_compassion/models/contract_group.py
@@ -43,10 +43,11 @@ class ContractGroup(models.Model):
         super(
             ContractGroup, self.with_context(open_invoices_sponsorship_only=True)
         )._generate_invoices(invoicer, contract_id)
+        # We can generate christmas even if the contract isn't active as it is a fund.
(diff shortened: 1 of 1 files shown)


sponsorship_compassion/models/contract_group.py
```diff
@@ -43,10 +43,11 @@ def _generate_invoices(self, invoicer, contract_id=None):
         super(
             ContractGroup, self.with_context(open_invoices_sponsorship_only=True)
         )._generate_invoices(invoicer, contract_id)
+        # We can generate christmas even if the contract isn't active as it is a fund.
+        self.mapped("contract_ids")._generate_gifts(invoicer, CHRISTMAS_GIFT)
```
ecino, line 47: This would also generate gifts for terminated contracts, do we want to avoid it? I would propose to keep a filter but maybe only keep `waiting` and `active` contracts.

ecino (changes requested):

sponsorship_compassion/models/contract_group.py
ecino: @davidwul What do you think?

PR #1986, opened 2024-10-25
T1894 improve error messages on scan import letters

jordyBSK:
improve error messages on scan import letters

### DIFF
--- a/sbc_compassion/models/import_letters_history.py
+++ b/sbc_compassion/models/import_letters_history.py
@@ -300,11 +300,13 @@ class ImportLettersHistory(models.Model):
             # pylint: disable=invalid-commit
             self._cr.commit()
         except Exception:
-            self.write({
-                "import_completed": False,
-                "failed_file_name": file_name,
-                "state": "failed"
-            })
+            self.write(
+                {
+                    "import_completed": False,
+                    "failed_file_name": file_name,
+                    "state": "failed",
+                }
+            )
             self.env.user.notify_danger(f"Couldn't import file {file_name}")
         else:
             self.env.user.notify_success("Letters import completed !")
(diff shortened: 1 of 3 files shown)


sbc_compassion/models/import_letters_history.py
```diff
@@ -274,10 +306,9 @@ def _analyze_pdf(self, pdf_data, file_name):
...
-            self.env.user.notify_danger(message)
-            logger.error(
-                message,
-                exc_info=True,
-            )
+            self.import_completed = False
+            self.failed_file_name = file_name
+            self.state = "failed"
+            self.env.user.notify_danger(f"Couldn't import file {file_name}")
+            self._compute_state = False
```
ecino: A call to the write method is preferable when changing many values at the same time. Also assigning the compute method seems wrong.
```suggestion
            self.write({
                "import_completed": False,
                "failed_file_name": file_name,
                "state": "failed"
            })
            self.env.user.notify_danger(f"Couldn't import file {file_name}")
```

ecino (changes requested):

sbc_compassion/models/import_letters_history.py
jordyBSK: when I remove the line "self._compute_state = False" the import is recalculated and the import state is set to ready instead of failed

ecino: Can you then please put this as a comment above the line in order to explain the reason of this line? Otherwise won't understand and may remove it in the future.

jordyBSK:
there was an error when testing the pre-commit on files in the addons folder that I didn't modify

ecino (approved):

PR #1988, opened 2024-10-30
Port the gift with no invoice ability from 12

davidwul:
TODO: 
- add field in view
- check sub sponsorship "inherit" the no_birthday_invoice

### DIFF
--- a/gift_compassion/wizards/generate_gift_wizard.py
+++ b/gift_compassion/wizards/generate_gift_wizard.py
@@ -8,11 +8,14 @@
 #
 ##############################################################################
 import logging
-
-from odoo import _, fields, models
-from odoo.addons.sponsorship_compassion.models.product_names import GIFT_PRODUCTS_REF
 from datetime import datetime
+
 from dateutil.relativedelta import relativedelta
+
+from odoo import models
+
+from odoo.addons.sponsorship_compassion.models.product_names import GIFT_PRODUCTS_REF
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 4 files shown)


NoeBerdoz:
the subsponsorship should inherit from the no_birthday_invoice as it copy the recurring contracts (and change a few value sduring the copy)
[Related code here](https://github.com/CompassionCH/compassion-modules/blob/70e43f4c2f5077e15df9605a192224344a3a1ad7/sponsorship_sub_management/wizards/sub_sponsorship_wizard.py#L49)

gift_compassion/wizards/generate_gift_wizard.py
```diff
@@ -0,0 +1,76 @@
...
+                # Create a gift record
+                gift_vals["sponsorship_id"] = self.contract_id.id
+                gift_vals["date_partner_paid"] = gift_date
+                gift_vals["gift_date"] = gift_date
+                gift_vals["amount"] = self.contract_id.birthday_invoice
+                gift_obj.create(gift_vals)
+        else:
+            super().generate_invoice(due_date)
+
+    def compute_date_birthday_invoice(self, child_birthdate, payment_date=datetime.today()):
```
NoeBerdoz: Default arguments are evaluated only once, when the function is defined, not each time the function is called.

```suggestion
    def compute_date_birthday_invoice(self, child_birthdate, payment_date=None):
        """Set date of invoice two months before child's birthdate"""
        if payment_date is None:
        payment_date = datetime.today()
```

NoeBerdoz (changes requested):

NoeBerdoz (approved):

PR #1989, opened 2024-10-30
T1951 : fix b2s_state error

EliasGajo:


ecino:
This is not a desired change. See comment in the related task.

PR #1991, opened 2024-10-31
T1676 : update paragraphs management

EliasGajo:


sbc_translation/views/correspondence_view.xml
```diff
@@ -85,7 +85,7 @@
                     </group>
                     <group string="Letter content">
                         <field name="unread_comments"/>
-                        <field name="paragraph_ids">
+                        <field name="paragraph_ids" invisible="1">
```
ecino, line 88: Is this view change needed? I fear that it's not something we want and that it will be less informative in the new way. Can we keep only the code change in python?

EliasGajo: The paragraph_ids field is necessary for the button "Reply to comments". Indeed, this button set the default_paragraph_ids to the paragraph_ids field in the view.
The simplest solution for me was to put the paragraph_ids field to invisible so that I don't break the "Reply to comments" feature.

ecino (approved):

PR #1992, opened 2024-11-01
T1924 wp sponsorships not visible in my compassion

nlachat-compassion:
Fixes the missing contract lines for SWP contracts.

### DIFF
--- a/sponsorship_compassion/models/contracts.py
+++ b/sponsorship_compassion/models/contracts.py
@@ -301,23 +301,6 @@ class SponsorshipContract(models.Model):
         for contract in self:
             contract.fully_managed = contract.partner_id == contract.correspondent_id
 
-    def is_correspondent_WP(self):
-        """
-        Does the correspondent have the W&P category ?
-        """
-        self.ensure_one()
-        return "W&P" in self.correspondent_id.category_id.mapped("name")
-
-    @api.onchange("correspondent_id")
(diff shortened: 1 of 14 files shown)


sponsorship_compassion/models/contracts.py
```diff
@@ -1350,3 +1352,27 @@ def migrate_gmc_correspondence_commitment(self):
...
+            list: fixed, previously inconsistent, contracts
+        """
+
+        inconsistent_contracts = self.search([("type", "=", "SWP")])
+        for contract in inconsistent_contracts:
+            if len(contract.contract_line_ids) == 0:
+                # Fix missing contract lines
+                contract._create_empty_lines_for_correspondence()
+                # Update contracts on the partner side
+                contract.correspondent_id._compute_related_contracts()
```
nlachat-compassion, line 1371: HI @ecino ,
I think I finally understood what could be the real cause of the issue after our discussion of yesterday.
Given that the type of the Write&Pray contracts are correctly set, it is still possible that those contracts do not have any contract lines, which causes them to not appear on mycompassion.
What I propose here is to find all those contracts and update them with default lines.
Do you think my reasoning is correct ? For me, it is the most likely explanation, given also the screenshots which were attached to the initial ticket description (showing the empty contract lines).

ecino (approved):

PR #1993, opened 2024-11-01
T1816 : automatic call activity for "Sponsorship Activation Reminder 1"

EliasGajo:


ecino:
The change is not needed. It can be done directly in the communication config record.
image

PR #1994, opened 2024-11-05
T1715 - FCPs are wrongly marked as suspended

nlachat-compassion:
Implements a cron job to fetch the updated lifecycle events and details from FCPs from the GMC api.
I implemented unit tests to test the logic (by mocking the network requests) and performed some manual tests against the staging API (modifying the cron job to : `model.sync_projects_from_gmc(requests_throttle_seconds=1.0, log_period=1, max_projects_to_sync=10)`)


nlachat-compassion:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=CompassionCH_compassion-modules&pullRequest=1994) **Quality Gate failed**
> 
> Failed conditions  [2 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=CompassionCH_compassion-modules&pullRequest=1994&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)
> 
> [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=CompassionCH_compassion-modules&pullRequest=1994)

These are false positives. The generated random numbers are only used for testing

sponsorship_compassion/models/contract_group.py
```diff
@@ -67,7 +67,7 @@ def build_inv_line_data(
     def _get_partner_for_contract(self, contract, gift_wizard=False):
         if gift_wizard and contract.send_gifts_to:
             return contract[contract.send_gifts_to]
-        return super()._get_partner_for_contract(contract, gift_wizard)
+        return super()._get_partner_for_contract(contract)
```
NoeBerdoz: What does this change stands for?

child_compassion/models/project_compassion.py
```diff
@@ -733,6 +734,61 @@ def reactivate_project(self):
...
+        if max_projects_to_sync == 0:
+            raise ValueError(
+                f"Invalid parameter value {max_projects_to_sync=}. "
+                "Should be either -1 or strictly positive."
+            )
+        max_projects_to_sync = min(max_projects_to_sync, len(projects_to_sync))
+        if max_projects_to_sync > 0:
+            projects_to_sync = projects_to_sync[:max_projects_to_sync]
+
+        nb_projects_to_sync = len(projects_to_sync)
```
ecino: I would suggest a few changes in order to reduce a bit the number of projects taken into account. Why not focusing on those that were not updated since one month? I would also change the recurring interval of the CRON to one month as well since it's a very costly method.
```suggestion
        one_month_ago = fields.Date.today() - relativedelta(days=30)
        projects_to_sync = self.search([
            ("status", "in", ["Active", "Suspended"]),
            ("last_update_date", "<", one_month_ago)
            ("sponsorships_count", ">", 0)
        ], limit=max_projects_to_sync)
        nb_projects_to_sync = len(projects_to_sync)
```

For this to work you also need to move all these additions in the module `sponsorship_compassion` instead of `child_compassion`, because the field `sponsorships_count` is only available there. You also need to make the field searchable by adding the `search` attribute or by storing it database (I would prefer the first option).

In that case also the parameter `max_projects_to_sync` is no longer tested but passed directly in the search method. An error will be raised from there if it's invalid.

child_compassion/models/project_compassion.py
```diff
@@ -733,6 +734,61 @@ def reactivate_project(self):
             )
         return True
 
+    def sync_projects_from_gmc(
+        self, requests_throttle_seconds=1.0, log_period=100, max_projects_to_sync=-1
```
```suggestion
        self, requests_throttle_seconds=1.0, log_period=100, max_projects_to_sync=None
```

sponsorship_compassion/models/contract_group.py
ecino: I don't think it's a valid change, or what was your intent here?

ecino (changes requested):

sponsorship_compassion/models/contract_group.py
nlachat-compassion: It was to fix some unit test, but it's indeed a mistake (I forgot to upgrade another module when running the test). I'll revert this change

ecino (approved):

ecino (approved):

nlachat-compassion:
@ecino I wasn't done fixing this PR. Sorry if it wasn't clear. 
I still need to fix:
`2024-11-19 08:14:37,113 30255 ERROR unit_test_sponsorship_compassion odoo.osv.expression: Non-stored field compassion.project.sponsorships_count cannot be searched. `
Can we revert the merge until I'm done?

PR #1995, opened 2024-11-07
T1914 : remove the self.env.clear() in the count_pdf_page function

EliasGajo:


ecino (changes requested):
Thank you for the great analysis of the problem. In fact I think that the first solution is better and that the clear() call is wrong in the `count_pdf_page`. Can you please update your PR with that instead?

PR #1997, opened 2024-11-19
T1715 FCPs are wrongly marked as suspended: after merge fixes

nlachat-compassion:


nlachat-compassion:
I think that this PR addresses all your comments @ecino  ( https://github.com/CompassionCH/compassion-modules/pull/1994#discussion_r1846348795 ). 
I did not find a way to make the field sponsorships_count searchable (the search function can only transform one search domain into another, so I don't see how it could count. Maybe there's a possibility in odoo that I don't know about. The other alternative, setting store=True, seems to work however).
What do you think?

ecino (approved):

PR #1998, opened 2024-11-20
T1996 add a link for position in google maps

jordyBSK:


NoeBerdoz (changes requested):
Great job, the only remaining thing to do is to declare the  'show in Google maps' string to be translatable (hint: `_`)

child_compassion/models/project_compassion.py
```diff
@@ -531,6 +532,14 @@ def _compute_timezone(self):
                 lng=project.gps_longitude, lat=project.gps_latitude
             )
 
+    def _compute_google_link(self):
+        for project in self:
+            if project.gps_latitude and project.gps_longitude:
+                google_url = f"https://www.google.com/maps/search/?api=1&query={project.gps_latitude},{project.gps_longitude}"
+                project.google_link = (
+                    f'<a href="{google_url}" target="_blank">' "show in Google maps</a>"
+                )
```
```suggestion
               project.google_link = f"https://www.google.com/maps/search/?api=1&query={project.gps_latitude},{project.gps_longitude}"
            else:
                project.google_link = False

    def open_google_maps(self):
        self.ensure_one()
        return {
            "type": "ir.actions.act_url",
            "url": self.google_link,
            "target": "new",
        }
```

child_compassion/views/project_compassion_view.xml
```diff
@@ -141,6 +141,7 @@
                                     <field name="time_to_medical_facility" />
                                     <field name="gps_latitude" />
                                     <field name="gps_longitude" />
+                                    <field name="google_link" widget="html" />
```
ecino: Please make a button at the top of the form view instead of a field. It will be better accessible.

```
<button
              name="open_google_maps"
              type="object"
              string="Google Maps"
              icon="fa-external-link"
              class="oe_stat_button"
            />
```

ecino (changes requested):
I propose to add a button at the top instead of an URL field.

ecino (approved):

PR #1999, opened 2024-11-21
T2007 : translation letter counting

EliasGajo:


sbc_translation/wizards/translation_letter_counting.py
```diff
@@ -0,0 +1,73 @@
...
+    def search_active_wizard(self):
+        # Normally there should be maximum 1 wizard by user in the database.
+        # In case of multiple wizard in the database, we consider only the first one
+        wizards = self.env['translation.letter.counting.wizard'].search([
+            ('create_uid', '=', self.env.uid)
+        ])
+        if len(wizards) > 0:
+            return wizards[0]
+        else:
+            return False
```
NoeBerdoz: This could be simplified like this:
```suggestion
        # Normally there should be maximum 1 wizard by user in the database.
        # In case of multiple wizard in the database, we consider only the first one
        return self.search([('create_uid', '=', self.env.uid)], limit=1)
```

sbc_translation/wizards/translation_letter_counting_view.xml
```diff
@@ -0,0 +1,40 @@
...
+
+    <!-- Menu Definition -->
+    <menuitem
+    id="translation_letter_counting_menu"
+    parent="sbc_compassion.menu_correspondence"
+    name="Counting of translation letters"
+    action="action_translation_letter_counting"
+    />
+
+    <template id="asset_backend" name="test Elias" inherit_id="web.assets_backend">
```
NoeBerdoz: "test Elias" is not an appropriate template name ;)

sbc_translation/security/ir.model.access.csv
```diff
@@ -12,4 +12,5 @@ access_admin_translator_competence,Full access on translator competence,model_tr
 access_translation_competence,Read access on translation competence,model_translation_competence,group_user,1,0,0,0
 access_admin_translation_competence_user,Full access on translation competence,model_translation_competence,group_manager,1,1,1,1
 access_model_sbc_reply_to_comments_wizard,Full access to reply to comments,model_sbc_reply_to_comments_wizard,group_manager,1,1,1,1
-access_model_sbc_reply_to_issue_wizard,Full access to reply to issue,model_sbc_reply_to_issue_wizard,group_manager,1,1,1,1
\ No newline at end of file
+access_model_sbc_reply_to_issue_wizard,Full access to reply to issue,model_sbc_reply_to_issue_wizard,group_manager,1,1,1,1
+access_model_translation_letter_counting_wizard,Access on Translation letter counting,model_translation_letter_counting_wizard,group_user,1,1,1,1
```
NoeBerdoz: As discussed, make sure the Base User has access to it, not the Translation Platform / User

NoeBerdoz (changes requested):

NoeBerdoz:
@ecino @EliasGajo I think a bus implementation would have been a better approach for real-time updates. However, in my opinion we can stick to this current implementation for simplicity and fast delivery to our internal customer.
[Odoo bus library](https://www.cybrosys.com/blog/everything-you-want-to-know-about-odoo-bus-library)

NoeBerdoz (approved):

PR #2004, opened 2024-11-29
[T1906] FEAT: Add new communication config & email template for MyCompassion user signup 1/2

robinrolle:
Add the communication rules config and the email used the new MyCompassion signup.

This PR should be merged before 
[T1906] FEAT: Add new MyCompassion user signup logic using OCA module 2/2

Because it relies on this communication rule.


ecino:
In fact you can move this data directly in the `my_compassion` module, because it's not used anywhere else.

PR #2005, opened 2024-11-29
T1916 : Manage Christmas letters

EliasGajo:


ecino (changes requested):
See task comments. Needs improvements.

PR #2012, opened 2025-01-20
[T2149] : Fix for unnecessary picture updates log when printing

AndiR27:
printing a child picture no longer triggers unintended update logs or communications. 
The process now ensures that printing does not initiate a picture update unless required.

Must test with print childpack to verify that no logs or communications are generated during printing ! (wizard "print_childpack.py" was not changed, yet during my tests, update log were not showing anymore)


ecino (approved):
I improved the code a bit.

PR #2017, opened 2025-02-07
[T1995] FEAT: Add user in the interaction resume

auclerc:


ecino (changes requested):
Changes are good. But you should also add the `user_id` information in the `crm_request`, `other_interaction`, and `sms` objects.

ecino (approved):

PR #2026, opened 2025-04-03
T2039 project tags

Nath-S07:
AFTER PUTTING THIS SNIPPET IN PRODUCTION IT WILL CHANGE SOME COLOURS THAT ARE ALREADY IN EFFECT FOR TAGS/TASKS -> BECAUSE THERE WERE SOME COLOURS THAT HAD 2 DIFFERENT IDS, I REMOVED THEM AVOID REDUNDANCY. The extension of colours (not present in the odoo default addons) was being duplicated as it was extended in a for loop that was used to fetch the default odoo colours. The extension of colours was white (blank) because the logic to add the preview of the extended colours was only added for the colorpicker when you edit a colour (tag_color_x) and not for when you create a tag (oe_kanban_color_x). Now everything should work, there are no more duplicates and you can see all the previews of the colours.

### DIFF
--- a/crm_compassion/static/src/xml/kanban_colors.xml
+++ b/crm_compassion/static/src/xml/kanban_colors.xml
@@ -3,7 +3,7 @@
         <t t-jquery="t[t-set='colornames']" t-operation="attributes">
             <attribute
         name="value"
-      >['No color', 'Red', 'Orange', 'Yellow', 'Light blue', 'Dark purple', 'Salmon pink', 'Medium blue', 'Dark blue', 'Fushia', 'Green', 'Purple', 'Dark red', 'Blue', 'Red', 'Brown', 'Green', 'Light yellow', 'Beige', 'Pastel green', 'Tommy blue', 'Dark grey', 'Dark violet', 'Dark brown', 'Greyish blue']</attribute>
+      >['Dark red', 'Blue', 'Brown', 'Light yellow', 'Beige', 'Pastel green', 'Tommy blue', 'Dark grey', 'Dark violet', 'Dark brown', 'Greyish blue']</attribute>
         </t>
     </t>
     <t t-extend="FieldMany2ManyTag.colorpicker">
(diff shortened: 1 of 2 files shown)


NoeBerdoz (changes requested):
Thank you for providing the fix, but there are modification not related to the task like depart_communications.xml and child_protection related changes. Please make this PR concern only the tags fix

crm_compassion/static/src/xml/kanban_colors.xml
```diff
@@ -3,7 +3,7 @@
         <t t-jquery="t[t-set='colornames']" t-operation="attributes">
             <attribute
         name="value"
-      >['No color', 'Red', 'Orange', 'Yellow', 'Light blue', 'Dark purple', 'Salmon pink', 'Medium blue', 'Dark blue', 'Fushia', 'Green', 'Purple', 'Dark red', 'Blue', 'Red', 'Brown', 'Green', 'Light yellow', 'Beige', 'Pastel green', 'Tommy blue', 'Dark grey', 'Dark violet', 'Dark brown', 'Greyish blue']</attribute>
```
NoeBerdoz, line 6: What was the use of No color, are you sure it is safe to remove it?
What if we want a tag without a color

Nath-S07: Normally yes, because the colors I removed from the list were the colors already present in the base.xml : 
`<t t-set="colornames" t-value="['No color', 'Red', 'Orange', 'Yellow', 'Light blue', 'Dark purple', 'Salmon pink', 'Medium blue', 'Dark blue', 'Fushia', 'Green', 'Purple']"/>`
        
So, the no color was already present I did that remove the redundancy in the list and so that the color preview matches the correct color id.

NoeBerdoz: Alright, that makes sense, thank you

NoeBerdoz (approved):

ecino (approved):
Only Daniel uses many colors and asked that we have more than what Odoo provides in standard. It was broken since v14 and I think the change will be welcome even if we lose some assigned colors.

PR #2028, opened 2025-04-09
T2288 Feat: Max pages for a letter

Mcentamo:
Context:

Maximum number of pages for a letter is 15 pages, this is a must follow from Compassion International.

Todo:

Add error message to the user when his letter is more than 15 pages, attachments included.


ecino:
Doing it in the preview is probably a bad place. Let's see if this can be checked before.

PR #2031, opened 2025-05-02
[T2370] fixed church name editing error

Nath-S07:
The issue occurred when editing the name of certain Church records the def _compute_opportunity_count would run while some of the associated partner records still had temporary Odoo IDs (-> NewId_xxxx), which are strings rather than integers. This caused a type mismatch during the domain search, resulting in an error. The root cause is timing-related the compute method was being triggered before the record was fully saved.

To fix this, I added a condition to skip any partner with a non-integer ID. This ensures only fully saved records are processed, avoiding the error.

### DIFF
--- a/crm_compassion/models/res_partner.py
+++ b/crm_compassion/models/res_partner.py
@@ -67,7 +67,8 @@ class Partner(models.Model):
         super()._compute_opportunity_count()
         for partner in self:
             operator = "child_of" if partner.is_company else "="
-            # we use partner.ids[0] because when editing a church's name the id becomes temporarily a NewId and so an error is produced as the method uses child_of with a NewId and not the original id
+            # Using partner.ids[0] is a trick to avoid error with child_of
+            # when partner has a temporary NewId
             partner.opportunity_count += self.env["crm.lead"].search_count(
                 [
                     ("partner_id", operator, partner.ids[0]),
--- a/crm_compassion/models/res_partner.py
+++ b/crm_compassion/models/res_partner.py
@@ -66,14 +66,14 @@ class Partner(models.Model):
     def _compute_opportunity_count(self):
         super()._compute_opportunity_count()
         for partner in self:
-                operator = "child_of" if partner.is_company else "="
-                partner.opportunity_count += self.env["crm.lead"].search_count(
-                    [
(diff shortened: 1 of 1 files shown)


NoeBerdoz (changes requested):
The fix doesn't handle the core issue, NewId should be handled correctly, maybe the issue is deeper than we thought. It looks like that the issue appears when there is a company type member in the church's members

Nath-S07:
Following Noé's comments I changed the code to make sure that for each partner the correct ids are always taken even when updating the name (when the id for a partner becomes NewId temporarily). Digging deeper in this code snippet and discussing what I found with Noé, we came to the conclusion that the logic behind the opportunity count might not be up to date and a bit incorrect. A new task will be created to make sure everything is correct with this logic.

crm_compassion/models/res_partner.py
```diff
@@ -69,7 +69,7 @@ def _compute_opportunity_count(self):
             operator = "child_of" if partner.is_company else "="
             partner.opportunity_count += self.env["crm.lead"].search_count(
                 [
-                    ("partner_id", operator, partner.id),
+                    ("partner_id", operator, partner.ids[0]),
```
NoeBerdoz, line 74: As discussed, this is a fix that doesn't address the core issue, let's put a comment for other developers reading this to understand that `ids[0]`it is a way to get an id even in the case of generics NewId.

NoeBerdoz (changes requested):

NoeBerdoz (approved):

crm_compassion/models/res_partner.py
```diff
@@ -67,9 +67,10 @@ def _compute_opportunity_count(self):
         super()._compute_opportunity_count()
         for partner in self:
             operator = "child_of" if partner.is_company else "="
+            # we use partner.ids[0] because when editing a church's name the id becomes temporarily a NewId and so an error is produced as the method uses child_of with a NewId and not the original id
```
NoeBerdoz:
```suggestion
            # Using partner.ids[0] is a trick to avoid error with child_of
            # when partner has a temporary NewId
```

NoeBerdoz (changes requested):

PR #2033, opened 2025-05-23
remove the covid status (not maintained since more than two years!

davidwul:
add bool to be able to filter project suspended for more than two months without reactivation since


ecino:
Manually merged.

PR #2034, opened 2025-05-26
T2257 FIX gifts problem: sometimes, the name & address of sponsor appears in "instructions"

SamBachmann:
In this PR, I remove the line `"instructions": invoice_line.move_id.narration` from `get_gift_values_from_product`
The field narration of `invoice_line.move_id` may contain the name and the address. There's few cases where it would contain relevant information.

(More details about what I've found on Odoo)


davidwul:
It's a radical solution, in some case, the instruction are really needed. 
you can though leave this ticket as it as been already solved another way,

PR #2035, opened 2025-05-27
T2332 FIX Payment Reminder 2 & 3 generated at the same time

SamBachmann:
To fix the issue, I extended the logic used to handle the 1st reminder for the 2nd and 3rd reminder.
In more details:

In the original code, we had this :
```python
if has_first_reminder:
                has_second_reminder = partnerCommunicationJob.search_count(
                    reminder_search
                    + [
                        ("sent_date", ">=", older_threshold),
                        ("sent_date", "<", twenty_days_ago),
                        ("config_id", "=", reminder_confs[1].id),
                    ]
                )
                if has_second_reminder:
                    eligible_reminders["third"] += sponsorship
                eligible_reminders["second"] += sponsorship
 ```

This seems incorrect to me. The `if has_second_reminder:` clause is true iff a 2nd reminder have been sent more than 20 days ago. So if it  has been sent 10 days ago, we would simply send it again. Moreover, when creating the 3rd reminder, we also create a 2nd reminder every time due to the lack of `else` clause.

I have not find a way to test the create_reminder_communication method properly. What I have done is only based on code observation.

### DIFF
--- a/partner_communication_reminder/models/recurring_contract.py
+++ b/partner_communication_reminder/models/recurring_contract.py
@@ -43,10 +43,7 @@ class RecurringContract(models.Model):
     def create_reminder_communication(self):
         """Creation of the reminder for active and waiting contracts"""
         _logger.info("Creating Sponsorship Reminders")
-        # Change today() for debugging purposes
-        # today = datetime.now()
-        # set today to 1st october 2023 for testing
-        today = datetime(2024, 11, 15)
+        today = datetime.now()
         first_day_of_month = today.replace(day=1)
         reminder_confs = self.env["partner.communication.config"]
         for i in range(1, 4):
--- a/partner_communication_reminder/models/recurring_contract.py
+++ b/partner_communication_reminder/models/recurring_contract.py
@@ -103,7 +103,7 @@ class RecurringContract(models.Model):
                     # create a 3rd reminder.
                     eligible_reminders["third"] += sponsorship
                 else:
-                    # Create 2nd reminder iff not already send
+                    # Create 2nd reminder if not already send
                     has_second_reminder = partnerCommunicationJob.search_count(
                         reminder_search
                         + [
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: partner_communication_reminder/models/recurring_contract.py [hunk 1]
```diff
@@ -99,8 +99,20 @@ def create_reminder_communication(self):
                     ]
                 )
                 if has_second_reminder:
+                    # If the 2nd reminder have been sent more than 20 days ago,
+                    # create a 3rd reminder.
                     eligible_reminders["third"] += sponsorship
-                eligible_reminders["second"] += sponsorship
+                else:
+                    # Create 2nd reminder iff not already send
```
NoeBerdoz: Typo
```suggestion
                    # Create 2nd reminder if not already send
```

partner_communication_reminder/models/recurring_contract.py
```diff
@@ -99,8 +99,20 @@ def create_reminder_communication(self):
...
+                    has_second_reminder = partnerCommunicationJob.search_count(
+                        reminder_search
+                        + [
+                            ("sent_date", ">=", twenty_days_ago),
+                            ("config_id", "=", reminder_confs[1].id),
+                        ]
+                    )
```
NoeBerdoz: What if a sponsorship already had a second reminder created on a past invoicing period, and needs a new one now? This would make the process wrong, it would fail to create a new second reminder for the current invoicing period.

partner_communication_reminder/models/recurring_contract.py
NoeBerdoz, line 1: I feel like the create_reminder_communication method is becoming messy, there are database calls that could be grouped in one call in my opinion. I would suggest to take the time to optimize its logic.

NoeBerdoz (changes requested):

SamBachmann:
Thanks for your feedback ! 

Now, it only perform one search for reminders in the database and not use several search_count anymore.
I also complete the logic to handle the case we have already sent a 3rd reminder.

I have test it and it seems to work well !

SamBachmann:
I forgot to remove tests changes in the code. It is good now :)

NoeBerdoz (approved):

PR #2043, opened 2025-07-11
add analytic account on interventions. The abilty to link interventio…

davidwul:
…n category to funds.


davidwul:
@ecino can you review/merge @Dylanvenancio made things correct now!

intervention_compassion/models/intervention_category.py
```diff
@@ -17,6 +17,7 @@ class InterventionCategory(models.Model):
 
     name = fields.Char(required=True, translate=False)
     type = fields.Selection("get_types")
+    fund = fields.Selection("get_fund")
```
ecino: We add this in intervention categories but then we will need to update Prod records to map the correct funds to the records.

Why not directly linking to the `account.analytic.group` model directly instead?

```suggestion
    analytic_group_id = fields.Many2one("account.analytic.group")
```

intervention_compassion/views/server_action.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
ecino: Why not doing this in a method directly in object (and using a button in the view) ?

intervention_compassion/views/compassion_intervention_view.xml
```diff
@@ -454,7 +455,7 @@
     <record id="open_categories_action" model="ir.actions.act_window">
         <field name="name">Intervention Categories</field>
         <field name="res_model">compassion.intervention.category</field>
-        <field name="view_mode">tree,form</field>
+        <field name="view_mode">tree,form,kanban</field>
```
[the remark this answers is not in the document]
ecino: I think this can be removed

ecino (changes requested):
I would prefer a direct link between intervention.category <-> account.analytic.group objects.

I would also like data provided for the desired account.analytic.group records (replacing the Selection list).

A migration script could be added to link the records in the xml data to existing records in Production if they are already created in Prod.

davidwul:
@ecino 
what needs to be done now?

PR #2044, opened 2025-07-22
T2511 [ADD] New fields for field offices

ecino:
- WorldBank and Factbook data can be synced

WARNING :

1. You need to add repository [community-data-files ](https://github.com/OCA/community-data-files.git) in your project and Odoo addons_path in order for the change to work.
2. You also need to install the wbgapi library (pip install wbgapi).
3. Then you need to update my_compassion module.


NoeBerdoz (approved):
Looks all good to me

PR #2050, opened 2025-09-26
T2624   make the letter creation progress bar show real checkpoints

Shayan105:
## :warning: Important note:
This PR, being used as basis for the following PR, should be merged first: 
https://github.com/CompassionCH/compassion-modules/pull/2051

This PR is currently linked to PR : https://github.com/CompassionCH/compassion-website/pull/163

## Summary
This PR is related to the PR : https://github.com/CompassionCH/compassion-website/pull/163 , being on another reposirtory.
The current PR adds more fields to the model `correspondence_s2b_generator` to describe the generation status of the letter. It also adds callback kwargs on some methods. 

### In this PR:
- `sbc_compassion/models/correspondence_s2b_generator.py` : Add callbacks keyword-arguments to methods  and add generation status field to the model. 
- `sbc_compassion/models/correspondence_template.py`: Add the callbacks and call them at checkpoints.

### In the related PR https://github.com/CompassionCH/compassion-website/pull/163
- `my_compassion/controllers/my2_letters.py`: Add routes to the controller to handle polling requests, define the callbacks to be called during the letter creation.
- `my_compassion/static/src/js/my2_new_letter.js`: Allow the client to poll the status of the creation and update the progress bar in consequence.
- `theme_compassion_2025/static/src/js/_progress_bar.js`: Complete the progress bar logic to allow forward progresion.

### DIFF
--- a/sbc_compassion/models/correspondence_template.py
+++ b/sbc_compassion/models/correspondence_template.py
@@ -112,7 +112,9 @@ class CorrespondenceTemplate(models.Model):
     ##########################################################################
     #                             PUBLIC METHODS                             #
     ##########################################################################
-    def generate_pdf(self, pdf_name, header, text, image_data, background_list=None):
+    def generate_pdf(
+        self, pdf_name, header, text, image_data, background_list=None, **callbacks
+    ):
         """
         Generate a pdf file
         This function is nearly as generic as it should be to be implemented
@@ -144,6 +146,8 @@ class CorrespondenceTemplate(models.Model):
             pages, header, background_list, temp_img
         )
         image_list = []
+
+        callbacks.get("apply_template_callback", lambda: None)()
 
         if background_list:
             # An original document is provided. We want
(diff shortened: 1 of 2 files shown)


sbc_compassion/models/correspondence_template.py
```diff
@@ -181,6 +185,7 @@ def generate_pdf(self, pdf_name, header, text, image_data, background_list=None)
                 text_list.append(text_box.get_json_repr())
             overflow_template = [add_background.name, header_data, text_list, []]
 
+        callbacks.get("apply_text_callback", lambda: None)()
```
ecino: Why is there a need to perform the callback like this only to update the progression? This introduces a lot of complexity in the code, but we could simply update the progession directly in the code or do you see some strong issues in doing like that?
```suggestion
        if s2b_generator is not None:
            s2b_generator.generation_status = "apply_text"
```
The generator would be passed in the arguments instead of the callbacks (see my other suggestion).

sbc_compassion/models/correspondence_template.py
```diff
@@ -112,7 +112,9 @@ def _compute_template_image(self):
     ##########################################################################
     #                             PUBLIC METHODS                             #
     ##########################################################################
-    def generate_pdf(self, pdf_name, header, text, image_data, background_list=None):
+    def generate_pdf(
+        self, pdf_name, header, text, image_data, background_list=None, **callbacks
```
ecino:
```suggestion
        self, pdf_name, header, text, image_data, background_list=None, s2b_generator=None
```

ecino (approved):

PR #2055, opened 2025-10-22
Fix resolve concurrency issues with letters status being polled

Shayan105:
## :warning: Important note
This PR is linked to this PR : https://github.com/CompassionCH/compassion-website/pull/181

## Summary

This PR fixes some issues resulting from the merging of the three following tasks : 
- T2379 Implement letter draft functionality : https://github.com/CompassionCH/compassion-website/pull/141
- T2624Make the letter creation progress bar show "real" checkpoints : https://github.com/CompassionCH/compassion-website/pull/163
- T2288 Force the maximum pages of a letter to be 15 : https://github.com/CompassionCH/compassion-modules/pull/2051

## 🐛 Problem Description

After the recent merges, several regressions were identified:

1.  **Page Limit Bypass**: The validation check to enforce the 15-page limit on letters was no longer being triggered, allowing oversized letters to be submitted.
2.  **UI Deadlock**: A **race condition** between the backend process updating the `generation_status` and the frontend polling for that status was causing a deadlock. The UI would get stuck in a loading state because it was reading an inconsistent status from the database before the transaction committed.
3.  **No Retry on Failure**: If a letter failed generation (e.g., for being over the page limit), its status was permanently set to 'failed'. Even after the user corrected the letter, the system would not allow a new generation attempt.

---

## ✅ Proposed Solutions

To resolve these issues, the following changes have been implemented:

1.  **Reinstate Page Limit Check**: The `preview()` method, which contains the page count validation, is now called during the final "send" action. This ensures the page limit is checked just before submission.
2.  **Eliminate Race Condition**: To guarantee the polling mechanism sees fresh data, the database update for the `generation_status` is now forced to **commit immediately** in its own transaction. This ensures atomicity and prevents the frontend from reading a stale state.
3.  **Enable Resubmission After Failure**: The letter generator's state is now correctly reset when a user attempts to resubmit a letter that previously failed. The `status` and `generation_status` fields are reverted to their initial values, allowing the corrected letter to be processed as a new attempt.

---

### DIFF
--- a/sbc_compassion/models/correspondence_s2b_generator.py
+++ b/sbc_compassion/models/correspondence_s2b_generator.py
@@ -252,14 +252,14 @@ class CorrespondenceS2bGenerator(models.Model):
             message = "Letters have been successfully generated."
             self.env.user.notify_success(message=message)
 
-            # Update state to done
-
-            self.state = "done"
-            self.date = fields.Datetime.now()
-            # Ensure atomicity
-            self.env.cr.commit()
-
-            return True
+            return self.write(
+                {
+                    "state": "done",
+                    "date": fields.Datetime.now(),
+                    "generation_status": "done",
+                    "generation_error_message": False,
+                }
+            )
 
         except Exception as error:
             # If the operation fails, notify the user with the error message
--- a/sbc_compassion/models/correspondence_s2b_generator.py
+++ b/sbc_compassion/models/correspondence_s2b_generator.py
@@ -168,7 +168,6 @@ class CorrespondenceS2bGenerator(models.Model):
             with Image(blob=pdf, resolution=96) as pdf_image:
                 preview = base64.b64encode(pdf_image.make_blob(format="jpeg"))
 
-                self.isolated_write({"generation_status": "done"})
                 return self.isolated_write(
                     {
                         "state": "preview",
@@ -176,7 +175,7 @@ class CorrespondenceS2bGenerator(models.Model):
                         "preview_pdf": base64.b64encode(pdf),
                     }
                 )
-                return
+
         except (PolicyError, TypeError, UserError, Exception) as error:
             error_message = (
                 _(
@@ -261,7 +260,6 @@ class CorrespondenceS2bGenerator(models.Model):
                 {
                     "state": "done",
                     "date": fields.Datetime.now(),
-                    "generation_status": "done",
                     "generation_error_message": False,
                 }
             )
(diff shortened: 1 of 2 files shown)


NoeBerdoz (approved):

ecino:
I'm not sure this change is needed. The new cursor you are creating is at the end of the method. At the end of the method when it returns, the main transaction will be committed so I don't see why this will effectively change something. Both polling the status and updating it should not cause deadlocks. It's doing simultaneous writes that can cause deadlocks.

Shayan105:
@ecino After discussion the following change has been applied : replace the update_generation_status() to a mode general isolated_write() do isolated trasactions to the DB. 
Some refactoring were applied on this repo and also in this one : https://github.com/CompassionCH/compassion-website/pull/181/files

sbc_compassion/models/correspondence_s2b_generator.py
```diff
@@ -165,14 +165,12 @@ def preview(self):
...
-
             with Image(blob=pdf, resolution=96) as pdf_image:
                 preview = base64.b64encode(pdf_image.make_blob(format="jpeg"))
-                return self.write(
+
+                return self.isolated_write(
                     {
                         "state": "preview",
-                        "generation_status": "done",
-                        "generation_error_message": False,
```
ecino, line 175: Shouldn't we keep this in case the preview is not called from the frontend controller but from the backend ?

sbc_compassion/models/correspondence_s2b_generator.py
```diff
@@ -252,11 +255,11 @@ def generate_letters_job(self):
             # If the operation succeeds, notify the user
             message = "Letters have been successfully generated."
             self.env.user.notify_success(message=message)
-            return self.write(
+
+            return self.isolated_write(
                 {
                     "state": "done",
                     "date": fields.Datetime.now(),
-                    "generation_status": "done",
```
ecino, line 259: Shouldn't we keep this for the backend calls?

sbc_compassion/models/correspondence_s2b_generator.py
Shayan105: Changes applied.

sbc_compassion/models/correspondence_s2b_generator.py
Shayan105: Changes applied. 
BPnQGGY

PR #2059, opened 2025-11-27
T2792 Ensure child protection on children page

Shayan105:
This PR is linked to this one : https://github.com/CompassionCH/compassion-website/pull/211


mobile_app_connector/models/compassion_project.py
```diff
@@ -29,6 +29,27 @@ def get_location_json(self, multi=False):
...
+            return {
+                "Latitude": int(self.gps_latitude),
+                "Longitude": int(self.gps_longitude),
+            }
+
+        return {
+            "Latitude": self.gps_latitude_obfuscated,
+            "Longitude": self.gps_longitude_obfuscated,
+        }
+
```
ecino: mobile_app_connector is a deprecated module and should not be changed. You can remove completely this method based on my other suggestion.
```suggestion
  ```

File: child_compassion/models/project_compassion.py [hunk 2]
```diff
@@ -85,6 +85,8 @@ class CompassionProject(models.Model):
     zip_code = fields.Char(readonly=True)
     gps_latitude = fields.Float(readonly=True)
     gps_longitude = fields.Float(readonly=True)
+    gps_latitude_obfuscated = fields.Float(readonly=True)
+    gps_longitude_obfuscated = fields.Float(readonly=True)
```
ecino: You could make computed fields instead that will be updated only if the gps coordinates are changed.
```suggestion
    gps_latitude_obfuscated = fields.Float(compute="_compute_gps_obfuscated")
    gps_longitude_obfuscated = fields.Float(compute="_compute_gps_obfuscated")
```

child_compassion/models/project_compassion.py
```diff
@@ -510,6 +512,53 @@ def _get_materials(self):
             ("Plastic", "Plastic"),
         ]
 
+    def update_obfuscated_coordinates(self):
```
```suggestion
    @api.onchange("gps_latitude", "gps_longitude")
    def _compute_gps_obfuscated(self):
```

child_compassion/models/project_compassion.py
```diff
@@ -510,6 +512,53 @@ def _get_materials(self):
...
+        if not api_key:
+            return
+
+        base_url = "https://maps.googleapis.com/maps/api/geocode/json"
+
+        for project in self:
+            # Check if we already have coords to avoid wasting API calls
+            if project.gps_latitude_obfuscated or project.gps_longitude_obfuscated:
+                continue
+
```
```suggestion
```

child_compassion/models/project_compassion.py
```diff
@@ -510,6 +512,53 @@ def _get_materials(self):
...
+                if data["status"] == "OK":
+                    location = data["results"][0]["geometry"]["location"]
+                    project.gps_latitude_obfuscated = location["lat"]
+                    project.gps_longitude_obfuscated = location["lng"]
+                else:
+                    logging.error(f"Geocoding error for {project.id}: {data['status']}")
+
+            except Exception as e:
+                logging.error(f"Request failed: {e}")
```
```suggestion
                logging.error(f"Request failed: {e}")
                project.gps_latitude_obfuscated = int(project.gps_latitude or 0)
                project.gps_longitude_obfuscated = int(project.gps_longitude or 0)
```

child_compassion/models/project_compassion.py
```diff
@@ -510,6 +512,53 @@ def _get_materials(self):
...
+
+            # Build the list of available address parts
+            parts = [
+                project.closest_city,
+                project.state_province,
+                project.country_id.name,
+            ]
+            address_string = ", ".join(filter(None, parts))
+            if not address_string:
+                continue
```
ecino: Removing this check so that it will raise an error below and fallback to the integer part.
```suggestion
```

child_compassion/models/project_compassion.py
```diff
@@ -510,6 +512,53 @@ def _get_materials(self):
...
+            params = {"address": address_string, "key": api_key}
+
+            try:
+                response = requests.get(base_url, params=params, timeout=10)
```
ecino: 10 seconds seems blocking for me. It should be faster.
```suggestion
                response = requests.get(base_url, params=params, timeout=3)
```

child_compassion/models/project_compassion.py
```diff
@@ -510,6 +512,53 @@ def _get_materials(self):
...
+                data = response.json()
+
+                if data["status"] == "OK":
+                    location = data["results"][0]["geometry"]["location"]
+                    project.gps_latitude_obfuscated = location["lat"]
+                    project.gps_longitude_obfuscated = location["lng"]
+                else:
+                    logging.error(f"Geocoding error for {project.id}: {data['status']}")
```
ecino: Raising an error for applying the integer obfuscation.
```suggestion
                    raise UserError(f"Geocoding error for {project.id}: {data['status']}")
```

ecino (changes requested):

child_compassion/models/project_compassion.py
Shayan105: This logic should remain in case the compute method is called manually to avoid wasting API calls. However, when the compute method is triggered by a onchange(lat,long), this should indeed not be checked.

PR #2060, opened 2025-12-04
T2795 dont send b2s intro letter not working

SlWa99:
# FIX: [Sponsorship] Add Security Filter to Block Unwanted Welcome Dossier (GMC Sync)

## Problem Description

The system was sending the Welcome Dossier (Introduction Letter) to sponsors of **“Transfer”** or **“Reinstatement”** contracts, despite the security flag `send_introduction_letter` being correctly set to `False` for these cases.

- **Business Rule:** This field is editable by the SDS for exceptional error corrections (manual overrides), meaning the system must respect the field’s final `True` or `False` value.

---

## Justification and Root Cause

My detailed tests proved that the logic failure was a **workflow bypass** that occurred during contract activation:

1. **DB Consistency Confirmed:**  
   The `send_introduction_letter` field was correctly saved as `False` in the database for *Transfer* contracts (the data itself was correct).

2. **Filter Bypass:**  
   The problem was isolated to the `contract_activate` method.  
   This method executes the external synchronization calls (`upsert_sponsorship`) **unconditionally** for any activated contract.

3. **GMC Trigger:**  
   The unconditional call to `upsert_sponsorship` informs the external GMC system that a new sponsorship is active, which subsequently triggers the generation and sending of the Welcome Dossier, bypassing all Odoo internal filters.

---

## The Applied Correction and Impact

The solution ensures that the security flag is respected at the point of external communication trigger:

### 1. Modification  
The synchronization loop within `contract_activate` is **modified to execute only if the SDS has authorized the communication** (`send_introduction_letter is True`).

#### **Blocking**
If the field is `False` (standard Transfer case), the contract is excluded from the GMC sync, blocking the unwanted Welcome Dossier.

#### **Other Communications**
This change targets only the `upsert_sponsorship` step — the sending and generation of other communications (birthday letters, beneficiary letters, participation letters...) managed by separate workflows (CRON jobs) is **not affected** by this modification.

---

### See also
Related PR addressing the same issue : https://github.com/CompassionCH/compassion-switzerland/pull/1710

### DIFF
--- a/sponsorship_compassion/__manifest__.py
+++ b/sponsorship_compassion/__manifest__.py
@@ -45,6 +45,7 @@
         "partner_firstname",  # oca/partner-contact
         "account_analytic_attribution",  # compassion-accounting
         "sale",
+        "partner_communication_compassion",
     ],
     "data": [
         "views/invoice_line_view.xml",
--- a/sponsorship_compassion/models/contracts.py
+++ b/sponsorship_compassion/models/contracts.py
@@ -56,14 +56,14 @@ class SponsorshipContract(models.Model):
     birthday_invoice = fields.Float(
         "Annual birthday gift",
         help="Set the amount to enable automatic invoice creation each year "
-             "for a birthday gift. The invoice is set two months before "
-             "child's birthday.",
+        "for a birthday gift. The invoice is set two months before "
+        "child's birthday.",
         tracking=True,
     )
     christmas_invoice = fields.Float(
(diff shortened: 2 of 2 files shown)


ecino:
This is not a desired change. Sponsorship_compassion should not depend on communication and the contract_active method needs to be called as well for the sponsorships that have the flag. The flag is only use to prevent the communication generation.

PR #2061, opened 2025-12-23
T2715 [MIG] interaction_resume: Migration to 18.0

SamBachmann:
This PR migrates the interaction_resume module from Odoo v14 to v18.

### Key changes:

- **UI/UX**: Replaced deprecated tree views with the new list view type.
- **Framework**: Converted legacy attrs (invisible/readonly/required) to standard v18 Python-like expressions.
-  **JavaScript**: Fully rewrote the frontend logic using OWL Components and the ListController patch to replace legacy Widgets.
- **Backend**: Renamed the deprecated refresh() method to action_refresh() to comply with framework changes.

### ⚠️ Warning: Testing Status

I have successfully resolved all migration-related conflicts and ensured the module installs without manifest errors. However, I have not yet been able to fully verify the functional behavior in a live v18 environment due to local setup complexities. @ecino I would appreciate if you give me more info about the usage of this module.


SamBachmann:
@ecino As the idea is to simply migrate the module and not to totally refactor it, should I address gemini' review issues ?

ecino:
@SamBachmann I fixed some migration issues that prevented the installation (I forced pushed on the branch, you can fetch and reset your local branch to get the latest state). Now it hase OWL errors when trying to load the view which you can check. The module is adding the "Interaction Resume" button inside the partner view, which displays the history of all exchanged messages.

ecino (approved):

PR #2075, opened 2026-02-16
[T2986] FIX: direction is now explicitly set through message values

Danielgergely:
- FIX: fetch the whole action instead of only the id and set the id and direction for the message
- REFACTOR: added type to super().create() in gmc_message for clarity

Explanation: 
The direction field on `gmc.message `was not persisting to the database because of conflicts with Odoo's ORM logic:
- The related field approach failed because the source model (`gmc.action`) explicitly blocks writes to this field, which interrupted the ORM's synchronization process during creation. 
- The computed field also failed because even though the value was calculated in memory, the system failed to trigger a secondary SQL update to save it. 

The only approach which worked was to manually inject the value into the creation dictionary (vals) before calling `create()`, ensuring the data is included directly in the primary SQL INSERT statement.

### DIFF
--- a/message_center_compassion/models/gmc_message.py
+++ b/message_center_compassion/models/gmc_message.py
@@ -106,7 +106,7 @@ class GmcMessage(models.Model):
             )
 
         if not message:
-            message = super().create(vals)
+            message = super(GmcMessage, self).create(vals)
 
         if message.action_id.auto_process:
             message.process_messages()
--- a/sbc_compassion/models/correspondence.py
+++ b/sbc_compassion/models/correspondence.py
@@ -609,12 +609,13 @@ class Correspondence(models.Model):
         )
         messages = self.env["gmc.message"]
         for letter in self:
-            action_id = self.env.ref("sbc_compassion.create_letter").id
+            action = self.env.ref("sbc_compassion.create_letter")
             message_vals = {
-                "action_id": action_id,
+                "action_id": action.id,
                 "object_id": letter.id,
                 "child_id": letter.child_id.id,
                 "partner_id": letter.partner_id.id,
+                "direction": action.direction
             }
             if (
                 letter.sponsorship_id.state not in ("active", "terminated")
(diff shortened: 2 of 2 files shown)


sbc_compassion/models/correspondence.py
```diff
@@ -609,12 +609,13 @@ def create_commkit(self):
...
-            action_id = self.env.ref("sbc_compassion.create_letter").id
+            action = self.env.ref("sbc_compassion.create_letter")
             message_vals = {
-                "action_id": action_id,
+                "action_id": action.id,
                 "object_id": letter.id,
                 "child_id": letter.child_id.id,
                 "partner_id": letter.partner_id.id,
+                "direction": action.direction,
             }
```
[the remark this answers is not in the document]
Danielgergely: This was not working, that is why I changed this.

ecino:
Shouldn't we simply remove the "readonly" attribute on the field declaration (and maybe put it in the view instead) ?

message_center_compassion/models/gmc_message.py
```diff
@@ -106,7 +106,7 @@ def create(self, vals):
             )
 
         if not message:
-            message = super().create(vals)
+            message = super(GmcMessage, self).create(vals)
```
ecino, line 109: At times, pre-commit complained about such writings... Isn't it better to skip those in this case? What are your thoughts?

Danielgergely:
> Shouldn't we simply remove the "readonly" attribute on the field declaration (and maybe put it in the view instead) ?

I agree, it would be much cleaner and nicer, but I tried that along with multiple other better fixes (like computed fields) but unfortunately, it didn't work (for me on my machine)

message_center_compassion/models/gmc_message.py
Danielgergely: I think this is a cleaner and more readable way to explicitly pass on the type to the parent, but it is not necessary. If pre-commit complains, then I can remove it. I just saw that in some cases this is used, but in other cases its omitted.

ecino (approved):

PR #2080, opened 2026-02-23
T3017 [FIX] replace isolated_write with write in correspondence handling

ecino:
- Happy path should use main cursor
- Only exceptions should use a new cursor in DB
- Remove redundant calls

@Danielgergely This reverts part of the changes you did for T2986 (GMC message direction), but I think I found the root cause of the problem. For a reason, the separate transactions with isolated_writes prevented the relational and compute fields to be set correctly. Can you please review those changes and let me know what you think?

This change also fixes T3017

### DIFF
--- a/sbc_compassion/models/correspondence.py
+++ b/sbc_compassion/models/correspondence.py
@@ -607,14 +607,13 @@ class Correspondence(models.Model):
             datetime.date.today()
         )
         messages = self.env["gmc.message"]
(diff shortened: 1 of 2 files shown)


Danielgergely (approved):
@ecino Looks good to me (though I still have limited knowledge of the whole correspondence forrest :) )

NoeBerdoz (approved):
Looks good to me, thanks 👍

PR #2102, opened 2026-06-17
T3227 - Golden ticket - New info about survival program for churches in odoo

danpa32:
image

The first golden ticket! :grinning: 

The purpose was to add information about survival sponsorship for a supporter. For a church, the information is an aggregation of data using the church data and its members. For a member, it is only his/her data that are used.

https://github.com/user-attachments/assets/a7acbb7a-a6f8-452b-bbd8-6ebe4322b21c


survival_sponsorship_compassion/models/contracts_report.py
```diff
@@ -0,0 +1,168 @@
+##############################################################################
+#
+#    Copyright (C) 2018 Compassion CH (http://www.compassion.ch)
```
ecino: I wish you will once show me your time-travel machine! But in the meantime we no longer use those headers so feel free to remove it (they quickly become irrelevant in most cases).

survival_sponsorship_compassion/models/res_partner.py
```diff
@@ -35,3 +76,26 @@ def _compute_related_contracts(self):
...
+            "domain": [
+                ("type", "=", "CSP"),
+                ("state", "=", "active"),
+                "|",
+                ("partner_id", "=", self.id),
+                ("partner_id.church_id", "=", self.id),
+            ],
+            "context": {
+                "default_type": "CSP",
+                "default_state": "active",
```
ecino: Why this?
```suggestion
```

survival_sponsorship_compassion/models/res_partner.py
```diff
@@ -1,9 +1,50 @@
...
+from odoo import _, models, fields
 
 class ResPartner(models.Model):
     _inherit = "res.partner"
 
+    survival_sponsorship_count = fields.Integer(
+        string="Survival sponsorship(s)",
+        compute="_compute_active_csp_count",
+        store=False,  # Not stored in the database, computed on the fly
+        copy=False,
```
ecino: This is the default behavior, it could be ommited.
```suggestion
```

survival_sponsorship_compassion/models/res_partner.py
```diff
@@ -1,9 +1,50 @@
-from odoo import models
+##############################################################################
+#
+#    Copyright (C) 2018 Compassion CH (http://www.compassion.ch)
```
ecino: See my other comment

survival_sponsorship_compassion/models/res_partner.py
```diff
@@ -1,9 +1,50 @@
...
+        contract_counts = {}
+        for contract in contracts:
+            pid = contract.partner_id.id
+            contract_counts[pid] = contract_counts.get(pid, 0) + 1
+
+        for partner in self:
+            count = contract_counts.get(partner.id, 0)
+            if partner.is_church:
+                count += sum(contract_counts.get(mid, 0) for mid in partner.member_ids.ids)
+            partner.survival_sponsorship_count = count
```
ecino: A few things here.

- The beginning is not needed
- Using [read_group](https://www.odoo.com/documentation/14.0/developer/reference/addons/orm.html#odoo.models.Model.read_group) will be more performant and simplifies the code
```suggestion
        partners = self
        churches = self.filtered("is_church")
        if churches:
            partners |= churches.mapped("member_ids")

        group_data = self.env["recurring.contract"].read_group([
            ("partner_id", "in", partner_ids),
            ("type", "=", "CSP"),
            ("state", "=", "active"),
        ], ["partner_id"], ["partner_id"])
        contract_counts = {
            partner_data["partner_id"][0]: partner_data["partner_id_count"]
            for partner_data in group_data
        }
        for partner in self:
            count = contract_counts.get(partner.id, 0)
            if partner.is_church:
                count += sum(contract_counts.get(mid, 0) for mid in partner.member_ids.ids)
            partner.survival_sponsorship_count = count
```

survival_sponsorship_compassion/models/contracts_report.py
```diff
@@ -0,0 +1,168 @@
...
+    sr_countries_previous = fields.Char(
+        "Countries previously impacted",
+        compute="_compute_sponsorship_metrics",
+        help="List of previously impacted countries by the church and its members by the CSP program.",
+    )
+
+    def _compute_sponsorship_metrics(self):
+        if not self:
+            return
+
```
ecino, line 47: Not needed
```suggestion
```

survival_sponsorship_compassion/models/contracts_report.py
```diff
@@ -0,0 +1,168 @@
...
+        help="List of current countries impacted by the church and its members by the CSP program.",
+    )
+
+    sr_countries_previous = fields.Char(
+        "Countries previously impacted",
+        compute="_compute_sponsorship_metrics",
+        help="List of previously impacted countries by the church and its members by the CSP program.",
+    )
+
+    def _compute_sponsorship_metrics(self):
```
ecino, line 42: Please check Gemini and Greptide comments about using SQL. Why did you go this route, is it for performance, code readability, or based on your confidence? We should assess carefully if this is justified compared to use the regular ORM otherwise the good practice is to avoid as much as possible bypassing the ORM.

ecino (changes requested):

ecino:
Thank you for addressing this Golden Ticket and for the good work! Please check my comments. You need as well to run pre-commit on your end (in your python virtual environment):
```
pip install pre-commit
pre-commit run -a
```

survival_sponsorship_compassion/models/contracts_report.py
danpa32: That is called copy paste 😄

survival_sponsorship_compassion/models/res_partner.py
danpa32: Oh ok. I am gonna check what the read_group does. But in the meantime I will make the changes.

One correction though, I need the partner_ids so I will make this change 

```
churches = self.filtered("is_church")
        all_partner_ids = set(self.ids)
        if churches:
            all_partner_ids.update(churches.mapped("member_ids.id"))

        group_data = self.env["recurring.contract"].read_group(
            domain=[
                ("partner_id", "in", list(all_partner_ids)),
                ("type", "=", "CSP"),
                ("state", "=", "active"),
            ],
            fields=["partner_id"],
            groupby=["partner_id"],
        )

        contract_counts = {
            item["partner_id"][0]: item["partner_id_count"]
            for item in group_data
        }
```

survival_sponsorship_compassion/models/res_partner.py
danpa32: Oups, indeed I do not think we want that. In the process, Should I also make sure we cannot create a survival sponsorship on this part?

survival_sponsorship_compassion/models/contracts_report.py
danpa32: I used SQL first because I am more confident to use it and understand it better currently than the current ORM, especially due to the aggregation of information in one go into the database. After checking with gemini, here is the detailed answer it gives about using SQL vs ORM in this case. This differ from the res_partner SQL query I did which can be replaced by code not bypassing the ORM

**Gemini response**

**Why Raw SQL was Chosen**
1. The "N+1" Problem and Complex Aggregation
The ORM is excellent at fetching objects, but it struggles with complex multi-level aggregations across related records (e.g., calculating counts and distinct country sets for a partner plus their child members in a single sweep).

Without SQL: You would likely need to perform multiple search() calls or iterate through member_ids for every church. If you have 50 churches, each with 20 members, you are potentially triggering hundreds of individual database hits.

With SQL: You perform a UNION ALL or a single JOIN. The database engine executes this in one operation, returning exactly the aggregated data needed. For large datasets, the performance difference is often exponential.

2. Cross-Table Join Logic
Your logic requires joining res_partner to recurring_contract and, further down, to account_move and account_move_line.

The ORM's read_group is powerful, but it is limited to a single model. It cannot perform a JOIN across account_move_line to filter by payment_state and last_payment dates simultaneously while grouping by partner_id across both the parent and its members.

The raw SQL approach allows for a "Big Data" style retrieval where you grab the entire relevant state in two database round-trips.

ecino (approved):
👍 LGTM

NoeBerdoz (changes requested):
Thanks for your work @danpa32!

**Here is my review, there is a few things to fix before we can merge it:**

### 1. The "moms supported for 1 year" calculation:

The spec asks for lifetime donations: "_the number of CSP donation … (lifetime)_" and "_even though some sponsorships already ended_". It is confirmed with the calculations done from email Mirco's own examples that takes in account an 18-month sponsorship that
  "then stopped" and "a one time donation." 

The 12-month window implemented drops those cases. The 1 year parameter here stands for the divider ÷744 which represents 62 CHF (1 mom sponsorship) * 12 (months). It represents one mom for one year denominator. 

Analogy: _"I've paid 10,000 CHF in rent over the years, rent is 744CHF/year; so I've covered 13.44 years of rent."_ You sum all the rent you ever paid and divide by the yearly rate. You don't restrict the 10,000 CHF to the last 12 months, but the answer is still expressed "in years," because you divided by a yearly rate.


### 2. The 62.0 value in monthly_cost 

Your intent of declaring "use the product price, fall back to 62 if it's not set" is wrong. Odoo by default puts a default price of 1.0 on a new product. Relying on list_price is the correct direction in my opinion, but you can drop the 62.0 hard value fallback and you need to ensure that the data reflects our business data by seeding the key price in a data file `/data/survival_product_template.xml` is the great candidate here.


### 3. The wanted results was presented as decimal, not integer

`sr_nb_moms_supported_for_a_year` is declared as an Integer, and the compute does int(total / annual). Mirco in his examples wants decimal (ex. 13.44). rounding the count is also not optimal when dealing with <1 values, as 0.X  will show 0.


### 4. Not sure about this one, but survival-sponsorship count looks computed twice
The same number ends up in two fields right? `survival_sponsorship_count` in `res_partner.py` and `sr_survival_sponsorship_count` in `contracts_report.py` (for the report, via the raw SQL). This it's not an issue point, but more a maintainability consideration, if that definition changes later for whatever business rules reason, it has to be updated in both places, might be worth considering extracting the rule into a single helper that both fields use.


### 5. Relabeling drops the existing translation

Renaming `sr_sponsorship` from "Number of sponsorship" to "Number of child sponsorships" may change a translatable string, so the existing fr/de/it translations (if existent) for that label will be lost if the .po files are not updated.

NoeBerdoz (approved):
Great job @danpa32, thank you very much!

PR #2117, opened 2026-07-23
[T1984] FIX : Prevent automatic letter publishing on translation removal

Ylulu71D:
## Goal
Update `action_remove_local_translate` to safely remove B2S letters from the translation platform upon activation of the existing "Remove from translation" button, revert its state to `"Received in the system"`, and clean up translation metadata without triggering automated publishing or email workflows.

## Technical Aspects
* **Record Enforcement & State Reset:**
  * Enforced single-record execution using `self.ensure_one()`.
  * Updated record fields via `write()`:
    * Reverted `state` to `"Received in the system"`.
    * Cleared `translator` and `translation_status` (reset to `False`).

* **Retained The Existing Directional Handling for S2B**

## MISC
- In theory, this action intentionally avoids triggering standard automated publishing or email notifications by resetting the state to "Received in the system" (previously to "Published" which activates downstream workflows). 
- MailHog tests have been conducted in local instances but did not yield concrete validating results. Further investigation in Stage environment is advised.

### DIFF
--- a/sbc_translation/models/correspondence.py
+++ b/sbc_translation/models/correspondence.py
@@ -459,6 +459,7 @@ class Correspondence(models.Model):
             {
                 "state": "Received in the system",
                 "translator": False,
+                "new_translator_id": False,
                 "translation_status": False,
             }
         )
--- a/sbc_translation/models/correspondence.py
+++ b/sbc_translation/models/correspondence.py
@@ -449,16 +449,24 @@ class Correspondence(models.Model):
     def action_remove_local_translate(self):
         """
         Remove a letter from local translation platform and change state of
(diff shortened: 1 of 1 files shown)


Danielgergely (approved):
- Before: the B2S branch (Published to Global Partner + process_letter(force_publish=True)) actually raises AttributeError: 'super' object has no attribute 'process_letter' on the v18 branch — old code was broken, not just over-sending.
- After: B2S → Received in the system, metadata cleared, no gmc.message/comm job. S2B → Received in the system + commkit still created (preserved).

Confirms the requirement: removed B2S letters are parked, not auto-sent.

PR #2131, opened 2026-08-06
T3359-Childpack_typo

loris-fab:
## Goal

The childpack PDF printed a garbled line under the child photo. In French it read
`8 NÉ NÉ 01 JANV. 2018 )` instead of `8 (NÉ LE 01 JANV. 2018)`, and German and
Italian had the same problem.

## Technical aspect

Two distinct bugs in `child_compassion/report/childpack.xml`, both in the
`.summary` block:

- The parentheses around the birth date were wrapped in their own `<span>`, which
  exposed `(` as a translatable term. `fr_CH`, `de` and `it` had it translated as
  `Né` / `Geb.` / `Nato`, so the opening parenthesis was replaced by a second birth
  label. They are now drawn with `:before` / `:after` CSS rules, which no
  translation can reach. The `msgid "<span>(</span>"` entries are removed from the
  six `.po` files.
- The spans sat on separate source lines, so HTML whitespace collapsing added a
  space on both sides of each parenthesis. Nesting the label and the date inside a
  single wrapper span removes those.

French and Italian labels lacked prepositions (`Né` instead of `Né le`). Updating `msgstr` alone doesn't apply on standard Odoo upgrades due to `overwrite=False`. To force the update across environments without needing `--i18n-overwrite`, the source term was renamed from `Born` to `Born on`, creating a new `msgid`.

## Misc

No misc

### DIFF
--- a/child_compassion/report/childpack.xml
+++ b/child_compassion/report/childpack.xml
@@ -132,10 +132,10 @@
                             <t
               t-if="o.gender=='F' and o.env.lang in ('fr_CH','it_IT')"
             >
-                                <span>Born (F)</span>
+                                <span>Born on (F)</span>
                             </t>
                             <t t-else="">
-                                <span>Born</span>
+                                <span>Born on</span>
                             </t>
                             <span
               t-esc="o.get_date('birthdate', 'dd MMM YYYY')"
--- a/child_compassion/report/childpack.xml
+++ b/child_compassion/report/childpack.xml
@@ -125,20 +125,22 @@
                 <div class="summary">
                     <p t-field="o.preferred_name" class="summary_field name" />
                     <p class="summary_field birthday">
-                        <t t-if="o.age">
-                            <span t-esc="o.age" />
-                            <span>(</span>
-                        </t>
-                        <t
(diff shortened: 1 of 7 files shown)


child_compassion/report/childpack.xml
```diff
@@ -257,6 +259,14 @@
...
                 font-size: 9pt;
             }
+            /* Drawn in CSS so that the parentheses are neither exposed as
+               translatable terms nor spaced out by the HTML whitespace collapsing */
+            .summary_field.birthday .parenthesis:before {
+                content: "(";
+            }
+            .summary_field.birthday .parenthesis:after {
+                content: ")";
+            }
```
ecino, line 269: I'm not sure this is compatible with WKHTMLTOPDF rendering. Did you try generating a PDF file and does it work?

loris-fab: To be sure, I can double-check.

loris-fab: @ecino  So, I did a quick check and it works. It took me a while because, stupidly, I hadn't used `-u all` and so I thought it wasn't working.

PR #2145, opened 2026-09-02
T3434-FixChildpack

loris-fab:
## Goal
Companion PR in compassion-switzerland (mini childpack note + paper tray selection): CompassionCH/compassion-switzerland#1812.

When a childpack was sent to the printer ("Print background" unchecked), it always came out in the language of the user's interface (typically English) instead of the language selected in the wizard the downloaded PDF was correct, only the printed one was wrong. Additionally, when a child has no photo yet, the childpack should say so in the selected language instead of leaving an empty frame.

## Technical aspect
- **Language fix** (`child_compassion/models/ir_actions_report.py`, new): QWeb unconditionally overwrites the `lang` rendering value with the environment language (`ir.qweb._prepare_environment`). The wizard's language travels correctly in the report `data`, but `base_report_to_printer`'s `print_document` renders from the user's own environment, so the selected language was lost at rendering time. The new `_render_qweb_html` override aligns the environment language with `data["lang"]` for all childpack reports, they are recognised by their rendering model inheriting `report.child_compassion.childpack_full`, so the Swiss Mini childpack is covered too (its template has no `t-lang` at all).
- **Missing photo note** (`report/childpack.xml`): the photo frame of the full and small childpacks now shows a translated note ("No photo registered for this child") when the child has no fullshot. Translations added to the fr_CH/de/it `.po` files.

### DIFF
--- a/child_compassion/models/ir_actions_report.py
+++ b/child_compassion/models/ir_actions_report.py
@@ -1,0 +1,34 @@
+##############################################################################
+#
+#    Copyright (C) 2026 Compassion CH (http://www.compassion.ch)
+#    Releasing children from poverty in Jesus' name
+#
+#    The licence is in the file __manifest__.py
+#
+##############################################################################
+from odoo import models
+
+
+class IrActionsReport(models.Model):
+    _inherit = "ir.actions.report"
+
+    def _render_qweb_html(self, report_ref, docids, data=None):
+        # QWeb overwrites the ``lang`` rendering value with the language of the
+        # rendering environment (``ir.qweb._prepare_environment``), so the
+        # language selected in the print wizard is lost whenever the childpack
+        # is rendered from an environment in another language — typically
+        # base_report_to_printer's ``print_document``, which renders with the
+        # user's own language when the report is sent to the printer. Align the
+        # environment language with the requested one for all childpack reports
+        # (their rendering models inherit ``report.child_compassion.childpack_full``).
+        lang = data and data.get("lang")
+        if lang:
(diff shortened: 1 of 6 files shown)


child_compassion/models/ir_actions_report.py
```diff
@@ -0,0 +1,34 @@
...
+        if lang:
+            report_model = self._get_rendering_context_model(
+                self._get_report(report_ref)
+            )
+            childpack_class = self.env.registry[
+                "report.child_compassion.childpack_full"
+            ]
+            if report_model is not None and isinstance(report_model, childpack_class):
+                self = self.with_context(lang=lang)
+        return super()._render_qweb_html(report_ref, docids, data=data)
```
ecino: I think this is safe to apply to every report and not just only the childpack
```suggestion
        lang = data and data.get("lang") or self.env.lang
        return super(IrActionsReport, self.with_context(lang=lang)._render_qweb_html(
            report_ref, docids, data=data)
```

ecino (changes requested):

child_compassion/models/ir_actions_report.py
loris-fab: Agreed. I checked core, base_report_to_printer and our modules: the only callers putting a lang key in the report data are the childpack wizard and the tax receipt wizard (report_compassion), and the latter already renders with with_context(lang=data["lang"]), so applying it to every report changes nothing for existing ones.

ecino (approved):

PR #2150, opened 2026-09-07
14.0 T3436 visible letters during suspension

danpa32:
# T3436 — Sponsor's letter disappears from MyCompassion while the FCP is suspended

Reported: whenever an FCP is suspended, letters written by the sponsor automatically go into state `"Exception"` and vanish from MyCompassion. Sponsors aren't informed about a suspension until it has lasted three months, so a letter disappearing immediately (with no confirmation it was ever received) is a problem. 

Flagged as related to ticket T3322, which found and deliberately deferred this exact gap as a follow-up.

## Related PR

v14: https://github.com/CompassionCH/compassion-website/pull/380
v18: https://github.com/CompassionCH/compassion-modules/pull/2151

## What changed

Two repos, two versions each (v14 and v18 fixed in parallel, same root cause confirmed in both) 

- `compassion-modules` (`sbc_compassion`): adds `on_hold` (plain `Boolean`, not computed), set to `True` alongside `state = "Exception"` wherever a letter is put on hold for a benign reason 
- `compassion-website` (`my_compassion`): `is_published`'s compute now reads `on_hold` — an S2B letter in `"Exception"` is published if `on_hold` is `True`. **`on_hold` is deliberately excluded from `@api.depends`**

## How to test manually

1. Find an S2B letter whose project is currently suspended and already sits in `state = "Exception"` (or wait for/trigger a suspend→`hold_letters_action()` cycle on an active sponsorship with an existing letter).
2. Confirm pre-fix baseline: `is_published = False` for that letter, and it's absent from `/my2/children/letters` for that sponsor.
3. Apply the fix (upgrade `sbc_compassion` then `my_compassion` — order matters, see migration notes above), or run the migrations via `-u sbc_compassion,my_compassion`.
4. Confirm: `state` is still `"Exception"` (letter is still genuinely held, nothing sent to GMC), but `on_hold = True` and `is_published = True`.
5. Log in as that sponsor (or use `impersonate_login`'s "Switch Login" from Settings → Users) and confirm the letter now appears on `/my2/children/letters`, indistinguishable from a normally-processing letter (no badge/wording change — deliberate, matches the ticket's own framing of not revealing the suspension before 3 months).
6. Confirm a **genuine** exception (e.g. a letter in `"Quality check unsuccessful"`, or manually set `on_hold = False` on an `"Exception"` letter) still stays hidden — no regression on the original behavior for real failures.


sbc_compassion/migrations/14.0.1.0.4/pre-migration.py
```diff
@@ -0,0 +1,17 @@
...
+        # pre-migration, per odoo/modules/loading.py) finds it already
+        # exists and does not schedule a recompute for existing rows -
+        # matches odoo/tools/sql.py's own create_column for booleans.
+        openupgrade.logged_query(
+            env.cr,
+            """
+            ALTER TABLE correspondence
+            ADD COLUMN IF NOT EXISTS on_hold boolean DEFAULT false
+            """,
+        )
```
ecino: This is no longer needed as your field is not a compute field anymore.

ecino (changes requested):

